PR #3, opened 2022-09-12
Add basic documentation

fpetkovski:
cc @metalmatze


README.md
```diff
@@ -2,4 +2,61 @@
 
 A multi-threaded implementation of a PromQL Query Engine based on the [Volcano/Iterator model](https://paperhub.s3.amazonaws.com/dace52a42c07f7f8348b08dc2b186061.pdf).
 
-The project is currently under development.
+The project is currently under active development.
+
+## Design
+
+At the beggining of a PromQL query execution, the query engine computes a physical plan consisting of multiple independent operators, each responsible for calculating one part of the query expression.
```
bwplotka, line 9:
```suggestion
At the beggining of a PromQL query execution, the query engine computes [a physical plan](https://i0.wp.com/blog.knoldus.com/wp-content/uploads/2019/11/Fig-1-Plan-1.jpg?w=657&ssl=1) consisting of multiple independent operators, each responsible for calculating one part of the query expression.
```

README.md
```diff
@@ -2,4 +2,61 @@
...
+
+Operators are assembled in a tree-like structure with every operator calling `Next()` on its dependants until there is no more data to be returned. The result of the `Next()` function is a _column vector_  (also called a _step vector_) with elements in the vector representing samples with the same timestamp from different time series.
```
bwplotka, line 11: I wonder if we could mention IF this step vector contains ALL step series OR a partial view of series. As I see the design permits both, which might offer better scalability (no need for pulling all data for query in memory).

README.md
```diff
@@ -2,4 +2,61 @@
...
+
+<p align="center">
+  <img src="./assets/design.png" />
```
bwplotka, line 14: Hm, I am still unsure how rate/over_time fits into this operator model as they can be fully parallelized across series (series independent). I can see it's not yet implemented/used in the current implementation.

We can leave it for now, but let's add some todo/mention future work (:  Essentially e.g rate or sum_over_time requires multiple "steps" which might be different than root steps!

README.md
```diff
@@ -2,4 +2,61 @@
...
+This model allows for samples from individual time series to flow one execution step at a time from the left-most operators to the one at the very right. Since most PromQL expressions are aggregations, samples are reduced in number as they are pulled by the operators on the right. Because of this, samples from original timeseries can be decoded and kept in memory in batches instead of being fully expanded.
+
+In addition to operators that have a one-to-one mapping with PromQL constructs, the Volcano model also describes so-called Exchange operators which can be used for flow control and optimizations, such as concurrency or batched selects. An example of an _Exchange_ operator is described in the [Intra-operator parallelism](#intra-operator-parallelism) section.
+
+## Inter-operator parallelism
+
+Since operators are independent and rely on a common interface for pulling data, they can be run in parallel to each other. As soon as one operator has processed data from an evaluation step, it can pass the result onward so that its upstream can immediatelly start working on it.
+
+<p align="center">
+  <img src="./assets/promql-pipeline.png" />
```
bwplotka, line 26: `m[2]` -> `m[2m]`

README.md
```diff
@@ -2,4 +2,61 @@
...
+</p>
+
+
+## Memory management 
+
+### Step vector allocations
+
+One challenge with the streamed execution model is knowing how much memory to allocate in each operator for each step. 
+
+To work around this issue, operators expose a `Series()` method which returns the labels for all time series that they will ever produce. Operators at the very bottom of the tree, like vector and matrix selectors, have this information since they are responsible for loading data from storage. Other operators can then call `Series()` on the downstream operator and pre-compute all possible outputs.
```
bwplotka, line 45:
```suggestion
To work around this issue, operators expose a `Series()` method which returns the labels for all time series that they will ever produce (for all `Next()` calls). Operators at the very bottom of the tree, like vector and matrix selectors, have this information since they are responsible for loading data from storage. Other operators can then call `Series()` on the downstream operator and pre-compute all possible outputs.
```

bwplotka: Am I right?

bwplotka: If this is true I am not sure this is useful. With high churn I could have 100 million series in `Series` but `Next` would give me batches at max series at  the current moment which could be smaller e.g. 10k series.

README.md
```diff
@@ -2,4 +2,61 @@
...
+
+Even thought this might look like an expensive operation, its cost is identical to just one evaluation step. Knowing sizes of input and output vectors also allows us to:   
+* Allocate memory very precisely by properly sizing vector pools (see section bellow),
+* Use arrays instead of maps for indexing data and performing lookups, leading to even further memory usage and faster execution time, and
+* Use tight loops in operators by eliminating a large amount of conditional statements.
+
+### Vector pools
+
+Since time series are decoded one step at a time, vectors between execution execution steps can be recycled manually instead of relying on the garbage collector. Each operator has its own pool that it uses to allocate new step vectors and send results to its upstream. Whenever the upstream operator is finished with processing a step vector, it will return that vector to the pool of its downstream so that it can be reused again for subsequent steps.
```
bwplotka, line 54: 👍🏽

README.md
```diff
@@ -2,4 +2,61 @@
...
+
+## Concurrency control
```
bwplotka, line 56: What about memory control?

bwplotka (commented):
Nice! Some comments, otherwise good for now 💪🏽 

I would mention some `## TODO` which states things which are still todo, WDYT? 🤔

README.md
fpetkovski: I agree that for high churn, this is not that useful. But I would assume most cases are not that high churn so we would have good estimates. However, we can iterate based on real-world data, now we only have synthetic benchmarks.

README.md
fpetkovski: We don't have anything in place yet. Should we look at something like max samples as a proxy, or try to estimate memory more precisely?

README.md
fpetkovski: Right now we populate the step vector with all series. We could play with the partial mode as well and if that is possible to do, then memory management will also become better since we will be able to use fixed size vectors that are sent forward when full.

README.md
fpetkovski: Right now we can push rates into selectors so that the output is still a step vector. So the matrix selector would have a `FuncCall` that it executes over selected samples and produces a single point that is then sent as part of a vector to the next operator. 

Right now it's not implemented because there was too much refactoring and changing of interfaces, so it became cumbersome to keep everything up to date :)

README.md
fpetkovski: Also worth keeping in mind is that the result always has to have all series. So maybe we can only use this information when allocating vectors for the final result, and not for intermediate operators. Actually each operator might decide whether to use this information or not.

README.md
ahayworth, line 9:
```suggestion
At the beginning of a PromQL query execution, the query engine computes a physical plan consisting of multiple independent operators, each responsible for calculating one part of the query expression.
```

README.md
```diff
@@ -2,4 +2,61 @@
...
+  <img src="./assets/design.png" />
+</p>
+
+This model allows for samples from individual time series to flow one execution step at a time from the left-most operators to the one at the very right. Since most PromQL expressions are aggregations, samples are reduced in number as they are pulled by the operators on the right. Because of this, samples from original timeseries can be decoded and kept in memory in batches instead of being fully expanded.
+
+In addition to operators that have a one-to-one mapping with PromQL constructs, the Volcano model also describes so-called Exchange operators which can be used for flow control and optimizations, such as concurrency or batched selects. An example of an _Exchange_ operator is described in the [Intra-operator parallelism](#intra-operator-parallelism) section.
+
+## Inter-operator parallelism
+
+Since operators are independent and rely on a common interface for pulling data, they can be run in parallel to each other. As soon as one operator has processed data from an evaluation step, it can pass the result onward so that its upstream can immediatelly start working on it.
```
ahayworth, line 23:
```suggestion
Since operators are independent and rely on a common interface for pulling data, they can be run in parallel to each other. As soon as one operator has processed data from an evaluation step, it can pass the result onward so that its upstream can immediately start working on it.
```

README.md
ahayworth: As an end-user, "max samples" forces me to guess and experiment to find an acceptable limit that works with the resources I actually have. Personally, I tend to _under_ estimate heavily here, because I don't know precisely how the system will behave - better to have waste than to run out of resources, in my mind.

But that's not really optimal. As an end-user, what I _really_ want to do is say something like "use up to N Mb of memory" or "use no more than 75% of the available system memory." That allows me to maximize my resource usage with confidence (and cuts out the need to guess-and-check over and over again).

If possible, estimating the amount of memory used and expressing that as a limit would be preferable from my perspective.

README.md
```diff
@@ -2,4 +2,61 @@
...
+
+### Step vector allocations
+
+One challenge with the streamed execution model is knowing how much memory to allocate in each operator for each step. 
+
+To work around this issue, operators expose a `Series()` method which returns the labels for all time series that they will ever produce. Operators at the very bottom of the tree, like vector and matrix selectors, have this information since they are responsible for loading data from storage. Other operators can then call `Series()` on the downstream operator and pre-compute all possible outputs.
+
+Even thought this might look like an expensive operation, its cost is identical to just one evaluation step. Knowing sizes of input and output vectors also allows us to:   
+* Allocate memory very precisely by properly sizing vector pools (see section bellow),
+* Use arrays instead of maps for indexing data and performing lookups, leading to even further memory usage and faster execution time, and
```
saswatamcode, line 49: Does this mean that more memory is used, in lesser time?

saswatamcode (commented):
Amazing!

README.md
fpetkovski: Ouch, this should say `leading to even lower memory usage`. Will change!

README.md
fpetkovski: I agree, proxies like max samples and max series can be tricky to configure. Another argument for using memory limits directly is the fact that memory usage depends in part on the implementation itself. Users would somehow have to know that the engine got better or worse at managing memory and adjust max samples accordingly.

README.md
```diff
@@ -2,4 +2,61 @@
...
+Since operators are independent and rely on a common interface for pulling data, they can be run in parallel to each other. As soon as one operator has processed data from an evaluation step, it can pass the result onward so that its upstream can immediately start working on it.
+
+<p align="center">
+  <img src="./assets/promql-pipeline.png" />
+</p>
+
+
+## Intra-operator parallelism
+
+Parallelism can also be added within individual operators using a parallel coalesce exchange operator. Such exchange operators are indistinguishable from regular operators to their upstreams since they respect the same `Next()` interface.
```
metalmatze, line 32: Something specific to how PromQL might work: Are there any operators that need all the previous' step data to be there before it can ever be aggregated? Those would need to be treated slightly different. I can actually think of anything from the top of me head, it just happened a couple of times in FrostDB.

metalmatze (commented):
Love it! 

Maybe we can mention that for now it is intended to be used as a proxy that tries to come up with a better query plan and if it fails to falls back to the current engine.

README.md
bwplotka: This assumes function call is on lowest level?

README.md
bwplotka: > Also worth keeping in mind is that the result always has to have all series.

not really, what if we are grouping?

README.md
fpetkovski: AFAIK this is the only place where a function that returns multiple samples (like rate, sum_over_time, etc.) can be used.

README.md
fpetkovski: What I meant is that the result always has to have all series that would ever appear in the grouping, regardless of whether there is churn or not.

fpetkovski:
@metalmatze @bwplotka Thanks for the suggestions! I've added a roadmap section at the top which documents what's left to be done, and how we aim to roll out changes.

I will merge this for now and we can make further changes if needed.

bwplotka:
💪🏽

PR #7, opened 2022-09-15
Enabling fallback mode that uses prometheus PromQL engine for unsupported parts (can be disabled).

bwplotka:
This allows using it in Thanos behind some flag 💪🏽


fpetkovski (commented):
🚀 🚀 🚀 

I wonder if we can use a new struct `EngineWithFallback` that can do this for us, so that we keep the new engine a bit more independent. Not a blocker tough.

fpetkovski (approved):

PR #9, opened 2022-09-16
Adding min, max, count and avg over_time functions

sharadgaur:
Adding following functions:

- max_over_time
- min_over_time
- avg_over_time
- count_over_time

Related to issue https://github.com/thanos-community/promql-engine/issues/8


engine/existing_test.go
```diff
@@ -136,3 +136,339 @@ func TestRangeQuery(t *testing.T) {
 		})
 	}
 }
+
+func TestMinMaxOverTimeRangeQuery(t *testing.T) {
```
fpetkovski, line 140: Let's remove these additional tests, and keep only the ones in `engine_test`. It should be sufficient to have comparisons with the old engine only.

fpetkovski (approved):

PR #12, opened 2022-09-18
Add support for instant queries

saswatamcode:
This PR enables support for instant queries with existing implemented operators and functions.
Also adds new benchmark and testcases.

Not super familiar with the query side of things, so could be getting something wrong. 🙂


engine/instant_query.go
```diff
@@ -15,32 +17,82 @@ import (
...
+				Metric: series[i].Metric,
+				Point: promql.Point{
+					V: series[i].Points[0].V,
+					T: q.ts.UnixMilli(),
+				},
+			}
+		}
+		result = vector
+	case parser.ValueTypeString:
+		result = promql.Scalar{V: series[0].Points[0].V, T: q.ts.UnixMilli()}
```
yeya24, line 88: This should be `promql.String`?
And we also need to handle the scalar case with `parser.ValueTypeScalar`

physicalplan/scan/matrix_selector.go
```diff
@@ -96,7 +96,11 @@ func (o *matrixSelector) Next(ctx context.Context) ([]model.StepVector, error) {
 		return nil, err
 	}
 
-	totalSteps := (o.maxt+o.mint)/o.step + 1
+	// Instant evaluation is executed as a range evaluation with one step.
+	totalSteps := int64(1)
+	if o.step != 0 {
+		totalSteps = (o.maxt+o.mint)/o.step + 1
```
yeya24, line 102: I guess it should be `(o.maxt-o.mint)`?

physicalplan/scan/vector_selector.go
```diff
@@ -78,7 +78,12 @@ func (o *vectorSelector) Next(ctx context.Context) ([]model.StepVector, error) {
 		return nil, err
 	}
 
-	totalSteps := (o.maxt+o.mint)/o.step + 1
+	// Instant evaluation is executed as a range evaluation with one step.
+	totalSteps := int64(1)
+	if o.step != 0 {
+		totalSteps = (o.maxt+o.mint)/o.step + 1
```
yeya24, line 84: ditto

engine/instant_query.go
saswatamcode: Nice catch! I was planning to handle the parser.ValueTypeScalar here. Handling string might need a small change in plan.go. Will implement! 🙂

physicalplan/scan/matrix_selector.go
yeya24: How is this resolved?

saswatamcode: Sorry about that, I forgot to actually push. 😅

.gitignore
```diff
@@ -11,3 +11,7 @@ examples/tmp/
 
 # Ignore the MacOS Trash (DS-Store)
 .DS_Store
+
+# Ignore benchmarking output
+benchmarks/
+engine.test
```
fpetkovski, line 17:
```suggestion
engine.test

```

Makefile
```diff
@@ -7,9 +7,12 @@ help: ## Displays help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n\nTargets:\n"} /^[a-z0-9A-Z_-]+:.*?##/ { printf "  \033[36m%-10s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
 
 .PHONY: test
+test: ## Runs all Go unit tests.
+export GOCACHE=/tmp/cache
```
fpetkovski, line 11: Hm, isn't this file autogenerated by bingo?

engine/instant_query.go
```diff
@@ -15,32 +17,89 @@ import (
...
+	for _, vector := range vs {
+		for _, sample := range vector.Samples {
+			if len(series[sample.ID].Points) == 0 {
+				series[sample.ID].Points = make([]promql.Point, 0, 121)
+			}
+			series[sample.ID].Points = append(series[sample.ID].Points, promql.Point{
+				T: vector.T,
+				V: sample.V,
+			})
+			q.plan.GetPool().PutSamples(vector.Samples)
```
fpetkovski, line 70: Whoops, this should be one line bellow.

engine/instant_query.go
```diff
@@ -15,32 +17,89 @@ import (
...
+
+	series := make([]promql.Series, len(resultSeries))
+	for i := 0; i < len(resultSeries); i++ {
+		series[i].Metric = resultSeries[i]
+	}
+
+	for _, vector := range vs {
+		for _, sample := range vector.Samples {
+			if len(series[sample.ID].Points) == 0 {
+				series[sample.ID].Points = make([]promql.Point, 0, 121)
```
fpetkovski, line 64: We can allocate one point only for instant queries.

physicalplan/plan.go
```diff
@@ -70,6 +70,8 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
 		default:
 			return nil, errors.Wrapf(ErrNotSupportedExpr, "got: %s", t)
 		}
+	case *parser.StringLiteral:
+		return nil, nil
```
fpetkovski, line 74: Hm, what does this do?

Makefile
saswatamcode: This target does not use bingo in any way. This is just to always run tests and not cache results. 🙂

physicalplan/plan.go
saswatamcode: Without this, any expression of type `StringLiteral` would error as "unsupported expression". 

Ideally, we could handle this case right here by returning an operator that returns `StepVector` with `StepSample` of type string, but that needs some changes to the `model` types. This can be iterated on later! So not complicating things for now and checking this in `instantQuery.Exec` instead. 🙂

fpetkovski: I see, but why do we need to handle StringLiteral for instant queries?

engine/instant_query.go
```diff
@@ -15,32 +17,89 @@ import (
...
+
+	series := make([]promql.Series, len(resultSeries))
+	for i := 0; i < len(resultSeries); i++ {
+		series[i].Metric = resultSeries[i]
+	}
+
+	for _, vector := range vs {
+		for _, sample := range vector.Samples {
+			if len(series[sample.ID].Points) == 0 {
+				series[sample.ID].Points = make([]promql.Point, 0, 1)
```
fpetkovski, line 64: Sorry one more change here, can we move this bellow line 58 and avoid the `if` condition altogether?

physicalplan/plan.go
saswatamcode: Only instant queries can return strings as per docs: https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries and PromQL has StringLiteral support: https://prometheus.io/docs/prometheus/latest/querying/basics/#string-literals.

This is also kind of how they implement it: https://github.com/prometheus/prometheus/blob/release-2.38/promql/engine.go#L1546 so tried a similar pattern to support it here 🙂

fpetkovski: Got it, thanks for the explanation.

fpetkovski (approved):

PR #13, opened 2022-09-18
Add vectorized operators

fpetkovski:
When aggregating without grouping labels, the aggregation can be done with vectorized assembly operators on the entire step vector.

This commit introduces two aggregation tables as substitute for the old one: scalarTable and vectorizedTable. 
The scalar table uses go code, while the vectorizedTable uses the gonum [floats](https://pkg.go.dev/gonum.org/v1/gonum@v0.12.0/floats) package to execute vectorized aggregations.

Fixes https://github.com/thanos-community/promql-engine/issues/4

<pre>
benchstat old.out new.out 
name                                        old time/op    new time/op    delta
OldEngine/sum/current_engine-4                 123ms ± 0%     122ms ± 0%   ~     (p=1.000 n=1+1)
<strong>OldEngine/sum/new_engine-4                    67.1ms ± 0%    48.9ms ± 0%   ~     (p=1.000 n=1+1)</strong>
OldEngine/sum_rate/current_engine-4            219ms ± 0%     201ms ± 0%   ~     (p=1.000 n=1+1)
<strong>OldEngine/sum_rate/new_engine-4                157ms ± 0%      97ms ± 0%   ~     (p=1.000 n=1+1)</strong>

name                                        old alloc/op   new alloc/op   delta
OldEngine/sum/current_engine-4                4.94MB ± 0%    4.94MB ± 0%   ~     (p=1.000 n=1+1)
<strong>OldEngine/sum/new_engine-4                    18.5MB ± 0%    10.0MB ± 0%   ~     (p=1.000 n=1+1)</strong>
OldEngine/sum_rate/current_engine-4           6.01MB ± 0%    6.01MB ± 0%   ~     (p=1.000 n=1+1)
<strong>OldEngine/sum_rate/new_engine-4               19.0MB ± 0%    12.0MB ± 0%   ~     (p=1.000 n=1+1)</strong>

name                                        old allocs/op  new allocs/op  delta
OldEngine/sum/current_engine-4                 73.8k ± 0%     73.8k ± 0%   ~     (all equal)
OldEngine/sum/new_engine-4                     71.6k ± 0%     73.5k ± 0%   ~     (p=1.000 n=1+1)
OldEngine/sum_rate/current_engine-4            82.9k ± 0%     82.9k ± 0%   ~     (p=1.000 n=1+1)
OldEngine/sum_rate/new_engine-4                98.5k ± 0%    100.4k ± 0%   ~     (p=1.000 n=1+1)
</pre>

### DIFF
--- a/physicalplan/aggregate/hashaggregate.go
+++ b/physicalplan/aggregate/hashaggregate.go
@@ -27,7 +27,7 @@ type aggregate struct {
 	aggregation parser.ItemType
 
 	once   sync.Once
-	tables []*aggregateTable
+	tables []aggregateTable
 	series []labels.Labels
 
 	stepsBatch int
@@ -59,7 +59,7 @@ func NewHashAggregate(
 
 func (a *aggregate) Series(ctx context.Context) ([]labels.Labels, error) {
 	var err error
-	a.once.Do(func() { err = a.initOutputBuffers(ctx) })
+	a.once.Do(func() { err = a.initializeTables(ctx) })
 	if err != nil {
 		return nil, err
 	}
@@ -74,6 +74,7 @@ func (a *aggregate) GetPool() *model.VectorPool {
 func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 	in, err := a.next.Next(ctx)
 	if err != nil {
+		a.workers.Shutdown()
 		return nil, err
 	}
 	if in == nil {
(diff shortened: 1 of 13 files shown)


physicalplan/aggregate/hashaggregate.go
```diff
@@ -82,95 +83,76 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+		a.series = []labels.Labels{{}}
+		a.vectorPool.SetStepSize(1)
+		return nil
+	}
 
+	inputCache := make([]uint64, len(series))
+	outputMap := make(map[uint64]*model.Series)
+	outputCache := make([]*model.Series, 0)
 	for i := 0; i < len(series); i++ {
 		buf := make([]byte, 128)
```
yeya24, line 127: Can we move the buf out of the for loop? To reuse the buffer

physicalplan/aggregate/vector_table.go
```diff
@@ -0,0 +1,85 @@
...
+	case "count":
+		return func(float64s []float64) float64 {
+			return float64(len(float64s))
+		}, nil
+	case "avg":
+		return func(in []float64) float64 {
+			return floats.Sum(in) / float64(len(in))
+		}, nil
+	}
+	return nil, fmt.Errorf("unknown aggregation function %s", t)
```
yeya24, line 84: For other types of aggregations like `topk`, `quantile`, etc. I don't see `gonum` has a good vectorization support for them. How would you like to handle those cases?

fpetkovski: Actually there is some support for [quantile](https://pkg.go.dev/gonum.org/v1/gonum/stat), but we should check if it's compatible with the Prometheus implementation. In the worst case, we can just use scalar computation like we do when we have group by expressions. Probably we can just port the code from the Prometheus engine.

physicalplan/aggregate/hashaggregate.go
fpetkovski: Nice catch, updated.

physicalplan/aggregate/vector_table.go
fpetkovski: For `topk`, maybe we can use [SortWeighted](https://pkg.go.dev/gonum.org/v1/gonum/stat#SortWeighted). We can benchmark it to see how it performs compared to the current `topk` implementation in Prometheus.

yeya24: Right now we are just doing some optimization on aggr within the same series. It is just an optimization rather than vectorization.
Better vectorization should work like processing multiple series at the same time like calculating multiple series' max in a batching way.

fpetkovski: The aggregation works on a step vector, which has samples from all time series at a certain time stamp. So when we call floats.Sum(), that sums samples across all timeseries. Or did you have something else in mind?

fpetkovski: Did you mean that we should find a way to apply vectorization to a `sum by` aggregation so that all resulting series are calculated in vectorized manner?

Issue #15, opened 2022-09-20
Add support for NumberLiterals


saswatamcode:
Currently, we cannot use any numbers in queries, so common expressions fail. 
Also, this would be a common use case with binary operators in #14.

Would love to pick this up! 🙂

fpetkovski:
Thanks, this is definitely something that should be done soon. Let me just finish #14 and then you can pick this up. I am not sure if there are use cases for number literals outside of the binary operator.

saswatamcode:
I already kind of started working on it 🙈 in https://github.com/thanos-community/promql-engine/pull/16.
Should not block the binary operator stuff in any way tho.

fpetkovski:
Ok, that also works :)

PR #16, opened 2022-09-20
Add support for number literals

saswatamcode:
This PR adds support for number literals in the new engine (for both instant and range queries) as mentioned in #15.
Also adds `vector` function to make it easier to work with in query expressions.

### DIFF
--- a/physicalplan/plan.go
+++ b/physicalplan/plan.go
@@ -69,6 +69,8 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
 		default:
 			return nil, errors.Wrapf(ErrNotSupportedExpr, "got: %s", t)
 		}
+	case *parser.NumberLiteral:
+		return scan.NewLiteralSelector(model.NewVectorPool(stepsBatch), mint, maxt, step, stepsBatch, e.Val), nil
 	case *parser.StringLiteral:
 		return nil, nil
 	default:
--- a/physicalplan/plan.go
+++ b/physicalplan/plan.go
@@ -70,7 +70,7 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
 			return nil, errors.Wrapf(ErrNotSupportedExpr, "got: %s", t)
 		}
 	case *parser.NumberLiteral:
-		return scan.NewLiteralSelector(model.NewVectorPool(stepsBatch), mint, maxt, step, stepsBatch, e.Val), nil
+		return scan.NewNumberLiteralSelector(model.NewVectorPool(stepsBatch), mint, maxt, step, stepsBatch, e.Val), nil
 	case *parser.StringLiteral:
 		return nil, nil
 	default:
--- a/physicalplan/plan.go
+++ b/physicalplan/plan.go
@@ -66,11 +66,18 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
 				operators = append(operators, exchange.NewConcurrent(selector, 2))
 			}
 			return exchange.NewCoalesce(model.NewVectorPool(stepsBatch), operators...), nil
+		case *parser.NumberLiteral:
+			call, err := scan.NewFunctionCall(e.Func, step)
+			if err != nil {
+				return nil, err
+			}
+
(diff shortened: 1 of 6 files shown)


physicalplan/scan/literal_selector.go
```diff
@@ -0,0 +1,78 @@
...
+	"math"
+	"time"
+
+	"github.com/thanos-community/promql-engine/physicalplan/model"
+
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+// literalSelector returns []model.StepVector with same sample value across time range.
+type literalSelector struct {
```
yeya24, line 17: Probably rename it to `numberLiteralSelector` as it is not string literal?

physicalplan/scan/literal_selector.go
```diff
@@ -0,0 +1,97 @@
...
+	return o.vectorPool
+}
+
+func (o *numberLiteralSelector) Next(ctx context.Context) ([]model.StepVector, error) {
+	if o.currentStep > o.maxt {
+		return nil, nil
+	}
+
+	totalSteps := int64(1)
+	if o.step != 0 {
```
fpetkovski, line 62: I wonder if we can extract this logic somewhere so we don't have to repeat it. My initial idea was to add a `NumSteps()` method to the `query.Options`, and inject an instance in relevant scanners.

saswatamcode: Yup! That seems good to me too, but seems like we don't use `query.Options` anywhere else. Let me try to refactor this a bit in the next PR! 🙂

physicalplan/plan.go
```diff
@@ -66,9 +66,18 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
 				operators = append(operators, exchange.NewConcurrent(selector, 2))
 			}
 			return exchange.NewCoalesce(model.NewVectorPool(stepsBatch), operators...), nil
+		case *parser.NumberLiteral:
```
fpetkovski, line 69: Do we have an example of when this can happen?

physicalplan/scan/literal_selector.go
```diff
@@ -0,0 +1,97 @@
...
+		totalSteps = (o.maxt-o.mint)/o.step + 1
+	}
+
+	numSteps := int(math.Min(float64(o.stepsBatch), float64(totalSteps)))
+
+	vectors := o.vectorPool.GetVectorBatch()
+	ts := o.currentStep
+
```
fpetkovski, line 70: We can organize the variables a bit:
```suggestion
	vectors := o.vectorPool.GetVectorBatch()
	numSteps := int(math.Min(float64(o.stepsBatch), float64(totalSteps)))
	ts := o.currentStep
```

physicalplan/scan/literal_selector.go
```diff
@@ -0,0 +1,97 @@
...
+		result := promql.Sample{
+			Point: promql.Point{
+				T: ts,
+				V: o.val,
+			},
+		}
+
+		if o.call != nil {
+			result = o.call(labels.New(), []promql.Point{result.Point}, time.UnixMilli(ts))
```
fpetkovski, line 84: Can we use the existing labels initialized for the Series call instead of allocating a new slice?

fpetkovski (commented):
Just a few more nits :)

physicalplan/plan.go
saswatamcode: Yup, the example can be anything like `vector(1)` or PromQL functions which have scalar input. This is also present as a testcase. 🙂

physicalplan/scan/function.go
```diff
@@ -68,6 +68,13 @@ func NewFunctionCall(f *parser.Function, selectRange time.Duration) (FunctionCal
 				Metric: labels,
 			}
 		}, nil
+	case "vector":
+		return func(_ labels.Labels, points []promql.Point, stepTime time.Time) promql.Sample {
```
fpetkovski, line 72: Should the labels be reused instead of allocating new ones?

saswatamcode: Nice catch! Missed updating this.

fpetkovski (approved):
Thanks, this looks good 👍

yeya24 (approved):
Looks good!

PR #18, opened 2022-09-20
Implement binary operator between two vectors

fpetkovski:
This commit adds support for binary operations between two vectors. Support for scalar operands and group_left/group_right labels will be added in a follow up PR.

I also tried this in one of our envs, seems to be holding up well so far.
Cannot guarantee lack of dragons.

Benchmark 
```
BenchmarkOldEngineRange
BenchmarkOldEngineRange/binary_operation_with_many_to_one
BenchmarkOldEngineRange/binary_operation_with_many_to_one/current_engine
BenchmarkOldEngineRange/binary_operation_with_many_to_one/current_engine-8         	       3	 381090695 ns/op	63308880 B/op	  594486 allocs/op
BenchmarkOldEngineRange/binary_operation_with_many_to_one/new_engine
BenchmarkOldEngineRange/binary_operation_with_many_to_one/new_engine-8             	      50	  23757462 ns/op	32150253 B/op	  131289 allocs/op
```

cc @yeya24

### DIFF
--- a/physicalplan/binary/table.go
+++ b/physicalplan/binary/table.go
@@ -1,0 +1,128 @@
+package binary
+
+import (
+	"fmt"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/thanos-community/promql-engine/physicalplan/model"
+)
+
+type sample struct {
+	t int64
+	v float64
+}
+
+type table struct {
+	pool *model.VectorPool
+
+	operation operation
+	card      parser.VectorMatchCardinality
+
+	outputValues        []sample
+	highCardOutputCache []*uint64
+	lowCardOutputCache  [][]uint64
+}
+
+func newTable(
+	pool *model.VectorPool,
+	card parser.VectorMatchCardinality,
+	expr parser.ItemType,
+	outputValues []sample,
+	highCardOutputCache []*uint64,
+	lowCardOutputCache [][]uint64,
+) (*table, error) {
+	op, err := newOperation(expr)
+	if err != nil {
+		return nil, err
+	}
+	return &table{
+		pool: pool,
+		card: card,
+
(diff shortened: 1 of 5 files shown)


physicalplan/binary/table.go
```diff
@@ -0,0 +1,128 @@
+package binary
+
+import (
+	"fmt"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/thanos-community/promql-engine/physicalplan/model"
+)
+
+type sample struct {
```
saswatamcode, line 10: Is there a reason for not using `promql.Point`? 🙂

physicalplan/binary/table.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return step
+}
+
+type operation func(lhs float64, rhs float64) float64
+
+func newOperation(expr parser.ItemType) (operation, error) {
+	t := parser.ItemTypeStr[expr]
+	switch t {
+	case "+":
```
saswatamcode, line 109: Maybe we can use a `map[string]operation` here, to extend this later on?

saswatamcode (approved):
Thanks! This looks awesome!

Small questions! :)

physicalplan/binary/table.go
fpetkovski: Good idea, this makes it look much cleaner 👍

physicalplan/binary/table.go
fpetkovski: Hm, no good reason, but let's maybe keep the package self-contained. It's a small struct anyway.

saswatamcode (approved):
Thanks, this looks good to me! 🌟

saswatamcode (approved):

PR #20, opened 2022-09-21
Handle edge cases & fix panic

### LINKED ISSUE #17: Instant query with no data causes a panic
Looks like running an instant query against an empty series set series can cause a panic
```
panic: runtime error: index out of range [4] with length 0

goroutine 348 [running]:
github.com/thanos-community/promql-engine/physicalplan/aggregate.(*aggregate).workerTask(0x100e48a40?, 0x14000091668?, {0x0, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}})
	/Users/fpetkovski/Projects/promql-engine/physicalplan/aggregate/hashaggregate.go:155 +0x11c
github.com/thanos-community/promql-engine/worker.(*Worker).Start(0x14000eacc80)
	/Users/fpetkovski/Projects/promql-engine/worker/worker.go:65 +0xdc
created by github.com/thanos-community/promql-engine/worker.Group.Start
	/Users/fpetkovski/Projects/promql-engine/worker/worker.go:24 +0x30
```

saswatamcode:
Fixes #17. Also adds test cases for a couple of other edge cases, to align behavior with the old engine.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -261,6 +261,23 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 				foo{method="get", code="404"} 1+2.2x20`,
 			query: `1 - (100 * sum(foo{method="get"}) / sum(foo))`,
 		},
+		{
+			name:  "empty series",
+			load:  "",
+			query: "http_requests_total",
+		},
+		{
+			name:  "empty series with func",
+			load:  "",
+			query: "sum(http_requests_total)",
+		},
+		{
+			name: "empty result",
+			load: `load 30s
+					http_requests_total{pod="nginx-1"} 1+1x15
+					http_requests_total{pod="nginx-2"} 1+2x18`,
+			query: `http_requests_total{pod="nginx-3"}`,
+		},
 	}
 
 	for _, tc := range cases {
@@ -424,6 +441,23 @@ func TestInstantQuery(t *testing.T) {
 				foo{method="get", code="500"} 1+1.1x30
 				foo{method="get", code="404"} 1+2.2x20`,
 			query: `1 - (100 * sum(foo{method="get"}) / sum(foo))`,
+		},
+		{
+			name:  "empty series",
+			load:  "",
+			query: "http_requests_total",
+		},
+		{
+			name:  "empty series with func",
+			load:  "",
+			query: "sum(http_requests_total)",
+		},
+		{
+			name: "empty result",
+			load: `load 30s
+					http_requests_total{pod="nginx-1"} 1+1x15
+					http_requests_total{pod="nginx-2"} 1+2x18`,
+			query: `http_requests_total{pod="nginx-3"}`,
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


physicalplan/aggregate/hashaggregate.go
```diff
@@ -152,7 +152,11 @@ func (a *aggregate) initializeTables(ctx context.Context) error {
 }
 
 func (a *aggregate) workerTask(workerID int, vector model.StepVector) model.StepVector {
-	table := a.tables[workerID]
-	table.aggregate(vector)
-	return table.toVector(a.vectorPool)
+	if len(a.tables) > workerID {
```
fpetkovski, line 155: Hm, what happens if this condition is false? Does it mean there's nothing to evaluate? I wonder if the edge case should be prevented somewhere before calling this method 🤔

saswatamcode: So taking a deeper look here. For the HashAggregate operator, this panic only occurs when `Next()` is called before `Series()` with empty data. This was detected from instant queries as the `Exec` method calls them in this order but range query `Exec` does not.

So when we call `Next()` before `Series()`, we have a data race, as some of the workers have started executing their tasks by reading from the aggregate tables slice, but `initializeTables` has not been called yet. The `if` here, in this case, is actually wrong.

Should `Series()` always be called before `Next()`?

fpetkovski: Hm, but don't we initialize tables [before](https://github.com/thanos-community/promql-engine/blob/main/physicalplan/aggregate/hashaggregate.go#L86) we start [sending](https://github.com/thanos-community/promql-engine/blob/main/physicalplan/aggregate/hashaggregate.go#L94) tasks to workers?

saswatamcode: Yeah I thought so too, but the panic points to the table array being read in `workerTask` as out of range. There might be some race within `worker.Send` and `worker.Start` (this seems to execute the `workerTask` via `doWork`). I'll try to look into this a bit more today

saswatamcode: Ok so, it was actually the `Shutdown` method. It releases the `next` lock which allows `Start` to acquire it and execute `doWork` i.e `aggregate.workerTask`. In hashaggregate `Next()` if we don't get data from `a.next.Next()`  we call `Shutdown` on workers [here](https://github.com/thanos-community/promql-engine/blob/main/physicalplan/aggregate/hashaggregate.go#L81), which happened in case of empty series, and this allowed `Start` to execute `doWork` and cause panics. I fixed this, and also added comments about how the locks are used for syncing across worker methods! 🙂

fpetkovski (approved):
Thanks for this PR @saswatamcode, I think this is a good short-term fix. 

In general I am not a big fan of inlined comments, and this just tells me that we need to prioritize https://github.com/thanos-community/promql-engine/issues/19 :)

Let's merge this as is so that we resolve the issue we have.

PR #21, opened 2022-09-21
Implement binary operator with scalar operands

fpetkovski:
This commit adds support for the binary operator for cases where one operand is a scalar.

Benchmark:
```
BenchmarkOldEngineRange/binary_operation_with_vector_and_scalar
BenchmarkOldEngineRange/binary_operation_with_vector_and_scalar/current_engine
BenchmarkOldEngineRange/binary_operation_with_vector_and_scalar/current_engine-8   	       5	 234163942 ns/op	30723304 B/op	   84369 allocs/op
BenchmarkOldEngineRange/binary_operation_with_vector_and_scalar/new_engine
BenchmarkOldEngineRange/binary_operation_with_vector_and_scalar/new_engine-8       	      72	  18371273 ns/op	29613023 B/op	   89369 allocs/op
```

### DIFF
--- a/physicalplan/binary/scalar.go
+++ b/physicalplan/binary/scalar.go
@@ -1,0 +1,120 @@
+package binary
+
+import (
+	"context"
+	"sync"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/thanos-community/promql-engine/physicalplan/model"
+)
+
+// scalarOperator evaluates expressions where one operand is a scalarOperator.
+type scalarOperator struct {
+	seriesOnce sync.Once
+	series     []labels.Labels
+	scalar     float64
+
+	pool           *model.VectorPool
+	numberSelector model.VectorOperator
+	next           model.VectorOperator
+	getOperands    getOperandsFunc
+	operation      operation
+}
+
+func NewScalar(pool *model.VectorPool, next model.VectorOperator, numberSelector model.VectorOperator, op parser.ItemType, scalarSideLeft bool) (*scalarOperator, error) {
+	binaryOperation, err := newOperation(op)
+	if err != nil {
+		return nil, err
+	}
+	var getOperands getOperandsFunc
+	if scalarSideLeft {
+		getOperands = getOperandsScalarLeft
+	} else {
+		getOperands = getOperandsScalarRight
+	}
+
+	// Cache the result of the number selector since it
+	// will not change during execution.
+	v, err := numberSelector.Next(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	scalar := v[0].Samples[0]
+
+	return &scalarOperator{
+		pool:           pool,
+		next:           next,
+		scalar:         scalar,
+		numberSelector: numberSelector,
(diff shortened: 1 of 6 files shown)


physicalplan/binary/scalar.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, nil
+	}
+	o.seriesOnce.Do(func() { err = o.loadSeries(ctx) })
+	if err != nil {
+		return nil, err
+	}
+
+	out := o.pool.GetVectorBatch()
+	for _, vector := range in {
+		step := o.pool.GetStepVector(vector.T)
```
fpetkovski, line 79: Ideally we should reuse the same vector we get from the input, and modify the values in place.
But for that, we need to have a better way of recycling step vectors so that we know in which pool a vector needs to be returned. 

Maybe the step vector itself can have a `Recycle()` method which will put it back into the correct pool.

physicalplan/binary/scalar.go
```diff
@@ -0,0 +1,120 @@
...
+	binaryOperation, err := newOperation(op)
+	if err != nil {
+		return nil, err
+	}
+	var getOperands getOperandsFunc
+	if scalarSideLeft {
+		getOperands = getOperandsScalarLeft
+	} else {
+		getOperands = getOperandsScalarRight
+	}
```
saswatamcode, line 35:
```suggestion
	getOperands := getOperandsScalarRight
	if scalarSideLeft {
		getOperands = getOperandsScalarLeft
	} 
```

physicalplan/plan.go
```diff
@@ -83,24 +83,53 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint, maxt time.Ti
...
+	vector, err := newOperator(vectorExpr, storage, mint, maxt, step)
+	if err != nil {
+		return nil, err
+	}
+	scalar, err := newOperator(scalarExpr, storage, mint, maxt, step)
+	if err != nil {
+		return nil, err
+	}
+
+	return binary.NewScalar(model.NewVectorPool(stepsBatch), vector, scalar, e.Op, scalarSideLeft)
```
saswatamcode, line 134: Maybe we can simplify a bit here? 🙂 
```suggestion
	lhs, err := newOperator(e.LHS, storage, mint, maxt, step)
	if err != nil {
		return nil, err
	}
	rhs, err := newOperator(e.RHS, storage, mint, maxt, step)
	if err != nil {
		return nil, err
	}
	
	if e.LHS.Type() == parser.ValueTypeScalar {
		return binary.NewScalar(model.NewVectorPool(stepsBatch), rhs, lhs, e.Op, true)
	}
	return binary.NewScalar(model.NewVectorPool(stepsBatch), lhs, rhs, e.Op, false)
```

saswatamcode (approved):
Thanks for awesome work!

Some small suggestions.

physicalplan/binary/scalar.go
saswatamcode: Yup, this seems like a great idea! This could be iterated on later as well.

saswatamcode (approved):
Thanks!

PR #26, opened 2022-09-25
Improve workers package

fpetkovski:
The workers package uses mutexes for synchronization. This approach was put in place due to concerns related to channel performormance. However, the use of mutexes can leak resources and requires much more careful use of the package.

This commit replaces mutexes with channels and injects a cancellable context when starting a worker group. This way workers will be properly terminated when execution of a query is finished.

Fixes https://github.com/thanos-community/promql-engine/issues/19

### DIFF
--- a/physicalplan/aggregate/hashaggregate.go
+++ b/physicalplan/aggregate/hashaggregate.go
@@ -52,7 +52,6 @@ func NewHashAggregate(
 		stepsBatch:  stepsBatch,
 	}
 	a.workers = worker.NewGroup(stepsBatch, a.workerTask)
-	a.workers.Start()
 
 	return a, nil
 }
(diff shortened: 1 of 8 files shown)


physicalplan/scan/literal_selector.go
```diff
@@ -63,6 +63,8 @@ func (o *numberLiteralSelector) Next(ctx context.Context) ([]model.StepVector, e
 	totalSteps := int64(1)
 	if o.step != 0 {
 		totalSteps = (o.maxt-o.mint)/o.step + 1
+	} else {
+		o.step = 1
```
saswatamcode, line 67: Not sure I understand what this is doing for all the selector `Next` methods. Is it to prevent further divide by zero errors in the future? 🙂

physicalplan/aggregate/hashaggregate.go
```diff
@@ -109,6 +104,8 @@ func (a *aggregate) initializeTables(ctx context.Context) error {
 		return err
 	}
 
+	a.workers.Start(ctx)
```
saswatamcode, line 107: Starting the group here makes more sense! 💪🏻

saswatamcode (approved):
Thanks! This makes the worker package much more intuitive to use. 🌟 

One small question!

physicalplan/scan/literal_selector.go
fpetkovski: We have to set the step to a non-zero value to make sure the operator terminates for instant queries. Otherwise we won't make any changes to the currentStep here:  https://github.com/thanos-community/promql-engine/blob/main/physicalplan/scan/vector_selector.go#L109

saswatamcode: Thanks!

saswatamcode (approved):

PR #29, opened 2022-09-26
Extend fallback support

fpetkovski:
This commit adds the ErrNotSupportedExpr to unsupported aggregaion and functions, so that we can properly fall back to the old engine.

It also lints the codebase with goimports.


saswatamcode (approved):
💪🏻

PR #31, opened 2022-09-26
Remove aggregate panic

fpetkovski:
We currently panic in the hash aggregate function when an operation
is not supported.
    
This commit removes the panic and replaces it with an error, so that we can fall back to the old engine. 
I had to refactor the codebase a bit in order to enable the change.

### DIFF
--- a/physicalplan/aggregate/scalar_table.go
+++ b/physicalplan/aggregate/scalar_table.go
@@ -28,24 +28,18 @@ type scalarTable struct {
 	accumulators []*accumulator
 }
 
-func newScalarTables(stepsBatch int, inputCache []uint64, outputCache []*model.Series, a parser.ItemType) []aggregateTable {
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
LGTM!

physicalplan/aggregate/scalar_table.go
```diff
@@ -118,81 +112,100 @@ type accumulator struct {
...
-				value = 0
-			},
+		return func() *accumulator {
+			var value float64
+			var hasValue bool
+
+			return &accumulator{
+				AddFunc: func(v float64) {
+					hasValue = true
+					value = math.Max(value, v)
```
yeya24, line 144: We need to check `hasValue` in case v < 0 ? Or initialize value as the min float64

fpetkovski: I think you're right. Will create an issue for this 👍

fpetkovski: Created https://github.com/thanos-community/promql-engine/issues/37

PR #36, opened 2022-09-27
support last_over_time function

yeya24:

This pr adds support for `last_over_time` function. 
One thing different is that `last_over_time` requires the metric name to be kept in the final vector result. So I just simply pass the `parser.Call` to the `NewMatrixSelector` to check whether to drop the metric name or not.

### DIFF
--- a/physicalplan/scan/function.go
+++ b/physicalplan/scan/function.go
@@ -70,6 +70,16 @@ func NewFunctionCall(f *parser.Function, selectRange time.Duration) (FunctionCal
 				Metric: labels,
 			}
 		}, nil
+	case "last_over_time":
+		return func(labels labels.Labels, points []promql.Point, stepTime time.Time) promql.Sample {
+			return promql.Sample{
+				Point: promql.Point{
+					T: stepTime.UnixMilli(),
+					V: points[len(points)-1].V,
+				},
+				Metric: labels,
+			}
+		}, nil
 	case "rate":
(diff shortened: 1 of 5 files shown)


fpetkovski (approved):
Thanks @yeya24 👍

physicalplan/scan/function.go
```diff
@@ -70,6 +70,16 @@ func NewFunctionCall(f *parser.Function, selectRange time.Duration) (FunctionCal
 				Metric: labels,
 			}
 		}, nil
+	case "last_over_time":
+		return func(labels labels.Labels, points []promql.Point, stepTime time.Time) promql.Sample {
+			return promql.Sample{
+				Point: promql.Point{
+					T: stepTime.UnixMilli(),
+					V: points[len(points)-1].V,
```
fpetkovski, line 78: Do we need to check for out of bounds here? Will `points` always be a slice with at least one element?

yeya24: Good question, I think that's something we need to handle.

yeya24: Umm, is it possible to have the series with labels only but no sample? I guess no so this should be safe. If no series then the func won't be called

fpetkovski: Maybe it can happen if there are gaps in the values, but I also need to double check if this is a real problem.

PR #43, opened 2022-09-27
Adding stddev and stdvar aggregation operations

sharadgaur:
Issue: https://github.com/thanos-community/promql-engine/issues/41

### DIFF
--- a/physicalplan/scan/function.go
+++ b/physicalplan/scan/function.go
@@ -344,7 +344,7 @@ func avgOverTime(points []promql.Point) float64 {
 				continue
 			}
 		}
-		mean, c = kahanSumInc(v.V/count-mean/count, mean, c)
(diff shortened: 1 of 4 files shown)


physicalplan/scan/function.go
```diff
@@ -455,7 +455,7 @@ func linearRegression(samples []promql.Point, interceptTime int64) (slope, inter
 	return slope, intercept
 }
 
-func kahanSumInc(inc, sum, c float64) (newSum, newC float64) {
+func KahanSumInc(inc, sum, c float64) (newSum, newC float64) {
```
sharadgaur, line 458: Should I move this func somewhere outside or keep it here ?

yeya24: I think it is okay to expose it for now. If needed we can extract the function outside in the future

yeya24 (approved):

PR #44, opened 2022-09-28
Use GOMAXPROCS and add benchmarks

fpetkovski:
We use runtime.NumCPU to determine parallelism, which can create issues in containerized environments.

This commit changes the planner to use `GOMAXPROCS` instead. It also adds information from the latest benchmarks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,65 +80,80 @@ The current implementation uses goroutines very liberally which means the query 
 
 The current implementation creates a physical plan directly from the PromQL abstract syntax tree. Plan optimizations not yet implemented and would require having a logical plan as an intermediary step.
 
-
 ## Latest benchmarks
+
+These are the latest benchmarks captured on an Apple M1 Pro processor.
+
+Note that memory usage is higher when executing a query with parallelism greater than 1. This is due to the fact that the engine is able to execute multiple operations at once (e.g. decode chunks from multiple series at the same time), which requires using independent buffers for each parallel operation. 
 
 Single core benchmarks
 ```markdown
-vector_selector/current_engine         	      44	  27756507 ns/op	16146356 B/op	   69121 allocs/op
-vector_selector/new_engine             	      32	  36515120 ns/op	25709769 B/op	   76738 allocs/op
+name                                                old time/op    new time/op    delta
+RangeQuery/vector_selector                            28.1ms ± 2%    34.2ms ± 4%  +21.80%  (p=0.000 n=9+10)
+RangeQuery/sum                                        46.4ms ± 4%    27.5ms ± 0%  -40.59%  (p=0.000 n=10+10)
+RangeQuery/sum_by_pod                                  134ms ± 2%      37ms ± 0%  -72.72%  (p=0.000 n=10+8)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -79,3 +79,66 @@ The current implementation uses goroutines very liberally which means the query
 ### Plan optimization
 
 The current implementation creates a physical plan directly from the PromQL abstract syntax tree. Plan optimizations not yet implemented and would require having a logical plan as an intermediary step.
+
+
+## Latest benchmarks
+
+Single core benchmarks
+```markdown
+vector_selector/current_engine         	      44	  27756507 ns/op	16146356 B/op	   69121 allocs/op
```
saswatamcode, line 88: Maybe we can use `benchstat` and make these figures more readable both in rendered + unrendered markdown? 🙂

saswatamcode (approved):
Thanks! 

Small suggestion!

README.md
fpetkovski: Will do !

yeya24: I think it would be nice to run benchmark on each CI to see the stats. We can have an issue to track.

README.md
```diff
@@ -79,3 +79,66 @@ The current implementation uses goroutines very liberally which means the query
...
+vector_selector/new_engine             	      32	  36515120 ns/op	25709769 B/op	   76738 allocs/op
+
+sum/current_engine                     	      22	  47253905 ns/op	 4925751 B/op	   73843 allocs/op
+sum/new_engine                         	      42	  28137694 ns/op	 6933936 B/op	   71399 allocs/op
+
+sum_by_pod/current_engine              	       8	 135361672 ns/op	87259230 B/op	  568616 allocs/op
+sum_by_pod/new_engine                  	      30	  36838789 ns/op	16031030 B/op	  156454 allocs/op
+
+rate/current_engine                    	      24	  49954635 ns/op	17207644 B/op	   78176 allocs/op
+rate/new_engine                        	      20	  57753527 ns/op	28160190 B/op	  103748 allocs/op
```
yeya24, line 98: We know why it is worse in this case? `rate`, `vector`, etc

README.md
fpetkovski: That's a good idea 👍

README.md
fpetkovski: I believe we are worse because we iterate through the data twice for this case, while the new engine has to do it once.

The matrix selector operator produces step vectors, and the engine takes those vectors and converts them into points. Since there's a 1 to 1 mapping between the input and output data, we are wasting lots of time in this conversion. 

I don't think this is an issue though, people do not really query high cardinality data with raw selectors.

README.md
fpetkovski: Updated to the benchstat format. I also added a make target so we can easily reproduce benchmarks, and potentially hook them into the CI in the future.

README.md
yeya24: I am not very familiar with that part of code. Is this possible for us to handle the raw selectors specially and scan the data only once?

fpetkovski: Yes, this should be possible, especially if we have a planner. We can have a different execution path just for expressions that only have a selectors.

PR #45, opened 2022-09-29
Validate binary operator at creation time

fpetkovski:
This commit moves validation for the binary operator at creation time, so that we can use the fallback mechanism for unsupported operators.


saswatamcode (approved):
LGTM!

PR #46, opened 2022-09-29
Adding aggregation function quantile() and validation

sharadgaur:


physicalplan/plan.go
```diff
@@ -122,6 +134,22 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
...
+	switch n := node.(type) {
+	case *parser.NumberLiteral:
+		t = parser.ValueTypeScalar
+	case *parser.StringLiteral:
+		t = parser.ValueTypeString
+	default:
+		return errors.New(fmt.Sprintf("unknown node type: %T", n))
+	}
+	if t != want {
+		return errors.New(fmt.Sprintf("expected type %s in aggregation, got %s", parser.DocumentedType(want), parser.DocumentedType(t)))
```
yeya24, line 148: Probably also add the aggregation function name to the error message?

physicalplan/plan.go
```diff
@@ -122,6 +134,22 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
...
 
+func aggregateValidation(node parser.Node, want parser.ValueType) error {
+	var t parser.ValueType
+	switch n := node.(type) {
+	case *parser.NumberLiteral:
+		t = parser.ValueTypeScalar
+	case *parser.StringLiteral:
+		t = parser.ValueTypeString
+	default:
+		return errors.New(fmt.Sprintf("unknown node type: %T", n))
```
yeya24, line 145: I am not sure if this is accurate. Those node types are known, just not supported param types, right?
We can use the same error message at line 148

yeya24 (approved):
Overall really good work! Just some small nits.

physicalplan/plan.go
```diff
@@ -122,6 +134,22 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
 	}
 }
 
+func aggregateValidation(node parser.Node, want parser.ValueType) error {
```
fpetkovski, line 137: Hm should we do this validation when the aggregator is instantiated?

physicalplan/aggregate/hashaggregate.go
```diff
@@ -30,32 +30,33 @@ type aggregate struct {
 	tables         []aggregateTable
 	series         []labels.Labels
 	newAccumulator newAccumulatorFunc
-
-	stepsBatch int
-	workers    worker.Group
+	param          interface{}
```
fpetkovski, line 33: I think we should be able to use a more specific type here, at least `parser.Item` or something similar.

sharadgaur: I will look

physicalplan/plan.go
sharadgaur: good point.

physicalplan/aggregate/hashaggregate.go
```diff
@@ -140,7 +176,7 @@ func (a *aggregate) workerTask(workerID int, vector model.StepVector) model.Step
 }
 
 func (a *aggregate) initializeVectorizedTables() ([]aggregateTable, []labels.Labels, error) {
-	tables, err := newVectorizedTables(a.stepsBatch, a.aggregation)
+	tables, err := newVectorizedTables(a.stepsBatch, a.aggregation, a.param)
```
fpetkovski, line 179: Unfortunately we have to do validation when we create the operator. If we throw validation errors during execution, we cannot fall back to the old engine for unsupported operations.

yeya24: If the aggregation param validation fails, do we want to fallback to the old engine? I think they will also fail, too

fpetkovski: What I meant with validation is validating whether the engine has support for operators in the promql query. In case of no support, we don't want to fail mid-execution because that prevents us from using the fallback. Even if we did have a mechanism to fall back in that case, we would have wasted lots of time and made unnecessary network calls.

In case of actual PromQL validation issues, I think it should be safe to fail whenever we see fit. One example is many-to-many join which is detected at query execution time. The old engine would fail as well, so no reason to fall back as you indicated.

sharadgaur: I will try to unify the both interfaces so that we can validated early in execution call.

physicalplan/aggregate/scalar_table.go
```diff
@@ -282,7 +283,51 @@ func makeAccumulatorFunc(expr parser.ItemType) (newAccumulatorFunc, error) {
 				},
 			}
 		}, nil
+	case "quantile":
```
fpetkovski, line 286: I think it's best if we add validation here, so that we cannot even instantiate an accumulator with invalid parameters. This way each function can do its own validation.

physicalplan/aggregate/hashaggregate.go
```diff
@@ -32,39 +33,72 @@ type aggregate struct {
...
 	points *model.VectorPool,
 	next model.VectorOperator,
 	aggregation parser.ItemType,
+	param parser.Expr,
 	by bool,
 	labels []string,
 	stepsBatch int,
 ) (model.VectorOperator, error) {
-	newAccumulator, err := makeAccumulatorFunc(aggregation)
+	if err := validateFunc(aggregation, param); err != nil {
```
fpetkovski, line 49: Let's push this into `makeAccumulatorFunc`

fpetkovski (approved):
Looks great 👍

PR #47, opened 2022-09-29
engine: add some fuzzing for aggregation operators

GiedriusS:
Add some simple fuzzing tests for aggregation operators. Mainly using this as a starting point and because I wanted to try out how well fuzzing is supported in the Go toolset.

### DIFF
--- a/engine/enginefuzz_test.go
+++ b/engine/enginefuzz_test.go
@@ -3,6 +3,7 @@ package engine_test
 import (
 	"context"
 	"fmt"
+	"math"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


engine/enginefuzz_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/efficientgo/core/testutil"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/thanos-community/promql-engine/engine"
+)
+
+func FuzzEngineInstantQueryAggregations(f *testing.F) {
+	load := `load 30s
+	http_requests_total{pod="nginx-1"} 1+1x4
+	http_requests_total{pod="nginx-2"} 1+2x4`
```
fpetkovski, line 17: Does it make sense to fuzz the values too?

GiedriusS: Good point, let me add this.

GiedriusS: Added fuzzing for inputs as well. Fuzzing mode uncovers some problems:
```
--- FAIL: FuzzEngineInstantQueryAggregations (0.30s)
    --- FAIL: FuzzEngineInstantQueryAggregations (0.09s)
        enginefuzz_test.go:68: enginefuzz_test.go:68:       
                                                                                                     
                exp: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:0, V:-37.440000000000005}, Metric:labels.Labels{}}}, Warnings:storage.Warnings(nil)}
            
                got: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:0, V:-37.44}, Metric:labels.Labels{}}}, Warnings:storage.Warnings(nil)}
                                                                                                     
            Diff:                                                                                    
            --- Expected                    
            +++ Actual                                                                               
            @@ -1,2 +1,2 @@                                                                          
            -(*promql.Result)({} => -37.440000000000005 @[0])
            +(*promql.Result)({} => -37.44 @[0])     
  ```
  
  But the initial corpus succeeds. So, I think this could be merged and I'll try to fix these problems in follow-up PRs.

fpetkovski: Nice, it's actually good that we uncovered these problems 👍

fpetkovski (approved):

PR #50, opened 2022-09-30
Support lookbackdelta

yeya24:

I am not sure if this covers the lookback delta usecase, but currently most of the lookbackdelta is hardcoded to 5m. This pr removes the hardcoded part and use the engine option's lookback delta value.

### DIFF
--- a/physicalplan/scan/vector_selector.go
+++ b/physicalplan/scan/vector_selector.go
@@ -31,27 +31,29 @@ type vectorSelector struct {
 	once       sync.Once
 	vectorPool *model.VectorPool
 
-	mint        int64
-	maxt        int64
-	step        int64
-	currentStep int64
-	stepsBatch  int
+	mint          int64
+	maxt          int64
+	step          int64
+	currentStep   int64
+	stepsBatch    int
+	lookbackDelta int64
 
 	shard     int
 	numShards int
 }
 
-func NewVectorSelector(pool *model.VectorPool, storage *seriesSelector, mint, maxt time.Time, step time.Duration, stepsBatch, shard, numShards int) model.VectorOperator {
+func NewVectorSelector(pool *model.VectorPool, storage *seriesSelector, mint, maxt time.Time, step, lookbackDelta time.Duration, stepsBatch, shard, numShards int) model.VectorOperator {
 	// TODO(fpetkovski): Add offset parameter.
 	return &vectorSelector{
 		storage:    storage,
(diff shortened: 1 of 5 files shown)


fpetkovski:
If you want to test for correctness, you can maybe try leaving the value hardcoded to 5m, and see if the new tests pass or fail. If they still pass, we might need to extend test coverage for staleness.

yeya24:
Correctness wise, I think this works. But when I change the `lookbackDelta` value to 30s, then I got panic on https://github.com/thanos-community/promql-engine/blob/main/engine/instant_query.go#L92. We need to handle empty vector I think.

physicalplan/scan/vector_selector.go
```diff
@@ -31,27 +31,29 @@ type vectorSelector struct {
...
-		maxt:        maxt.UnixMilli(),
-		step:        step.Milliseconds(),
-		currentStep: mint.UnixMilli(),
-		stepsBatch:  stepsBatch,
+		mint:          mint.UnixMilli(),
+		maxt:          maxt.UnixMilli(),
+		step:          step.Milliseconds(),
+		currentStep:   mint.UnixMilli(),
+		stepsBatch:    stepsBatch,
+		lookbackDelta: lookbackDelta.Milliseconds(),
```
fpetkovski, line 56: Do we need to add this to the matrix selector too?

yeya24: I am not sure tbh. Need to check promql engine code.
But right now when we selecting points for matrix selector, we don't handle lookback delta now. https://github.com/thanos-community/promql-engine/blob/main/physicalplan/scan/matrix_selector.go#L191

fpetkovski:
Can you try adding this when creating the matrix selector, right before instantiating the filter: 
```
lookbackDelta = maxDuration(lookbackDelta, t.Range)
```

Also somewhere you can define
```
func maxDuration(a, b time.Duration) time.Duration {
	if a.Milliseconds() >= b.Milliseconds() {
		return a
	}
	return b
}
```

physicalplan/scan/vector_selector.go
fpetkovski: Yeah I think the matrix selector should be okay as it is, because it has its own lookback in the form of a select range.

yeya24:
@fpetkovski This works. Thanks!

fpetkovski (approved):
Nice work, thanks 👍

PR #52, opened 2022-10-01
Remove duplicated functions in hash aggregate

fpetkovski:
We currently have duplicated code for vector and scalar tables in the hash aggregate function.

This commit removes the duplicates by instantiating scalar tables whenever we don't have vectorized support for operations used in the aggregate operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,10 @@ The current implementation creates a physical plan directly from the PromQL abst
 
 These are the latest benchmarks captured on an Apple M1 Pro processor.
 
-Note that memory usage is higher when executing a query with parallelism greater than 1. This is due to the fact that the engine is able to execute multiple operations at once (e.g. decode chunks from multiple series at the same time), which requires using independent buffers for each parallel operation. 
+Note that memory usage is higher when executing a query with parallelism greater than 1. This is due to the fact that the engine is able to execute multiple operations at once (e.g. decode chunks from multiple series at the same time), which requires using independent buffers for each parallel operation.
 
 Single core benchmarks
+
 ```markdown
 name                                                old time/op    new time/op    delta
 RangeQuery/vector_selector                            28.1ms ± 2%    34.2ms ± 4%  +21.80%  (p=0.000 n=9+10)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -84,9 +84,10 @@ The current implementation creates a physical plan directly from the PromQL abst
 
 These are the latest benchmarks captured on an Apple M1 Pro processor.
 
-Note that memory usage is higher when executing a query with parallelism greater than 1. This is due to the fact that the engine is able to execute multiple operations at once (e.g. decode chunks from multiple series at the same time), which requires using independent buffers for each parallel operation. 
+Note that memory usage is higher when executing a query with parallelism greater than 1. This is due to the fact that the engine is able to execute multiple operations at once (e.g. decode chunks from multiple series at the same time), which requires using independent buffers for each parallel operation.
 
 Single core benchmarks
+
```
fpetkovski, line 90: I wonder why this wasn't caught by the CI lint job before.

physicalplan/aggregate/vector_table.go
```diff
@@ -88,32 +86,6 @@ func newVectorAccumulator(expr parser.ItemType) (vectorAccumulator, error) {
...
-			var mean, cMean float64
-			var aux, cAux float64
-			for _, v := range in {
-				count++
-				delta := v - (mean + cMean)
-				mean, cMean = scan.KahanSumInc(delta/count, mean, cMean)
-				aux, cAux = scan.KahanSumInc(delta*(v-(mean+cMean)), aux, cAux)
-			}
-			return (aux + cAux) / count
-		}, nil
```
yeya24, line 116: We don't need the 2 functions anymore?

fpetkovski: Since the implementation is the same as the scalar version, they should be covered here: https://github.com/thanos-community/promql-engine/pull/52/files#diff-0a6fbd7e1f59129205d1acd4db27f232b5d21f7e6327aa814ac4d062ae98bc6aR148 

Tests are also passing so I think this should work.

README.md
saswatamcode: Ah, I think this is due to the `require_clean_work_tree` function not being defined in Makefile, so diff isn't ever checked. You can add it from here https://github.com/thanos-io/thanos/blob/main/Makefile#L89! 🙂

saswatamcode (approved):
Thanks! LGTM minus one comment about Makefile!

README.md
fpetkovski: Thanks, this solves the issue 👍

PR #55, opened 2022-10-01
Resolve panics in instant queries

fpetkovski:
The series call in instant queries can return more series than the
query can produce. This is because operations like rate will not
produce an output when a series has only one sample in the evaluation
period. The case cannot be detected during the Series call because
it requires decoding chunks, a process that happens only during Next
calls.  This can lead to the series being returned by the Series call,
but never having samples produced by a Next call.

This commit resolves the issue by explicitly checking whether a
series has produced a sample before trying to access the sample.
An illustrative test case is also added in the commit.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -597,6 +597,14 @@ func TestInstantQuery(t *testing.T) {
 			query: "sum(rate(http_requests_total[1m]))",
 		},
 		{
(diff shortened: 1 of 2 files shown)


GiedriusS (approved):
:+1: this is one of the issues that I quite quickly ran into

yeya24 (approved):

PR #56, opened 2022-10-02
Support unary operator

yeya24:

Unary operator only includes `+` and `-`. We don't need to handle `+` specifically, same does the Promql engine. So this pr mainly adds support for the `-` operator.

### DIFF
--- a/physicalplan/unary/unary.go
+++ b/physicalplan/unary/unary.go
@@ -70,25 +70,21 @@ func (u *unaryNegation) Next(ctx context.Context) ([]model.StepVector, error) {
 	if in == nil {
 		return nil, nil
 	}
-	defer u.next.GetPool().PutVectors(in)
-
 	for i, vector := range in {
 		if err := u.workers[i].Send(vector); err != nil {
 			return nil, err
 		}
 	}
 
-	result := u.next.GetPool().GetVectorBatch()
-	for i, vector := range in {
-		output, err := u.workers[i].GetOutput()
-		if err != nil {
+	for i := range in {
+		// Make sure worker finishes the job.
+		// Since it is in-place so no need another buffer.
+		if _, err := u.workers[i].GetOutput(); err != nil {
 			return nil, err
 		}
-		result = append(result, output)
-		u.next.GetPool().PutStepVector(vector)
 	}
 
-	return result, nil
+	return in, nil
 }
 
 func (u *unaryNegation) workerTask(_ int, vector model.StepVector) model.StepVector {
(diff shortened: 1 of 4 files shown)


physicalplan/unary/unary.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	result := u.next.GetPool().GetVectorBatch()
+	for i, vector := range in {
+		output, err := u.workers[i].GetOutput()
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, output)
+		u.next.GetPool().PutStepVector(vector)
```
fpetkovski, line 88: I wonder if this is safe to do here, because the input vector and output vector the same. From what I understand, `floats.Scale` modify the input vector in place. So the next operator might end up with a reference to a vector that is recycled.

I think we can improve the pooling mechanism by adding a `Recycle` method `model.StepVector`, instead of having each operator get the pool of the previous operator. This way we can propagate input vectors forward because each vector will know in which pool it needs to be put back when it is recycled.

yeya24: Should I remove this line for now?
I am not sure about how I should handle the pool part as I don't even add a vector pool for this operator. Is it required to have a pool for each op? Right now I just reuse the next pool

fpetkovski: I think in this case, we can reuse both the vector batch and all step vectors inside the batch. So we shouldn't need to request or recycle anything, just modify in place and pass forward. 

Returning the downstream pool in `GetPool()` is the correct thing to do, and it will ensure that the vector is returned to the correct pool from which it was allocated.

fpetkovski: You can also run the benchmark you added to see whether there is a significantly higher memory usage compared to the old engine.

yeya24: If it is in-place modification, then seems I don't need to allocate any new buffer and I can simply reuse the `[]StepVector` I get at the beginning? Probably not a good practice. WDYT?

```
	in, err := u.next.Next(ctx)
	if err != nil {
		return nil, err
	}
	if in == nil {
		return nil, nil
	}
	for i, vector := range in {
		if err := u.workers[i].Send(vector); err != nil {
			return nil, err
		}
	}

	for i := range in {
		// Make sure worker finishes the job.
		// Since it is in-place so no need another buffer.
		if _, err := u.workers[i].GetOutput(); err != nil {
			return nil, err
		}
	}

	return in, nil
```

fpetkovski (approved):
Nice work 👍

physicalplan/unary/unary.go
fpetkovski: Yes, this seems like the best way to handle it.

PR #59, opened 2022-10-03
engine: add test case

GiedriusS:
Fix TBD.


GiedriusS:
Interesting... locally this fails but works on CI? :thinking: looking...

PR #60, opened 2022-10-04
Consolidate functions

fpetkovski:
Our functions currently do not handle empty inputs or invalid inputs. This can lead to panics in places like max_over_time where a window contains no points.

This commit consolidates all functions in one place, and adds validation helpers that are inlined as middleware. 
The commit also adds fuzzing for these functions.


yeya24 (approved):
Great work!

Issue #39, opened 2022-09-27
Add support for `matrixSelector` not within function


saswatamcode:
We currently don't support instant queries like `http_requests_total[30s]`, as we only parse matrix selectors if they are within functions, but these might be good to have. Not super urgent.

yeya24:
I believe we already support those queries, resolving

yeya24:
Actually this is still not done, reopen

PR #62, opened 2022-10-04
Refactor PromQL function handling

### LINKED ISSUE #23: Refactor PromQL function call implementation
Currently `plan` checks for only the first argument type in `*parser.Call` before deciding on further plan. But we can have functions with multiple different argument types with combinations of instant/range/scalar vectors as outlined [here](https://prometheus.io/docs/prometheus/latest/querying/functions). 

Refactoring it to handle such cases accurately and to also check arg types before executing, will align behavior with old engine. 🙂

saswatamcode:
This PR modifies physical plan to ensure that the engine is able to support multi-argument PromQL functions. It categorizes PromQL functions into different categories and changes query plan tree accordingly, using a function_selector operator.

With this, the engine can now support single instant vector, single range vector, single scalar input functions as well as combined instant vector and scalar input functions. WIll update this PR to also handle combined range vector and scalar functions and variadic ones.

Also while refactoring function logic, this adds support for `parser.MatrixSelector` queries like `http_requests_total[1m]`.

Fixes #23 and #39.


fpetkovski (commented):
Thanks for this PR, I think the direction needs a bit of adjustment, but there is already a lot of code we can reuse. I would make the following suggestions:

First, I would make function to a standalone operator similar to how unary was implemented here: https://github.com/thanos-community/promql-engine/pull/56, it probably does not belong in the scan package. I would also suggest that each function does its own argument parsing instead of having this logic bundled in one central place, similar to what was done in https://github.com/thanos-community/promql-engine/pull/46 and https://github.com/thanos-community/promql-engine/pull/60

I would also leave the special handling of call + matrix selector as it is, we can resolve that case later once we have proper support for arbitrary functions. Let's rather start small and add support for clamp_min and clamp_max, then we can thing about more complex cases.

saswatamcode:
Thanks!

> I would make function to a standalone operator similar to how unary was implemented here: https://github.com/thanos-community/promql-engine/pull/56, it probably does not belong in the scan package. I would also suggest that each function does its own argument parsing instead of having this logic bundled in one central place, similar to what was done in https://github.com/thanos-community/promql-engine/pull/46 and https://github.com/thanos-community/promql-engine/pull/60

Yup I was thinking the same after looking at unary PR! Will do.
For the argument parsing, the reason why I was doing it in centralized way, was that `FunctionCall` only has `promql.Point` args, but it is difficult to deduce type from them. Will try to look on ways of handling that better! 🙂 

> I would also leave the special handling of call + matrix selector as it is, we can resolve that case later once we have proper support for arbitrary functions. Let's rather start small and add support for clamp_min and clamp_max, then we can think about more complex cases.

I think call + matrix selector already works here. I'll try to clean it up a bit. This doesn't add support for multi-arg range vector functions or variadic ones. Will iterate on those in future PRs!

physicalplan/scan/matrix_selector.go
```diff
@@ -27,10 +27,15 @@ type matrixScanner struct {
...
+// different sub-trees of query AST.
+type Injectable struct {
+	Call function.FunctionCall
+}
+
 type matrixSelector struct {
-	funcExpr *parser.Call
-	call     FunctionCall
 	storage  *seriesSelector
+	in       *Injectable
```
fpetkovski, line 38: Hm I don't understand why this change is necessary. We should always be able to inject a function in a matrix selector. Otherwise the return type will not match.

physicalplan/scan/literal_selector.go
```diff
@@ -45,30 +41,7 @@ func NewNumberLiteralSelector(pool *model.VectorPool, mint, maxt time.Time, step
...
-		maxt:        maxt.UnixMilli(),
-		step:        step.Milliseconds(),
-		currentStep: mint.UnixMilli(),
-		stepsBatch:  stepsBatch,
-		val:         val,
-		call:        call,
-		callName:    f.Name,
-	}, nil
-}
-
```
fpetkovski, line 67: 👍

physicalplan/scan/matrix_selector.go
```diff
@@ -127,19 +127,44 @@ func (o *matrixSelector) Next(ctx context.Context) ([]model.StepVector, error) {
...
+					StepTime:    time.UnixMilli(seriesTs),
+					SelectRange: time.Duration(o.selectRange) * time.Millisecond,
+				})
+
+				if result.Point != function.InvalidSample.Point {
+					vectors[currStep].T = result.T
+					vectors[currStep].Samples = append(vectors[currStep].Samples, result.V)
+					vectors[currStep].SampleIDs = append(vectors[currStep].SampleIDs, series.signature)
+				}
+			} else {
```
fpetkovski, line 152: Let's leave this out for now and revert this change. We can fall back to the old engine in this case and maybe open an issue to tackle it in the future.

physicalplan/scan/matrix_selector.go
```diff
@@ -170,9 +195,6 @@ func (o *matrixSelector) loadSeries(ctx context.Context) error {
 		o.series = make([]labels.Labels, len(series))
 		for i, s := range series {
 			lbls := s.Labels()
-			if o.funcExpr.Func.Name != "last_over_time" {
```
fpetkovski, line 173: Is this safe to do?

physicalplan/unary/unary.go
```diff
@@ -27,16 +30,22 @@ type unaryNegation struct {
...
 		next:       next,
 		stepsBatch: stepsBatch,
+		op:         op,
 	}
 
 	u.workers = worker.NewGroup(stepsBatch, u.workerTask)
 	return u, nil
 }
 
+func (o *unaryNegation) Explain() (me string, next []model.VectorOperator) {
```
fpetkovski, line 45: Interesting that this wasn't added. Should we have a separate `Explainable` interface and make `Explain` return 
```
(me string, next []Explainable) 
```
so that we don't miss such cases in the future?

We can also add the method on `model.VectorOperator`, but having a separate interface will help with reusing the same mechanism for visualizing the logical plan.

cc @bwplotka

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
 )
 
 var InvalidSample = promql.Sample{Point: promql.Point{T: -1, V: 0}}
 
+type FunctionArgs struct {
+	Labels       labels.Labels
+	Points       []promql.Point
+	StepTime     time.Time
+	SelectRange  time.Duration
+	ScalarPoints []promql.Point
```
fpetkovski, line 30: Can we make this:
```suggestion
	ScalarPoints []float64
```

I don't think we can specify points in PromQL functions.

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
+	majorArgIndex int
+	majorArg      model.VectorOperator
+	scalarArgs    []model.VectorOperator
+
+	call FunctionCall
+}
+
+func NewFunctionSelector(pool *model.VectorPool, funcExpr *parser.Call, mint, maxt time.Time, step time.Duration, stepsBatch int, call FunctionCall, majorArgIndex int, majorArg model.VectorOperator, scalarArgs ...model.VectorOperator) model.VectorOperator {
+	return &functionSelector{
+		vectorPool:    pool,
```
fpetkovski, line 136: I think this operator can make all changes in-place, like for the unary operator. So we should not need a new pool here, and we can return the `next` pool in the `GetPool()` function.

physicalplan/function/function.go
```diff
@@ -378,6 +556,32 @@ func dropMetricName(l labels.Labels) labels.Labels {
 	return labels.NewBuilder(l).Del(labels.MetricName).Labels(nil)
 }
 
+func clamp(points []promql.Point, scalars []promql.Point) (float64, bool) {
```
fpetkovski, line 559:
```suggestion
func clamp(points []promql.Point, scalars []float64) (float64, bool) {
```

physicalplan/function/function.go
```diff
@@ -410,13 +614,46 @@ func maybeSample(lbls labels.Labels, stepTime time.Time, points []promql.Point,
...
+}
+
 func ignoreEmpty(points []promql.Point, f func() promql.Sample) promql.Sample {
 	if len(points) == 0 {
 		return InvalidSample
 	}
 	return f()
 }
 
+func ignoreEmptyScalar(points, scalars []promql.Point, f func() promql.Sample) promql.Sample {
```
fpetkovski, line 643: Why do we need this function? Scalars should always be available and the syntax will be validated by the PromQL parser.

physicalplan/plan.go
```diff
@@ -121,18 +177,19 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
 		return newCancellableOperator(e.Expr, storage, mint, maxt, step, lookbackDelta)
 
 	case *parser.StringLiteral:
+		// TODO(saswatamcode): Ensure Explain does not panic.
```
fpetkovski, line 180: Let's maybe remove handling of `StringLiteral` to resolve the panic. We should stabilize what we have now so that we can run the engine in actual environments.

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
+
+// functionSelector returns []model.StepVector after processing input with desired function.
+type functionSelector struct {
+	vectorPool    *model.VectorPool
+	funcExpr      *parser.Call
+	mint          int64
+	maxt          int64
+	step          int64
+	currentStep   int64
+	stepsBatch    int
```
fpetkovski, line 124: Do we need these parameters? Functions don't usually select directly from storage.

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
+
+		var scalarPoints []promql.Point
+		if len(o.funcExpr.Func.ArgTypes) > 1 && len(scalarInput) > 0 {
+			for j := range scalarInput {
+				scalarPoints = append(scalarPoints, promql.Point{V: scalarInput[j][0].Samples[0]})
+			}
+		}
+
+		switch o.funcExpr.Args[o.majorArgIndex].Type() {
+		case parser.ValueTypeVector, parser.ValueTypeScalar:
```
fpetkovski, line 206: Hm, keep in mind this decision is not made during plan execution, but is really possible to make it at plan compilation because the syntax tree won't change. So Maybe we can precalculate things in the operator constructor.

The next function should be as tight as possible, and basically do
* call next on dowstream
* apply function
* sent result

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
+	vectorPool    *model.VectorPool
+	funcExpr      *parser.Call
+	mint          int64
+	maxt          int64
+	step          int64
+	currentStep   int64
+	stepsBatch    int
+	series        []labels.Labels
+	once          sync.Once
+	majorArgIndex int
```
fpetkovski, line 127: Do we need to know the index?

physicalplan/function/function.go
```diff
@@ -1,101 +1,279 @@
...
+	majorArg      model.VectorOperator
```
fpetkovski, line 128:
```suggestion
	next      model.VectorOperator
```

fpetkovski (commented):
This looks much much better now. We might want to add a benchmark for clamp_max to make sure we are not actually worse than the previous engine for some reason.

physicalplan/scan/matrix_selector.go
yeya24: This seems to be moved to https://github.com/thanos-community/promql-engine/pull/62/files#diff-c5565c814b1f8018448d49f7dc602d00b7cf0225383932129b1d5ba967bfafa0R257

physicalplan/unary/unary.go
```diff
@@ -27,16 +30,22 @@ type unaryNegation struct {
 func NewUnaryNegation(
 	next model.VectorOperator,
 	stepsBatch int,
+	op parser.ItemType,
 ) (model.VectorOperator, error) {
 	u := &unaryNegation{
 		next:       next,
 		stepsBatch: stepsBatch,
+		op:         op,
```
yeya24, line 38: Umm why do we need OP? It is unary negation so op is `-`. For `+` we don't have an additional operator to handle that, same does promql.

physicalplan/plan.go
saswatamcode: Will remove!

physicalplan/scan/matrix_selector.go
saswatamcode: Yup, issue is already open #39. I'll add a TODO for now. This is also why I added the injectable type.

physicalplan/function/function.go
saswatamcode: So this is basically data points we get from the scalar arguments of a function like `clamp(instant vector, scalar, scalar)`. But yes this can be float64. I'll update!

physicalplan/function/function.go
saswatamcode: Will remove!

physicalplan/function/function.go
saswatamcode: I'll try to modify this!

physicalplan/function/function.go
saswatamcode: Yes, as we use that to categorize the type of function. While the major arg is usually at [0] of PromQL function arguments, there are cases like `histogram_quantile(scalar, instant vector)` where it is at [1] instead. So need to look out for those cases.

physicalplan/unary/unary.go
saswatamcode: You're right! Will remove the `op`.

physicalplan/scan/matrix_selector.go
saswatamcode: So earlier when we only handled `parser.MatrixSelector` via `parser.Call`, we could always inject the function directly. But now that it might be called a separate operator using queries like `http_requests_total[30s]`, it might not always need a call.

Also, I wanted to create a mechanism to inject some modifications while creating a new operator recursively, and this seemed like a way to achieve that. This can be used in the future when we might want to create different modified operators based on queries. Wdyt? 🙂

fpetkovski: For a query like `http_requests_total[30s]`, the return type is a really a matrix, not a vector. So I don't think that query will fit into the model that we have now, and I suspect we might need a completely different operator for that case.

Let's maybe try to tackle that problem separately, and use this PR to enable function calls with an arbitrary number of parameters.

saswatamcode: Yup, I'll add it as a TODO and remove.

physicalplan/plan.go
saswatamcode: Have removed the string literal support that we had previously for now. This would likely need some new operator!

physicalplan/function/function.go
saswatamcode: Have modified this to now parse in the operator constructor itself!

physicalplan/scan/matrix_selector.go
saswatamcode: Have modified this to only have `call` for now. Added TODO

saswatamcode:
Have rebased the PR and addressed most of the feedback! This also removes the string literal support that was there previously, as it was a bit hacky. String literal and range vector would likely need some changes in model, so plan to do those in the future. Stabilizing what is already there! 🙂

physicalplan/function/function.go
```diff
@@ -1,247 +1,479 @@
...
+	StepTime     int64
+	SelectRange  int64
+	ScalarPoints []float64
+}
+
 // FunctionCall represents functions as defined in https://prometheus.io/docs/prometheus/latest/querying/functions/
-type FunctionCall func(labels labels.Labels, points []promql.Point, stepTime int64, selectRange int64) promql.Sample
+type FunctionCall func(f FunctionArgs) promql.Sample
 
 var Funcs = map[string]FunctionCall{
```
saswatamcode, line 202: This map can probably be split into different categories of function, so that the arguments align with PromQL. This would make processing them easier.

physicalplan/function/function.go
```diff
@@ -1,247 +1,479 @@
...
+	var scalarPoints []float64
+	if len(o.funcExpr.Func.ArgTypes) > 1 && len(scalarInput) > 0 {
+		for j := range scalarInput {
+			scalarPoints = append(scalarPoints, scalarInput[j][0].Samples[0])
+		}
+	}
+
+	out := o.next.GetPool().GetVectorBatch()
+	for _, vector := range input {
+		step := o.next.GetPool().GetStepVector(vector.T)
```
saswatamcode, line 116: We can modify step vector in-place here.

saswatamcode (commented):
Some comments from 1:1 with @fpetkovski. 🙂

saswatamcode:
Simplified it and added support for scalars which can select from storage. This still needs some work to add support for `scalar()`, and to streamline operator tree construction for different types of functions!

physicalplan/unary/unary.go
bwplotka: 👍🏽

engine/engine_test.go
```diff
@@ -765,6 +765,60 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
...
+			query: `clamp(1 - http_requests_total, 10 - 5, 10)`,
+		},
+		{
+			name: "func within func query",
+			load: `load 30s
+			http_requests_total{pod="nginx-1"} 1+1x15
+			http_requests_total{pod="nginx-2"} 1+2x18`,
+			query: `clamp(rate(http_requests_total[30s]), 10 - 5, 10)`,
+		},
+		// {
```
bwplotka, line 796: ?

bwplotka (commented):
Generally looks good (:

engine/engine_test.go
saswatamcode: Yes I need to implement `scalar()` as mentioned here https://github.com/thanos-community/promql-engine/pull/62#issuecomment-1280002880 🙂

physicalplan/scan/matrix_selector.go
```diff
@@ -83,9 +81,6 @@ func NewMatrixSelector(
 
 func (o *matrixSelector) Explain() (me string, next []model.VectorOperator) {
 	r := time.Duration(o.selectRange) * time.Millisecond
-	if o.call != nil {
-		return fmt.Sprintf("[*matrixSelector] %v({%v}[%s] %v mod %v)", o.funcExpr.Func.Name, o.storage.matchers, r, o.shard, o.numShards), nil
```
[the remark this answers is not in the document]
saswatamcode: This goes through the function operator, so explain statements would be added there.

physicalplan/plan.go
```diff
@@ -86,46 +87,58 @@ func newOperator(expr parser.Expr, storage storage.Queryable, opts *query.Option
...
-			operators := make([]model.VectorOperator, 0, numShards)
-			for i := 0; i < numShards; i++ {
-				operator := exchange.NewConcurrent(
-					exchange.NewCancellable(
-						scan.NewMatrixSelector(model.NewVectorPool(stepsBatch), filter, e, call, opts, t.Range, vs.Offset, i, numShards),
-					), 2)
-				operators = append(operators, operator)
-			}
+	case *parser.Call:
+		// We can classify PromQL functions into the following categories,
```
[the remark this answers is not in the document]
saswatamcode: Will add

execution/scan/matrix_selector.go
```diff
@@ -123,14 +118,32 @@ func (o *matrixSelector) Next(ctx context.Context) ([]model.StepVector, error) {
...
+
+				if result.Point != function.InvalidSample.Point {
+					vectors[currStep].T = result.T
+					vectors[currStep].Samples = append(vectors[currStep].Samples, result.V)
+					vectors[currStep].SampleIDs = append(vectors[currStep].SampleIDs, series.signature)
+				}
+			} else {
+				// TODO(saswatamcode): Range vector result might need new operator.
+				// https://github.com/thanos-community/promql-engine/issues/39
+				return nil, parse.ErrNotImplemented
```
fpetkovski, line 144: This error needs to be returned when the operator is created.

physicalplan/function/function.go
```diff
@@ -1,247 +1,479 @@
...
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
 
+	"github.com/thanos-community/promql-engine/physicalplan/model"
 	"github.com/thanos-community/promql-engine/physicalplan/parse"
 )
 
+// functionSelector returns []model.StepVector after processing input with desired function.
+type functionSelector struct {
```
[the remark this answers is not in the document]
saswatamcode: Renamed!

execution/scan/matrix_selector.go
saswatamcode: Moved the error

engine/engine_test.go
saswatamcode: This is done in https://github.com/thanos-community/promql-engine/pull/62/commits/95ece7c570466cc589079d31224a11826bc5fd64

engine/engine_test.go
```diff
@@ -786,6 +786,78 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
...
 		},
+		// Result is correct but this likely fails due to https://github.com/golang/go/issues/12025.
+		// TODO(saswatamcode): Test NaN cases separately.
+		// {
+		// 	name: "scalar func with NaN",
+		// 	load: `load 30s
+		//  	http_requests_total{pod="nginx-1"} 1+1x15
+		//  	http_requests_total{pod="nginx-2"} 1+2x18`,
+		// 	query: `scalar(http_requests_total)`,
+		// },
```
fpetkovski, line 797: Maybe we can add a new test case which does not use `testutil`. Possibly a dedicated test for error cases. We can create a new issue/PR for this.

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+
+	call FunctionCall
+}
+
+func NewfunctionOperator(funcExpr *parser.Call, call FunctionCall, nextOps []model.VectorOperator) model.VectorOperator {
+	f := &functionOperator{
+		nextOps:   nextOps,
+		call:      call,
+		funcExpr:  funcExpr,
+		nextIndex: 0,
```
fpetkovski, line 40:
```suggestion
		vectorIndex: 0,
```

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+	}
+
+	for i := range funcExpr.Args {
+		if funcExpr.Args[i].Type() == parser.ValueTypeMatrix || funcExpr.Args[i].Type() == parser.ValueTypeVector {
+			f.nextIndex = i
```
fpetkovski, line 45:
```suggestion
			f.nextIndex = i
			break
```

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+	}
+
+	// Decide based on next type.
+	switch funcExpr.Args[f.nextIndex].Type() {
+	case parser.ValueTypeVector, parser.ValueTypeScalar:
+		f.nextFunc = f.instantVectorOrScalarNext
+	case parser.ValueTypeMatrix:
+		f.nextFunc = f.rangeVectorNext
+	default:
+		f.nextFunc = f.defaultNext
```
fpetkovski, line 56: Let's throw the exception here so that we can catch it and fall back.

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(input) == 0 {
+		return nil, nil
+	}
+
+	// Call next on all scalar inputs.
+	scalarInput := make([][]model.StepVector, len(o.nextOps)-1)
```
fpetkovski, line 104: Can we remove allocs by using a shared slice scoped to the operator?

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+				return nil, err
+			}
+
+			scalarInput[s] = si
+			defer o.nextOps[i].GetPool().PutVectors(scalarInput[s])
+			s++
+		}
+	}
+
+	scalarPoints := make([]float64, len(scalarInput))
```
fpetkovski, line 123: Can be potentially be shared across `Next` invocations

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+	out := o.nextOps[o.nextIndex].GetPool().GetVectorBatch()
```
fpetkovski, line 124: Let's see if we can modify the input batch in place.

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+		return nil, err
+	}
+
+	if len(input) == 0 {
+		return nil, nil
+	}
+
+	// We inject function call in tree, so only need to relay input from here.
+	return input, nil
+}
```
fpetkovski, line 184: Let's revert the Matrix selector nesting under Call so that we can remove the indirection here

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+func (o *functionOperator) defaultNext(ctx context.Context) ([]model.StepVector, error) {
+	return nil, errors.Wrapf(parse.ErrNotImplemented, "got: %s", o.funcExpr)
+}
+
+func (o *functionOperator) loadSeries(ctx context.Context) error {
+	var err error
+	o.once.Do(func() {
+		if o.funcExpr.Func.Name == "vector" {
+			o.series = []labels.Labels{labels.New()}
+			o.nextOps[o.nextIndex].GetPool().SetStepSize(len(o.series))
```
fpetkovski, line 196:
```suggestion
```

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+			return
+		}
+
+		if o.funcExpr.Func.Name == "scalar" {
+			o.series = []labels.Labels{}
+			o.nextOps[o.nextIndex].GetPool().SetStepSize(len(o.series))
```
fpetkovski, line 202:
```suggestion
```

execution/function/function.go
```diff
@@ -0,0 +1,833 @@
...
+			lbls := s
+			if o.funcExpr.Func.Name != "last_over_time" {
+				lbls = dropMetricName(s)
+			}
+
+			sort.Sort(lbls)
+			o.series[i] = lbls
+		}
+
+		o.nextOps[o.nextIndex].GetPool().SetStepSize(len(o.series))
```
fpetkovski, line 222:
```suggestion
```

engine/engine_test.go
saswatamcode: Created issue #88! Not implementing in this PR as it might be complicated, will follow up with a future one. Keeping TODO comment. 🙂

execution/function/function.go
saswatamcode: Added error.

execution/function/function.go
saswatamcode: Have added slice to struct! 🙂

execution/function/function.go
saswatamcode: Have added slice to struct! 🙂

execution/function/function.go
saswatamcode: Reverted.

execution/function/function.go
```diff
@@ -1,247 +1,482 @@
...
+			input[v].SampleIDs = input[v].SampleIDs[:1]
+			input[v].Samples[0] = math.Float64frombits(value.NormalNaN)
+			continue
+		}
+
+		for i := range input[v].Samples {
+			// Scalar expressions can select from storage as well.
+			if len(o.funcExpr.Func.ArgTypes) > 1 && len(o.scalarInput) > 0 {
+				for l := range o.scalarInput {
+					if len(o.scalarInput[l]) >= 1 {
```
fpetkovski, line 133: What happens if this condition is not met?

execution/function/function.go
```diff
@@ -1,247 +1,482 @@
...
+				continue
+			}
+
+			si, err := o.nextOps[i].Next(ctx)
+			if err != nil {
+				return nil, err
+			}
+
+			o.scalarInput[s] = si
+			defer o.nextOps[i].GetPool().PutVectors(o.scalarInput[s])
```
fpetkovski, line 110: I think this will miss the recycling of the vector operator samples

execution/function/function.go
```diff
@@ -1,247 +1,482 @@
...
+		}
+
+		o.series = make([]labels.Labels, len(series))
+		for i, s := range series {
+			lbls := s
+			if o.funcExpr.Func.Name != "last_over_time" {
+				lbls = DropMetricName(s)
+			}
+
+			sort.Sort(lbls)
```
fpetkovski, line 180: I think labels will always be sorted, so maybe we can remove this to save some CPU cycles?

execution/function/function.go
saswatamcode: I thought of adding it as a safeguard against empty input. But yup, a test case is necessary here I think. Might need to handle this differently. Empty series cases could also be a part of #88 I think. 🙂 
Will update this one

execution/function/function.go
saswatamcode: Have added defer `PutStepVector` statements to release samples. Both for `scalarInput` as well as `input`

execution/function/function.go
saswatamcode: I've added handling for this and verified empty cases manually. But a lot of it returns as NaN in certain cases. So will add test cases for this in #88

fpetkovski (approved):
Great work on this, looks good to me 👍 

Let's give other reviewers a bit more time and we can merge.

execution/function/function.go
```diff
@@ -1,247 +1,474 @@
...
+		// scalar() depends on number of samples per vector and returns NaN if len(samples) != 1.
+		// So need to handle this separately here, instead of going via call which is per point.
+		if o.funcExpr.Func.Name == "scalar" {
+			if len(vector.Samples) <= 1 {
+				continue
+			}
+
+			vector.Samples = vector.Samples[:1]
+			vector.SampleIDs = vector.SampleIDs[:1]
+			vector.Samples[0] = math.NaN()
```
yeya24, line 128: Why do we set this to nan?

execution/function/function.go
```diff
@@ -1,247 +1,474 @@
...
+
+		if o.funcExpr.Func.Name == "scalar" {
+			o.series = []labels.Labels{}
+			return
+		}
+
+		series, loadErr := o.nextOps[o.vectorIndex].Series(ctx)
+		if loadErr != nil {
+			err = loadErr
+		}
```
yeya24, line 164: Do we need to return in this case?

engine/engine_test.go
```diff
@@ -786,6 +786,78 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
...
+			http_requests_total{pod="nginx-1"} 1+1x15
+			http_requests_total{pod="nginx-2"} 1+2x18`,
+			query: `clamp_min(http_requests_total, scalar(max(http_requests_total)))`,
+		},
+		{
+			name: "func with scalar arg that selects storage + number",
+			load: `load 30s
+			http_requests_total{pod="nginx-1"} 1+1x15
+			http_requests_total{pod="nginx-2"} 1+2x18`,
+			query: `clamp_min(http_requests_total, scalar(max(http_requests_total)) + 10)`,
```
yeya24, line 859: Can we update the readme with the new benchmark result?

execution/execution.go
```diff
@@ -81,50 +82,74 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
...
+		// We can classify PromQL functions into the following categories,
+		// - Single vector input functions (majority).
+		// - Single matrix input functions (majority).
+		// - Vector with scalar input functions.
+		// - Matrix with scalar input functions. TODO(saswatamcode)
+		// - String based operations on vectors (only label_replace()). TODO(saswatamcode)
+		// - No inputs (only pi() and time()). TODO(saswatamcode)
+		// - Only scalar input (only vector()).
+		// - Variadics (all time based functions like month() + round() and label_join()). TODO(saswatamcode)
+		// Tracked in https://github.com/thanos-community/promql-engine/issues/23
```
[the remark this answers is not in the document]
saswatamcode: Sure will remove!

execution/function/function.go
saswatamcode: `scalar()` will return NaN if the given input vector has more than one sample as mentioned here: https://prometheus.io/docs/prometheus/latest/querying/functions/#scalar.

engine/engine_test.go
saswatamcode: Sure will do!

execution/function/function.go
saswatamcode: Will add

engine/engine_test.go
saswatamcode: Updated the single-core one. For multi-core I don't have an 8-core machine 😅

README.md
```diff
@@ -89,38 +89,68 @@ Note that memory usage is higher when executing a query with parallelism greater
...
+RangeQuery/unary_negation-8                              78.1k ± 0%     90.3k ± 0%   +15.59%  (p=0.000 n=10+10)
+RangeQuery/vector_and_scalar_comparison-8                75.3k ± 0%     91.9k ± 0%   +22.09%  (p=0.000 n=10+9)
+RangeQuery/positive_offset_vector-8                      57.1k ± 0%     66.1k ± 0%   +15.69%  (p=0.000 n=10+10)
+RangeQuery/at_modifier_-8                                75.1k ± 0%     58.7k ± 0%   -21.85%  (p=0.000 n=10+9)
+RangeQuery/at_modifier_with_positive_offset_vector-8     69.1k ± 0%     52.7k ± 0%   -23.76%  (p=0.000 n=8+10)
+RangeQuery/clamp-8                                       84.7k ± 0%    803.2k ± 0%  +848.76%  (p=0.000 n=10+10)
+RangeQuery/clamp_min-8                                   84.4k ± 0%    802.2k ± 0%  +850.57%  (p=0.000 n=10+9)
+RangeQuery/complex_func_query-8                           100k ± 0%      817k ± 0%  +717.42%  (p=0.000 n=10+9)
+RangeQuery/func_within_func_query-8                      93.7k ± 0%    818.3k ± 0%  +772.88%  (p=0.000 n=10+10)
+RangeQuery/aggr_within_func_query-8                      93.7k ± 0%    818.3k ± 0%  +772.81%  (p=0.000 n=10+9)
```
yeya24, line 153: Probably let's have an issue to track the allocation regression. 10x seems not normal, maybe we could have some kind of pooling.

execution/parse/errors.go
```diff
@@ -17,3 +17,5 @@ func UnsupportedOperationErr(op parser.ItemType) error {
 	msg := fmt.Sprintf("operation not supported: %s", t)
 	return errors.Wrap(ErrNotSupportedExpr, msg)
 }
+
+var ErrNotImplemented = errors.New("expression not implemented")
```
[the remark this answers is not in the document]
yeya24: Thanks for the explanation.

execution/function/function.go
yeya24: Thanks for the explanation.

yeya24 (approved):
Thanks for the great work!

README.md
fpetkovski: Agree, I also noticed this which is strange. But total bytes used seems to be in line which is why I was not concerned too much.

saswatamcode: Yup! Will create an issue. There isn't a memory/cpu regression but allocs seem abnormally high

PR #64, opened 2022-10-04
Removing validation as parser is already going to validate the inputs

sharadgaur:


fpetkovski (approved):
Thanks for the clean up.

PR #68, opened 2022-10-05
engine: print query in case of panic

### LINKED ISSUE #51: Print the query in case of a panic
I think we should print the query in case of a panic to help with debugging. For example, I got this:

```
2022/09/30 09:03:56 http: panic serving 1.2.3.4:50046: runtime error: index out of range [0] with length 0
goroutine 16089 [running]:
net/http.(*conn).serve.func1()
        /usr/lib/go-1.19/src/net/http/server.go:1850 +0xbf
panic({0x205da60, 0xc0002530b0})
        /usr/lib/go-1.19/src/runtime/panic.go:890 +0x262
github.com/thanos-community/promql-engine/physicalplan.newOperator({0x26a6b20?, 0xc06255b3e0?}, {0x2689e20, 0xc2bc0b8750}, {0x20?, 0x2160e20?, 0x39470c0?}, {0xc0c5ca1f15418614, 0x970375c00e, 0x39470c0}, 
...)
        /home/giedrius/go/pkg/mod/github.com/thanos-community/promql-engine@v0.0.0-20220926123320-18a0a4b4be8f/physicalplan/plan.go:52 +0xb85
github.com/thanos-community/promql-engine/physicalplan.New(...)
        /home/giedrius/go/pkg/mod/github.com/thanos-community/promql-engine@v0.0.0-20220926123320-18a0a4b4be8f/physicalplan/plan.go:26
github.com/thanos-community/promql-engine/engine.(*engine).NewInstantQuery(0x40dce7?, {0x2689e20, 0xc2bc0b8750}, 0x10000026b9d01?, {0xc01da23b18?, 0x0?}, {0x7f760533e498?, 0x10?, 0x39470c0?})
```

But have no clue what query was being executed. It's probably worth adding a `recover()` in the query execution to print out the query in case it panics.

GiedriusS:
Recover from panics and print the error in case of a panic.

It looks like the following:

```
level=error ts=2022-10-05T08:12:27.931977143Z caller=engine.go:146 msg="runtime panic in parser" expr=sum(up) err="runtime error: invalid memory address or nil pointer dereference" stacktrace="goroutine 45 [running]:\ngithub.com/thanos-community/promql-engine/engine.recoverEngine({0x26ff760, 0xc00011c640}, {0x271d740, 0xc00091c600}, 0xc000926b80)\n\t/home/giedrius/dev/promql-engine/engine/engine.go:144 +0xc6\npanic({0x1eb2580, 0x39b01b0})\n\t/usr/lib/go-1.19/src/runtime/panic.go:884 ...
```

We need to set `*promql.Result` in advance so that later on it wouldn't return `nil` in case of a panic.

This is needed so that we could debug problems in production if / when the engine crashes.

Closes #51.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -4,6 +4,8 @@
 package engine
 
 import (
+	"fmt"
+	"runtime"
 	"time"
 
 	"github.com/efficientgo/core/errors"
@@ -20,7 +22,7 @@ import (
 )
 
 type engine struct {
-	logger promql.QueryLogger
+	logger log.Logger
 
 	lookbackDelta time.Duration
 }
@@ -46,6 +48,7 @@ func New(opts Opts) v1.QueryEngine {
 
 	core := &engine{
 		lookbackDelta: opts.LookbackDelta,
+		logger:        opts.Logger,
 	}
 	if opts.DisableFallback {
 		return core
@@ -106,7 +109,7 @@ func (e *engine) NewInstantQuery(q storage.Queryable, opts *promql.QueryOpts, qs
 		return nil, err
 	}
 
-	return newInstantQuery(plan, expr, ts), nil
+	return newInstantQuery(e.logger, plan, expr, ts), nil
 }
 
 func (e *engine) NewRangeQuery(q storage.Queryable, opts *promql.QueryOpts, qs string, start, end time.Time, interval time.Duration) (promql.Query, error) {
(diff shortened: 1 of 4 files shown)


fpetkovski:
Nice, is there a way we can add a test for this?

GiedriusS:
> Nice, is there a way we can add a test for this?

We could add tests for helper functions but to test how all of this works holistically, I think we'd have to introduce some kind of fault into the code :smile: would that work?

fpetkovski:
Hm, what about throwing a panic from storage?

GiedriusS:
> Hm, what about throwing a panic from storage?

Thanks for the suggestion! Added a test that checks for recovery from panics via storage.

fpetkovski (approved):
Thanks, LGTM

PR #69, opened 2022-10-07
Add Explain() to unary operator

fpetkovski:
The unary operator has no Explain() method, which breaks some tests and will probably break usage.

### DIFF
--- a/physicalplan/unary/unary.go
+++ b/physicalplan/unary/unary.go
@@ -24,6 +24,10 @@ type unaryNegation struct {
 	workers    worker.Group
 }
 
+func (u *unaryNegation) Explain() (me string, next []model.VectorOperator) {
+	return "[*unaryNegation]", []model.VectorOperator{u.next}
+}
+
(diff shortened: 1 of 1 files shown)


physicalplan/unary/unary.go
```diff
@@ -24,6 +24,10 @@ type unaryNegation struct {
 	workers    worker.Group
 }
 
+func (u *unaryNegation) Explain() (me string, next []model.VectorOperator) {
+	return "[*unary]", []model.VectorOperator{u.next}
```
saswatamcode, line 28: Maybe this should be `unaryNegation` as per this [comment](https://github.com/thanos-community/promql-engine/pull/62#discussion_r989283683)?
```suggestion
	return "[*unaryNegation]", []model.VectorOperator{u.next}
```

fpetkovski: Makes sense, updated 👍

saswatamcode (approved):
Thanks!

PR #70, opened 2022-10-07
Support offset modifier

yeya24:
fixes https://github.com/thanos-community/promql-engine/issues/33

### DIFF
--- a/physicalplan/plan.go
+++ b/physicalplan/plan.go
@@ -42,7 +42,7 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
 		return scan.NewNumberLiteralSelector(model.NewVectorPool(stepsBatch), mint, maxt, step, stepsBatch, e.Val), nil
 
 	case *parser.VectorSelector:
-		filter := scan.NewSeriesFilter(storage, mint, maxt, 0, lookbackDelta, e.LabelMatchers)
+		filter := scan.NewSeriesFilter(storage, mint, maxt, 0, lookbackDelta, e.OriginalOffset, e.LabelMatchers)
 		numShards := runtime.GOMAXPROCS(0) / 2
 		if numShards < 1 {
 			numShards = 1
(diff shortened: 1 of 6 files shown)


physicalplan/plan.go
```diff
@@ -42,7 +42,7 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
 		return scan.NewNumberLiteralSelector(model.NewVectorPool(stepsBatch), mint, maxt, step, stepsBatch, e.Val), nil
 
 	case *parser.VectorSelector:
-		filter := scan.NewSeriesFilter(storage, mint, maxt, 0, lookbackDelta, e.LabelMatchers)
+		filter := scan.NewSeriesFilter(storage, mint, maxt, 0, lookbackDelta, e.OriginalOffset, e.LabelMatchers)
```
fpetkovski, line 45: Would everything still work if we just adjusted `mint` and `maxt` to avoid propagating this value in different operators?

yeya24: Good point, I think this works.

physicalplan/plan.go
```diff
@@ -52,7 +52,7 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
 			operator := exchange.NewConcurrent(
 				exchange.NewCancellable(
 					scan.NewVectorSelector(
-						model.NewVectorPool(stepsBatch), filter, mint, maxt, step, lookbackDelta, stepsBatch, i, numShards)), 2)
+						model.NewVectorPool(stepsBatch), filter, mint, maxt, step, lookbackDelta, e.OriginalOffset, stepsBatch, i, numShards)), 2)
```
fpetkovski, line 55: Can we also do the same mint/maxt adjustment for vector and matrix selectors?

yeya24: No, this offset needs to be propagated anyway since we need to change `o.currentStep` value.

fpetkovski: Got it 👍

fpetkovski (approved):
Thanks, lgtm

PR #73, opened 2022-10-07
Fix multi-label hashing

fpetkovski:
The HashForLabels function from prometheus expects the labels to be provided in sorted order. We don't respect this constraint currently, which leads to multi-label grouping to now work correctly do to series not getting correct hashes.

This commit ensures that all grouping labels are sorted in the binary and hash aggregate operators.


saswatamcode (approved):
Nice catch! Thanks!

PR #75, opened 2022-10-10
Support at modifier

yeya24:
Fixes https://github.com/thanos-community/promql-engine/issues/72

This pr adds support for at modifier. 
The current code is a bit messy tbh but I don't know if there are better ways to improve it. We don't support sub query and matrix selector without function so some code have not been added to make this feature complete.

### DIFF
--- a/physicalplan/plan.go
+++ b/physicalplan/plan.go
@@ -1,11 +1,29 @@
 // Copyright (c) The Thanos Community Authors.
 // Licensed under the Apache License 2.0.
 
+// Copyright 2013 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package physicalplan
 
 import (
 	"runtime"
 	"time"
+
+	"github.com/efficientgo/core/errors"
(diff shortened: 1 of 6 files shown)


physicalplan/step_invariant/step_invariant.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (u *stepInvariantOperator) Next(ctx context.Context) ([]model.StepVector, error) {
+	in, err := u.next.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if in == nil || len(in) == 0 || len(in[0].Samples) == 0 {
+		return nil, nil
+	}
+	if !u.duplicateResults {
```
fpetkovski, line 88: What does this flag indicate?

yeya24: Used in https://github.com/prometheus/prometheus/blob/main/promql/engine.go#L1675.
If not matrix or subquery then we will duplicate the values with steps.

fpetkovski:
Thanks, I'll take a look at this more closely tomorrow. I want to make sure we are not simply importing tech debt from the old engine.

physicalplan/plan.go
```diff
@@ -1,48 +1,69 @@
...
 )
 
 const stepsBatch = 10
 
 // New creates new physical query execution plan for a given query expression.
 func New(expr parser.Expr, storage storage.Queryable, mint, maxt time.Time, step, lookbackDelta time.Duration) (model.VectorOperator, error) {
-	return newCancellableOperator(expr, storage, mint, maxt, step, lookbackDelta)
+	// Pre-process the expression to check whether the
+	// expression is step invariant.
+	expr = promql.PreprocessExpr(expr, mint, maxt)
```
fpetkovski, line 44: I wonder if we should add this to the logical plan after https://github.com/thanos-community/promql-engine/pull/79?

yeya24: Yeah that's a great idea.
If we have logical plan then dealing with this should be much more easier.

yeya24:
Thanks!

PR #78, opened 2022-10-12
Handle out of bounds exception in binary operator

fpetkovski:
The vector-to-vector binary operator can throw a panic when the right-hand-side vector is smaller than the left-hand-side vector.

This commit fixes that issue by adding an explicit check for the length of the right-hand-side vector.


saswatamcode (approved):
Thanks! This makes sense.

PR #79, opened 2022-10-12
Logical plan

fpetkovski:


logicalplan/merge_selects.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+// MergeSelects optimizes a binary expression where
+// one select is a superset of the other select.
+// For example, the expression:
+//   metric{a="b", c="d"} / scalar(metric{a="b"}) becomes:
+//   Filter(a="b", metric) / scalar(metric).
+type MergeSelects struct{}
```
bwplotka, line 13: I think it should be called Optimizer in suffix

physicalplan/storage/pool.go
```diff
@@ -0,0 +1,65 @@
...
+	return p.selectors[key]
+}
+
+func (p *SelectorPool) GetFilteredSelector(mint, maxt time.Time, lookbackDelta time.Duration, matchers, filters []*labels.Matcher) SeriesSelector {
+	key := hashMatchers(matchers, mint, maxt, lookbackDelta)
+	if _, ok := p.selectors[key]; !ok {
+		p.selectors[key] = newSeriesSelector(p.queryable, mint, maxt, lookbackDelta, matchers)
+	}
+
+	return NewFilteredSelector(p.selectors[key], NewFilter(filters))
```
fpetkovski, line 44: Maybe we can also cache filtered selectors.

logicalplan/merge_selects.go
```diff
@@ -0,0 +1,132 @@
...
+
+func matcherToMap(matchers []*labels.Matcher) map[string]*labels.Matcher {
+	r := make(map[string]*labels.Matcher, len(matchers))
+	for i := 0; i < len(matchers); i++ {
+		r[matchers[i].Name] = matchers[i]
+	}
+	return r
+}
+
+// matcherHeap is a set of the most selective label matchers
```
yeya24, line 88: What does it mean by selective here? Fewer postings?

logicalplan/merge_selects.go
```diff
@@ -0,0 +1,132 @@
...
+		if !ok {
+			return
+		}
+
+		for _, l := range e.LabelMatchers {
+			if l.Name == labels.MetricName {
+				replacement, found := selectors.findReplacement(l.Name, e.LabelMatchers)
+				if found {
+					// All replacements are done on metrics only,
+					// so we can drop the explicit metric name selector.
```
yeya24, line 57: `All replacements are done on metrics only` Cannot get this one. Did you mean all replacements are done on labels other than the metric name?

logicalplan/plan_test.go
```diff
@@ -0,0 +1,46 @@
...
+		{
+			name:     "different selectors",
+			expr:     `sum(metric{a="b"}) / sum(metric{c="d"})`,
+			expected: `sum(metric{a="b"}) / sum(metric{c="d"})`,
+		},
+		{
+			name:     "different operator",
+			expr:     `sum(metric{a="b"}) / sum(metric{a=~"b"})`,
+			expected: `sum(metric{a="b"}) / sum(metric{a=~"b"})`,
+		},
```
yeya24, line 33: Can we add some test cases for different metric names within the same query?

physicalplan/plan.go
```diff
@@ -144,24 +160,52 @@ func newOperator(expr parser.Expr, storage storage.Queryable, mint time.Time, ma
...
+	case *parser.VectorSelector:
+		return t, nil, nil
+	case *logicalplan.FilteredSelector:
+		return t.VectorSelector, t.Filters, nil
+	default:
+		return nil, nil, parse.ErrNotSupportedExpr
+	}
+}
+
+func newVectorSelector(selector engstore.SeriesSelector, mint time.Time, maxt time.Time, offset, lookbackDelta time.Duration, step time.Duration) (model.VectorOperator, error) {
```
yeya24, line 174: What's the difference between this and `scan.NewVectorSelector`? They are all called `VectorSelector`

physicalplan/storage/pool.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func hashMatchers(matchers []*labels.Matcher, mint time.Time, maxt time.Time, delta time.Duration) uint64 {
+	sb := xxhash.New()
```
yeya24, line 48: Having a pool would be nice

logicalplan/merge_selects.go
fpetkovski: Most selective means the one that selects the most amount of postings. The idea is to select the highest amount of postings we can in a single select, and then filter down in series in selectors that don't need all postings. Let me know how we can change the wording here to make it more clear.

logicalplan/merge_selects.go
fpetkovski: So what I wanted to say here is that we only replace selectors that match for metric names. Something like sum({a="b"}) will not be processed with this optimizer. Any suggestions how we can clarify this?

logicalplan/plan_test.go
fpetkovski: Good point, added 👍

logicalplan/sort_matchers.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (m SortMatchers) Optimize(expr parser.Expr) parser.Expr {
+	traverse(&expr, func(node *parser.Expr) {
+		e, ok := (*node).(*parser.VectorSelector)
+		if !ok {
+			return
+		}
+
+		sort.Slice(e.LabelMatchers, func(i, j int) bool {
+			return e.LabelMatchers[i].Name == e.LabelMatchers[j].Name
```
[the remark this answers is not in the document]
fpetkovski: Good point, fixed

yeya24 (approved):
LGTM. We need to resolve conflicts.
Btw any benchmark for the optimization we've done? Where did you get the idea of this optimization?

logicalplan/plan.go
```diff
@@ -0,0 +1,99 @@
...
+var DefaultOptimizers = []Optimizer{
+	SortMatchers{},
+	MergeSelectsOptimizer{},
+}
+
+type Plan interface {
+	RunOptimizers([]Optimizer) parser.Expr
+}
+
+type Optimizer interface {
```
saswatamcode, line 22: Maybe renaming to `RuleBasedOptimizer` would be good here? In case we choose to implement plan cost estimation in the future? 🙂 
```suggestion
type RuleBasedOptimizer interface {
```

logicalplan/plan.go
```diff
@@ -0,0 +1,99 @@
...
+	Optimize(parser.Expr) parser.Expr
```
saswatamcode, line 23: Perhaps we should also add `Explain()` methods here for sake of debuggability?

logicalplan/filter.go
```diff
@@ -0,0 +1,28 @@
...
+package logicalplan
+
+import (
+	"fmt"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+type FilteredSelector struct {
```
saswatamcode, line 13: So this seems to change PromQL spec and adds a new `filter()` function to it. I'm wondering how safe this is for example `filter` is implemented upstream, in which case it may do something completely different and we end up with diverging specs.

logicalplan/merge_selects.go
```diff
@@ -0,0 +1,132 @@
...
+// Licensed under the Apache License 2.0.
+
+package logicalplan
+
+import (
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+// MergeSelectsOptimizer optimizes a binary expression where
```
saswatamcode, line 11: So it feels like this optimizer is more of a cache-based optimization rather than a rule-based one? I think we can also implement this in the physical plan directly as an optimization effort, instead of extending PromQL and having a separate `filter()` for it as a rule in the logical plan. It feels like this doesn't really "rewrite the query", but rather implements something new.

We definitely should do this caching, but I wonder if the logical plan is the right place to implement it.
This is new to me, so feel free to correct me if I'm getting this wrong somewhere! 🙂

saswatamcode (commented):
Thanks for the awesome work! This already looks super impressive. 🚀 
Would love to see some benchmarks too! 💪🏻 

Some questions/suggestions! 🙂

fpetkovski:
I added a benchmark for a binary query like 
```
sum(http_requests_total{code="200"}) / sum(http_requests_total)
```

Here are the results:
```
goos: darwin
goarch: arm64
pkg: github.com/thanos-community/promql-engine/engine
BenchmarkMergeSelectorsOptimizer
BenchmarkMergeSelectorsOptimizer/withoutOptimizers
BenchmarkMergeSelectorsOptimizer/withoutOptimizers-8         	      37	  31050148 ns/op	44220461 B/op	  275174 allocs/op
BenchmarkMergeSelectorsOptimizer/withOptimizers
BenchmarkMergeSelectorsOptimizer/withOptimizers-8            	      40	  27853710 ns/op	40639897 B/op	  225706 allocs/op
PASS
```

I would expect the optimization to work even better in Thanos where we fetch data over the network. With a local TSDB, I believe we can reference the same series without copying them in both the nominator and denominator. In Thanos, both selectors would hold copies of the data which would increase memory usage even more, compared to not having the optimization.

I also added an explicit flag to enable/disable optimizers. Should this be set to true or false by default?

fpetkovski:
@yeya24 yes https://github.com/thanos-io/thanos/issues/4407 should be fixed by this. We don't cache decoded samples, this is hard because we decode chunks in time-based steps. But we do cache fetched series, which means encoded chunks will be cached. Maybe as a next improvement, we can figure out how to only decode shared chunks once and reuse them in all operators that need them.

logicalplan/filter.go
fpetkovski: That's true, we are adding a new AST node type, similar to how StepInvariant was added upstream: https://github.com/prometheus/prometheus/blob/96d5a32659f0e3928c10a771e50123fead9828bd/promql/parser/ast.go#L178-L183

However, this is not changing the actual spec of PromQL. It only changes the parsed AST. In the best case scenario we would have our own structs copied from the AST and work with them. I thought that might be an overkill for now so I decided to start with something simpler. 

If something like this was added to upsteram PromQL, I would expect some tests to fail somewhere :) I am not sure if there is a better way to prevent such incompatibilities, and I don't also see how exactly they would manifests.

logicalplan/merge_selects.go
fpetkovski: Yeah that's true, we are adding two optimizations in this case. The physical plan can cache series calls, and logical optimizers make sure that the cache-hit ratio is optimized. 

I think adding optimizers to the physical plan would not be straightforward, since the phyisical plan is essentially a set of instantiated operators. We would need to figure out how to transform and rewire them.

yeya24:
I am wondering if we need the flag for optimization? In which case we want to disable it?

fpetkovski:
Ok, let's go with that. I will change the flag to `DisableOptimizers`. The use case I see is for debugging purposes if some queries return wrong results.

logicalplan/plan.go
fpetkovski: Yeah I'd love to add this, but maybe in a follow up PR? I am concerned this change might just keep growing.

logicalplan/plan.go
fpetkovski: Let's maybe introduce such distinctions when we have different optimizer types?

logicalplan/filter.go
saswatamcode: Ack! I see! This makes sense, thanks!

logicalplan/plan.go
saswatamcode: Sure!

saswatamcode (approved):
LGTM! ❤️

GiedriusS:
> Ok, let's go with that. I will change the flag to `DisableOptimizers`. The use case I see is for debugging purposes if some queries return wrong results.

Maybe this could also cover https://github.com/thanos-community/promql-engine/issues/66?

logicalplan/plan_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestDefaultOptimizers(t *testing.T) {
+	cases := []struct {
+		name     string
+		expr     string
+		expected string
+	}{
+		{
+			name:     "common selectors",
+			expr:     `sum(metric{a="b", c="d"}) / sum(metric{a="b"})`,
+			expected: `sum(filter([a="b" c="d"], metric{a="b"})) / sum(metric{a="b"})`,
```
GiedriusS, line 25: Do we need to keep the old matchers in the filter when the vector selector has them? For example, I'd expect to see

```go
			expected: `sum(filter([c="d"], metric{a="b"})) / sum(metric{a="b"})`,
```

Here? :thinking:

engine/engine.go
```diff
@@ -14,25 +14,31 @@ import (
...
+	debugWriter      io.Writer
+	lookbackDelta    time.Duration
+	enableOptimizers bool
 }
 
 type Opts struct {
 	promql.EngineOpts
 
+	// DisableOptimizers disables query optimizations using logicalPlan.DefaultOptimizers.
+	DisableOptimizers bool
```
bwplotka, line 40: My worry is that this has to be extended to have different optimization modes, possible with https://yourbasic.org/golang/bitmask-flag-set-clear/

I guess we can break compatibility still and change it in future if needed.

engine/engine.go
```diff
@@ -122,7 +129,11 @@ func (e *engine) NewInstantQuery(q storage.Queryable, _ *promql.QueryOpts, qs st
 		return nil, err
 	}
 
-	plan, err := physicalplan.New(expr, q, ts, ts, 0, e.lookbackDelta)
+	logicalPlan := logicalplan.New(expr, ts, ts)
+	if e.enableOptimizers {
+		logicalPlan = logicalPlan.RunOptimizers(logicalplan.DefaultOptimizers)
```
bwplotka, line 134:
```suggestion
		logicalPlan = logicalPlan.Optimize(logicalplan.DefaultOptimizers)
```

engine/engine_test.go
```diff
@@ -1317,37 +1333,42 @@ func TestInstantQuery(t *testing.T) {
...
+	disableOptimizers := []bool{true, false}
 	lookbackDeltas := []time.Duration{30 * time.Second, time.Minute, 5 * time.Minute, 10 * time.Minute}
-	for _, lookbackDelta := range lookbackDeltas {
-		opts.LookbackDelta = lookbackDelta
-		for _, tc := range cases {
-			t.Run(tc.name, func(t *testing.T) {
-				test, err := promql.NewTest(t, tc.load)
-				testutil.Ok(t, err)
-				defer test.Close()
+	for _, withOptimizers := range disableOptimizers {
```
bwplotka, line 1338:
```suggestion
	for _, withoutOptimizers := range disableOptimizers {
```

logicalplan/filter.go
bwplotka: Yes, we have to extend AST if we want to reuse it and avoid transforming types too many times.

logicalplan/merge_selects.go
bwplotka: Let's keep it simple -> optimizer changing logical plan

logicalplan/merge_selects.go
```diff
@@ -0,0 +1,132 @@
...
+					filters := dropMetricName(e.LabelMatchers)
+					e.LabelMatchers = replacement
+					*node = &FilteredSelector{
+						Filters:        filters,
+						VectorSelector: e,
+					}
+					return
+				}
+			}
+		}
```
bwplotka, line 67:
```suggestion
		for _, l := range e.LabelMatchers {
			if l.Name != labels.MetricName {
			  continue
			}
			replacement, found := selectors.findReplacement(l.Value, e.LabelMatchers)
			if !found {
			   continue
   			}
			// All replacements are done on metrics name only,
			// so we can drop the explicit metric name selector.
			filters := dropMetricName(e.LabelMatchers)
			e.LabelMatchers = replacement
			*node = &FilteredSelector{
				Filters:        filters,
				VectorSelector: e,
			}
			return
		}
```

logicalplan/merge_selects.go
bwplotka: Proposed alternative.

logicalplan/merge_selects.go
```diff
@@ -0,0 +1,132 @@
...
+// For example, the expression:
+//
+//	metric{a="b", c="d"} / scalar(metric{a="b"}) becomes:
+//	Filter(c="d", metric{a="b"}) / scalar(metric{a="b"}).
+//
+// The engine can then cache the result of `metric{a="b"}`
+// and apply an additional filter for {c="d"}.
+type MergeSelectsOptimizer struct{}
+
+func (m MergeSelectsOptimizer) Optimize(expr parser.Expr) parser.Expr {
```
bwplotka, line 22: I don't really get why this merge selector can work only on `__name__`? Why we have to treat metric name specially here?

bwplotka, line 22: ` Optimize(expr parser.Expr) parser.Expr {`

We need to invest in good documentation to make it easy to optimize with just this interface (:

logicalplan/plan.go
bwplotka: 👍🏽

logicalplan/plan.go
bwplotka: fine to add later

logicalplan/plan_test.go
bwplotka: That's a good point - I guess we do double selection here which is not a harm - we can optimize this later (:

bwplotka (approved):
Amazing work! Small nits and LGTM - can't wait to feed real data to optimizers for optimization decisions 😱 

(e.g imagine logical plan would knew series for next few steps)

logicalplan/merge_selects.go
metalmatze: Optimizers should only optimize the logical plan indeed. The physical plan is then constructed from that optimized logical plan and can make certain assumptions while being constructed.

logicalplan/plan.go
metalmatze: It'd be such a killer feature to have `EXPLAIN` for PromQL indeed. Seeing how the physical storage is going to be queried. :100:

metalmatze (commented):
Really awesome stuff. Lots of good discussions!

logicalplan/merge_selects.go
fpetkovski: It's done mainly to constraint the problem and make it easier to solve. We can generalize it in subsequent iterations, I mainly wanted to prototype something simple which, and make it safe to use and match most cases.

logicalplan/plan_test.go
fpetkovski: True, it would be safe and better to strip selectors from filters. It will make also make filtering faster.

logicalplan/merge_selects.go
fpetkovski: Thanks, I modified the comment a bit.

logicalplan/plan_test.go
fpetkovski: This ended up being fairly straightforward, so I implemented it in this PR.

fpetkovski:
Thanks everyone for the comments, they should all be resolved now. Let's merge this and start playing with it.

bwplotka:
💪🏽

PR #80, opened 2022-10-13
Consolidate operator arguments

fpetkovski:
This commit consolidates arguments into various scan operators by using the query.Options data structure to capture arguments and common calculations.

### DIFF
--- a/query/options.go
+++ b/query/options.go
@@ -3,11 +3,34 @@
 
 package query
 
-import "time"
+import (
+	"time"
+)
 
 type Options struct {
-	Start time.Time
-	End   time.Time
-	Step  time.Duration
-	Range time.Duration
+	Start         time.Time
+	End           time.Time
+	Step          time.Duration
+	LookbackDelta time.Duration
+
+	StepsBatch int64
+}
+
+func (o *Options) NumSteps() int {
+	// Instant evaluation is executed as a range evaluation with one step.
+	if o.Step.Milliseconds() == 0 {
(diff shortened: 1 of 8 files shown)


saswatamcode (approved):
LGTM!

query/options.go
```diff
@@ -3,11 +3,28 @@
...
-import "time"
+import (
+	"time"
+)
 
 type Options struct {
-	Start time.Time
-	End   time.Time
-	Step  time.Duration
-	Range time.Duration
```
yeya24, line 12: @fpetkovski Do we still need SelectRange here?

fpetkovski: I think we cannot have it here, because these are global options for the entire query. The range is really a parameter for a matrix selector and can be different for different selectors in a single query.

yeya24: Okay okay, so these options are not per operator, but for the whole query.
Then for expressions like sub query, we might need to use another start/end time to keep track of the iteration?

yeya24:
Sorry looks like the PR I just merged breaks this one. Would you mind fixing the conflict and then we can merge? It might be very messy sorry

query/options.go
fpetkovski: Yes possibly. I am actually not that familiar with subqueries so I will also need to investigate more.

fpetkovski:
No worries, I will rebase this tomorrow and resolve all conflicts 👍

PR #81, opened 2022-10-13
Add a counter metric for number of fallbacks

yeya24:

Fixes https://github.com/thanos-community/promql-engine/issues/77

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -64,18 +64,19 @@ func New(opts Opts) v1.QueryEngine {
 	return &compatibilityEngine{
 		core: core,
 		prom: promql.NewEngine(opts.EngineOpts),
-		fallbacks: promauto.With(opts.Reg).NewCounter(
+		queries: promauto.With(opts.Reg).NewCounterVec(
 			prometheus.CounterOpts{
-				Name: "thanos_engine_fallbacks_total",
+				Name: "thanos_engine_queries_total",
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -62,12 +64,18 @@ func New(opts Opts) v1.QueryEngine {
 	return &compatibilityEngine{
 		core: core,
 		prom: promql.NewEngine(opts.EngineOpts),
+		fallbacks: promauto.With(opts.Reg).NewCounter(
+			prometheus.CounterOpts{
+				Name: "thanos_engine_fallbacks_total",
```
saswatamcode, line 69: Maybe we can add a label to denote if the engine did fallback, and the metric could be `thanos_engine_queries_total` instead? It would be useful to see the percentage of queries that fallback under real-world use. 🙂

saswatamcode (commented):
Thanks!

I think `go.mod` needs to be regenerated. Also, one small idea!

engine/engine.go
yeya24: Updated. Thanks for the suggestion!

saswatamcode (approved):
LGTM minus the `go.mod` and metric `Help` text! 🙂

saswatamcode (approved):

fpetkovski:
Ty!

PR #83, opened 2022-10-17
Refactor instant & range queries; Renamed physical plan to execution.

bwplotka:
No logic change, only refactor that makes us to experiment on consistent our engine API (we are not tied to methods we don't want to implement). We need new interface to combine more optimizations - caller might want to decide when (construction vs execution).

Typically we see split in plans (https://github.com/google/lovefield/blob/master/docs/dd/04_query_engine.md#overview-life-of-a-query) - we can add that later if needed, just wanted to avoid confusion. Our physical plan was just implementation not a plan (IMO).

Signed-off-by: bwplotka <bwplotka@gmail.com>


engine/engine.go
```diff
@@ -64,26 +65,24 @@ func New(opts Opts) v1.QueryEngine {
...
 	return &compatibilityEngine{
-		core: core,
-		prom: promql.NewEngine(opts.EngineOpts),
+		core:            core,
+		disableFallback: opts.DisableFallback,
+		prom:            promql.NewEngine(opts.EngineOpts),
 		queries: promauto.With(opts.Reg).NewCounterVec(
 			prometheus.CounterOpts{
-				Name: "thanos_engine_queries_total",
+				Name: "promql_engine_queries_total",
```
bwplotka, line 74: Kind of unrelated, but IMO cleaner.

engine/engine.go
```diff
@@ -92,84 +91,212 @@ func (e *compatibilityEngine) SetQueryLogger(l promql.QueryLogger) {
...
+var errNotImplemented = errors.New("not implemented")
+
+func (e *engine) NewExecutor(q storage.Queryable, _ *promql.QueryOpts, expr parser.Expr, start, end time.Time, step time.Duration) (model.VectorOperator, error) {
+	// if step = 0 it is an instant query.
+	// Use same check as Prometheus for range queries.
+	if step > 0 && expr.Type() != parser.ValueTypeVector && expr.Type() != parser.ValueTypeScalar {
 		return nil, errors.Newf("invalid expression type %q for range query, must be Scalar or instant Vector", parser.DocumentedType(expr.Type()))
 	}
 
+	// TODO(bwplotka): We plan to have single plan that can be optimized in both physical and logical context. Rename.
```
bwplotka, line 286: Would you agree @saswatamcode @fpetkovski ?

saswatamcode: Just to clarify, this means that for naive plan -> executor -> execute flow, we would do rule-based optimizer step in naive plan, create tree of operators in executor and then execute that?

bwplotka: Yes. We will have `plan` which we can do rule based or cost esimtation optimizations. This will give us optimized plan. This can be then transformed to something that can be exected

saswatamcode: I think the `plan` terminology is getting a bit confusing, especially after the logical plan PR. 😅 

So afaiu theoretically, `logical plan` refers to the query tree and is basically a high-level arrangement of "logical" operators i.e operators without implementations (we don't do this yet, we only have `Plan` interface, which takes in and runs `Optimizers` that traverse and transform nodes in the query tree). This `logical plan` can be optimized via rules which change the relational expression of the query to make the query more efficient and have the engine do less work. With cost-based optimization, we can create several different `logical plans` and pick the one with the least estimated cost (this is also not implemented yet).

From this `logical plan` we can create a `physical plan` (we currently just pass expression with new `logicalplan` nodes), which forms a tree of executable operators that passes actual lower-level storage details to the executable operators. It can be executed by calling `Next()` and `Series()` on root node. Optimizations here include optimized operators which can take advantage of lower-level storage details. This means it can add and implement new operators for intra-operator parallelism, concurrency, flow control, batched selects, etc like `exchange` operators.

> Yes. We will have plan which we can do rule based or cost esimtation optimizations. This will give us optimized plan. This can be then transformed to something that can be executed

So with single `plan` in future, you mean a combined logical + physical plan, that would create an executable operator tree with `exchange` operators as well right from the start, and perform rule-based/cost-based optimizations on it before executing? 🙂 

I think the range/instant query method refactor great! For the combined or split plan, I'm not sure what would be easier to implement and change. 🤔 

Maybe @fpetkovski would have better opinions on this?

bwplotka:
Closing for now, adding more changes.

engine/engine.go
fpetkovski: I think we might be overcomplicating things. Logical operators (AST) = logical plan. Physical operators  = physical plan.
https://learn.microsoft.com/en-us/sql/relational-databases/showplan-logical-and-physical-operators-reference?view=sql-server-ver16

bwplotka: Removed this comment to avoid confusion.

saswatamcode (approved):
LGTM! 🌟

engine/engine.go
bwplotka: Yea, I want to do it literally now (control concurrency, distributions etc) - and I blocked by this.

fpetkovski (approved):

PR #86, opened 2022-10-18
Fix edge case with duplicate matchers

fpetkovski:
Duplicate matchers are not properly detected during MergeSelect optimizations. There was also a subtle bug where the original selectors were being modified when trimming filters because they both were pointing at the same slice.

This commit fixes both issues and adds additional unit tests for those cases.


saswatamcode (approved):
Thanks!

yeya24 (approved):

PR #89, opened 2022-10-24
Propagate select hints to storage

fpetkovski:
Thanos Stores rely on select hints for returning downsampled data when querying long time ranges. Unless hints are properly set, stores cannot return the correct downsampled chunks, leading to incorrect queries. Hints are also used for query pushdown inside sidecars.

This commit adds support for propagating select hints to the underlying storage.

### DIFF
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -54,7 +54,7 @@ func New(expr parser.Expr, queryable storage.Queryable, mint, maxt time.Time, st
 		StepsBatch:    stepsBatch,
 	}
 	selectorPool := engstore.NewSelectorPool(queryable)
-	hints := &storage.SelectHints{
+	hints := storage.SelectHints{
 		Start: mint.UnixMilli(),
 		End:   maxt.UnixMilli(),
 		// TODO(fpetkovski): Adjust the step for sub-queries once they are supported.
@@ -63,7 +63,7 @@ func New(expr parser.Expr, queryable storage.Queryable, mint, maxt time.Time, st
 	return newCancellableOperator(expr, selectorPool, opts, hints)
 }
 
-func newCancellableOperator(expr parser.Expr, selectorPool *engstore.SelectorPool, opts *query.Options, hints *storage.SelectHints) (*exchange.CancellableOperator, error) {
+func newCancellableOperator(expr parser.Expr, selectorPool *engstore.SelectorPool, opts *query.Options, hints storage.SelectHints) (*exchange.CancellableOperator, error) {
 	operator, err := newOperator(expr, selectorPool, opts, hints)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


execution/execution.go
```diff
@@ -45,40 +45,49 @@ const stepsBatch = 10
...
 		LookbackDelta: lookbackDelta,
 		StepsBatch:    stepsBatch,
 	}
-	selectorPool := engstore.NewSelectorPool(storage)
-	return newCancellableOperator(expr, selectorPool, opts)
+	selectorPool := engstore.NewSelectorPool(queryable)
+	hints := &storage.SelectHints{
+		Start: mint.UnixMilli(),
+		End:   maxt.UnixMilli(),
+		Step:  step.Milliseconds(),
```
GiedriusS, line 60: Maybe we can add a TODO here about this: https://github.com/prometheus/prometheus/pull/11163/files#diff-ad2761bb162a2f12326b18f2714ef54f4cd476cfe3388f77b2b20c5ff82e4faaR857? Subqueries are not supported at the moment so we can't add this but maybe we can add a comment so that we would not forget.

execution/storage/pool.go
```diff
@@ -27,37 +28,56 @@ func NewSelectorPool(queryable storage.Queryable) *SelectorPool {
...
+	writeString(sb, hints.Func)
+	writeString(sb, strings.Join(hints.Grouping, ";"))
+	writeBool(sb, hints.By)
 
 	key := sb.Sum64()
 	return key
 }
+
+func writeMatcher(sb *xxhash.Digest, m *labels.Matcher) {
+	writeString(sb, m.Name)
```
GiedriusS, line 65: Do we need to write the separator here like previously? as far as I can tell the label value potentially could be the same as the label name? :thinking:

yeya24:
> Thanos Stores rely on select hints for returning downsampled data when querying long time ranges.

Sorry didn't get this one. Select hints is a Prometheus thing, do we really use it to propogate downsampling information? IIRC downsampling info is on the store API level only.

execution/execution.go
fpetkovski: Makes sense, I added a todo for this. I also copied the tests from the old engine from the latest main, so that they fail if we don't adjust the step hint properly.

fpetkovski:
@yeya24 yes this is a bug I saw in Thanos when using the new engine. Stores gateways use the function hint to know which downsampled chunk to return (whether average, sum, count etc..).

execution/execution.go
```diff
@@ -45,40 +45,50 @@ const stepsBatch = 10
...
-func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.Options) (model.VectorOperator, error) {
+func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.Options, hints *storage.SelectHints) (model.VectorOperator, error) {
 	switch e := expr.(type) {
 	case *parser.NumberLiteral:
 		return scan.NewNumberLiteralSelector(model.NewVectorPool(stepsBatch), opts, e.Val), nil
 
 	case *parser.VectorSelector:
 		start, end := getTimeRangesForVectorSelector(e, opts, 0)
-		filter := storage.GetSelector(start, end, e.LabelMatchers)
+		hints.Start = start
```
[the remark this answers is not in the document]
fpetkovski: I think this is a good idea 👍 This is exactly why hints for binary selectors are copied, but it's best to be safe and propagate copies everywhere. Will make that change.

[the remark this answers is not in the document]
fpetkovski: Updated

GiedriusS (approved):

PR #95, opened 2022-10-25
Return error when iterating a series set

csmarchbanks:
The series set may have an error for some reason which was currently being ignored as .Next() would return false and .Err() was never checked.


fpetkovski (approved):
Thanks for the contribution 👍

PR #100, opened 2022-10-26
Keep __name__ label when running comparisions

csmarchbanks:
In the case of queries like foo < bar, the result should continue to have the name of the metric, unlike operations such as multiplication.

I am not sure this is the best solution, but it does seem to make the tests pass.

### DIFF
--- a/execution/binary/scalar.go
+++ b/execution/binary/scalar.go
@@ -12,6 +12,7 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql/parser"
 
+	"github.com/thanos-community/promql-engine/execution/function"
 	"github.com/thanos-community/promql-engine/execution/model"
 )
 
@@ -34,8 +35,7 @@ type scalarOperator struct {
 	getOperands   getOperandsFunc
 	operandValIdx int
 	operation     operation
-	opName        string
-	isComparison  bool
+	opType        parser.ItemType
 }
 
 func NewScalar(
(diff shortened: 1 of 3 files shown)


execution/binary/scalar.go
```diff
@@ -145,8 +147,11 @@ func (o *scalarOperator) loadSeries(ctx context.Context) error {
 	series := make([]labels.Labels, len(vectorSeries))
 	for i := range vectorSeries {
 		if vectorSeries[i] != nil {
-			lbls := labels.NewBuilder(vectorSeries[i]).Del(labels.MetricName).Labels(nil)
-			series[i] = lbls
+			lbls := labels.NewBuilder(vectorSeries[i])
```
saswatamcode, line 150: Maybe using `function.DropMetricName` instead of full `Builder` would save some mem, as we only drop name label here? We do the same in matrix selector. 🙂

execution/binary/scalar.go
```diff
@@ -63,6 +64,7 @@ func NewScalar(
 		scalar:        scalar,
 		operation:     binaryOperation,
 		opName:        parser.ItemTypeStr[op],
+		isComparison:  op.IsComparisonOperator(),
```
saswatamcode, line 67: Maybe we can directly pass `op` in struct and use that instead `Explain` and `loadSeries`? 
```suggestion
		op:        op,
```

saswatamcode (approved):
Thanks! I think we already do something like this for functions! 

Some small suggestions!

csmarchbanks:
Thanks for the review! I added a second commit with the suggestions.

saswatamcode (approved):
LGTM!

fpetkovski (approved):
Thanks, lgtm

Issue #99, opened 2022-10-26
sum(rate(...)) divided by itself gives no values (in some cases?)


GiedriusS:
Example query:
```
(sum(rate(thanos_cache_redis_hits_total{name="caching-bucket"}[2m])) by (service)) /
(sum(rate(thanos_cache_redis_hits_total{name="caching-bucket"}[2m])) by (service))
```

Explain:
```
[*CancellableOperator]: [*vectorOperator] / one-to-one ignoring false group []:
├──[*CancellableOperator]: [*CancellableOperator]: [*concurrencyOperator(buff=2)]:
│  └──[*CancellableOperator]: [*aggregate] sum by ([service]):
│     └──[*CancellableOperator]: [*coalesceOperator]:
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 0 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 1 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 2 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 3 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 4 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
│        │  └──[*CancellableOperator]: [*matrixSelector] rate({[__name__="thanos_cache_redis_hits_total" name="caching-bucket"]}[2m0s] 5 mod 8)
│        ├──[*concurrencyOperator(buff=2)]:
[...]

PR #101, opened 2022-10-27
engine: add repro for #99 + potential fix

GiedriusS:
Repro for #99 plus potential fix.

### DIFF
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -180,7 +180,13 @@ func (o *matrixSelector) loadSeries(ctx context.Context) error {
 		for i, s := range series {
 			lbls := s.Labels()
 			if o.funcExpr.Func.Name != "last_over_time" {
-				lbls = function.DropMetricName(lbls)
+				// This modifies the array in place. Because labels.Labels
+				// can be re-used between different Select() calls, it means that
+				// we have to copy it here.
+				// TODO(GiedriusS): could we identify somehow whether labels.Labels
+				// is reused between Select() calls?
+				lbls = function.DropMetricName(lbls.Copy())
+
 			}
 
 			sort.Sort(lbls)
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -97,6 +97,12 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 		step  time.Duration
 	}{
 		{
+			name: "func with scalar arg that selects storage, checks whether same series handled correctly",
+			load: `load 30s
+			    thanos_cache_redis_hits_total{name="caching-bucket",service="thanos-store"} 1+1x30`,
+			query: `clamp_min(thanos_cache_redis_hits_total, scalar(max(thanos_cache_redis_hits_total) by (service))) + clamp_min(thanos_cache_redis_hits_total, scalar(max(thanos_cache_redis_hits_total) by (service)))`,
+		},
+		{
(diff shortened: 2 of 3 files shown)


execution/scan/matrix_selector.go
```diff
@@ -180,7 +180,13 @@ func (o *matrixSelector) loadSeries(ctx context.Context) error {
 		for i, s := range series {
 			lbls := s.Labels()
 			if o.funcExpr.Func.Name != "last_over_time" {
-				lbls = function.DropMetricName(lbls)
+				// This modifies the array in place. Because labels.Labels
+				// can be re-used between different Select() calls, it means that
+				// we have to copy it here.
+				// TODO(GiedriusS): could we identify somehow whether labels.Labels
+				// is reused between Select() calls?
+				lbls = function.DropMetricName(lbls.Copy())
```
GiedriusS, line 188: I found out that labels can be reused across Select() calls however making a copy each time is costly. Perhaps @fpetkovski you have some ideas on how to solve this better?

GiedriusS:
Interesting thing is that even with this fix with the same form of the query but different matchers, I get results only for the first 5 minutes or so in a range query:

image

... but at least

```
(sum(rate(thanos_cache_redis_hits_total{name="caching-bucket"}[2m])) by (service)) /
(sum(rate(thanos_cache_redis_hits_total{name="caching-bucket"}[2m])) by (service))
```

works so I might have run into another bug.

execution/scan/matrix_selector.go
fpetkovski: Hmm so if I understand correctly, a downstream operator can modify the labels from its upstream, and thereby modify the labels of the upstream as well. I assume this is because we modify the slice in place, something that @saswatamcode implemented to reduce memory usage. 

I think the correct thing would be to make a copy after all. Otherwise we can end up with this sort of subtle bugs that are hard to track. When you say it's costly to copy the labels, did you see anything in a benchmark that stood out?

saswatamcode: Hmm, I thought that labels were essentially "handed-off" to downstream operators but that doesn't seem to be the case.

saswatamcode:
Are you using `rate` in the query above?

execution/scan/matrix_selector.go
saswatamcode, line 188: Can you try doing the same change in the function operator? Instant vector and scalar functions are implemented via that https://github.com/thanos-community/promql-engine/blob/main/execution/function/function.go#L175

GiedriusS: Added your change plus a test for it.

GiedriusS:
https://github.com/thanos-community/promql-engine/pull/101#issuecomment-1293290420 this is some kind of different bug for a different PR. PTAL @fpetkovski @saswatamcode

saswatamcode (approved):
LGTM!

fpetkovski:
Were you able to fix the problem with this PR? I wonder if somehow series caching introduced the issue.

fpetkovski:
@GiedriusS I think the problem from that comment is fixed  here: https://github.com/thanos-community/promql-engine/pull/103

execution/scan/matrix_selector.go
yeya24: I feel we should maintain the semantics for every operators series function to prevent this kind of bug in the future.

Should we always do a labels copy if labels are modified?

PR #102, opened 2022-10-27
Reduce allocs in function operator

fpetkovski:
The function operator allocates a point slice per step, which leads to ~10x more allocations compared to the old engine.

This commit modifies the operator to reuse the same slice with in-place modifications. This reduces allocs to a level comparable to the old engine. This commit also updates the README with the latest benchmarks.


Makefile
```diff
@@ -85,12 +85,12 @@ benchmarks:
 .PHONY: bench-old
 bench-old: benchmarks
 	@echo "Benchmarking old engine"
-	@go test ./... -bench 'BenchmarkRangeQuery/.*/old_engine'  -run none -count 10 | sed -u 's/\/old_engine//' > benchmarks/old.out
+	@go test ./... -bench 'BenchmarkRangeQuery/.*/old_engine'  -run none -count 5 | sed -u 's/\/old_engine//' > benchmarks/old.out
```
fpetkovski, line 88: Now that the amount of benchmarks has grown, we can reduce this number to get results faster. Five iterations seem to be enough for statistical significance.

GiedriusS (approved):
:muscle: impressive

PR #103, opened 2022-10-28
Add support for histogram_quantile

fpetkovski:
This commit adds support for the histogram_quantile function. It does that through a separate operator to avoid complicating the existing function operator.

This commit also fixes a bug with the step invariant operator found while testing.

Fixes https://github.com/thanos-community/promql-engine/issues/10


execution/execution.go
```diff
@@ -204,6 +217,10 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		}
 
 	case *parser.StepInvariantExpr:
+		switch t := e.Expr.(type) {
```
saswatamcode, line 220: What case does this handle?

fpetkovski: This is fixing a regression. The case was added here but got removed at one point: https://github.com/thanos-community/promql-engine/pull/75/files#diff-9fd654a57fa6fe30931b10efaf6f7d1eaf5bebdac900785e7d3b79faa9bfcc85R166

README.md
```diff
@@ -90,132 +90,138 @@ Single core benchmarks
...
+RangeQuery/vector_and_scalar_comparison               32.7MB ± 0%    30.9MB ± 0%   -5.48%  (p=0.016 n=5+4)
+RangeQuery/positive_offset_vector                     17.7MB ± 0%    27.2MB ± 0%  +54.20%  (p=0.008 n=5+5)
+RangeQuery/at_modifier_                               28.3MB ± 0%    35.1MB ± 0%  +24.20%  (p=0.016 n=4+5)
+RangeQuery/at_modifier_with_positive_offset_vector    28.1MB ± 0%    35.0MB ± 0%  +24.36%  (p=0.008 n=5+5)
+RangeQuery/clamp                                      34.1MB ± 0%    29.2MB ± 0%  -14.36%  (p=0.008 n=5+5)
+RangeQuery/clamp_min                                  34.1MB ± 0%    29.1MB ± 0%  -14.41%  (p=0.008 n=5+5)
+RangeQuery/complex_func_query                         48.5MB ± 0%    31.1MB ± 0%  -35.99%  (p=0.008 n=5+5)
+RangeQuery/func_within_func_query                     35.0MB ± 0%    30.6MB ± 0%  -12.51%  (p=0.008 n=5+5)
+RangeQuery/aggr_within_func_query                     35.0MB ± 0%    30.6MB ± 0%  -12.51%  (p=0.008 n=5+5)
+RangeQuery/histogram_quantile                         49.4MB ± 0%    96.1MB ± 0%  +94.43%  (p=0.008 n=5+5)
```
fpetkovski, line 134: I wasn't able to track down where the excess memory originates from. I don't think it is that bad and maybe we can open an issue to track this?

saswatamcode: Yup makes sense!

execution/function/histogram.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/thanos-community/promql-engine/execution/model"
+)
+
+type histogramSeries struct {
+	outputID   int
+	upperBound float64
+}
+
+// functionOperator returns []model.StepVector after processing input with desired function.
```
yeya24, line 25: Comment needs to be updated.

execution/function/histogram.go
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
```
yeya24, line 2: I think it is okay to have a separate operator for this function. Just wondering you make it a separate opterator because of the function arguments?
What about other functions that take different args. Shall we use a new operator every time?

fpetkovski: It's not just the arguments, but also this operator has its own pool, it has separate series as outputs (we need to drop the "le" label) and it has a different Next function. So everything is really different which is why I thought it is cleaner to have a separate operator.

execution/function/histogram.go
```diff
@@ -0,0 +1,199 @@
...
+		once:         sync.Once{},
+		scalarOp:     nextOps[0],
+		vectorOp:     nextOps[1],
+		scalarPoints: make([]float64, stepsBatch),
+	}, nil
+}
+
+func (o *histogramOperator) Explain() (me string, next []model.VectorOperator) {
+	next = []model.VectorOperator{o.scalarOp, o.vectorOp}
+	return fmt.Sprintf("[*functionOperator] %v(%v)", "histogram_quantile", o.funcArgs), next
```
yeya24, line 58: We don't need `%v` for `histogram_quantile`

execution/execution.go
GiedriusS: Just tested that this fixes a problem for us that I've mentioned here: https://github.com/thanos-community/promql-engine/pull/101#issuecomment-1293290420

saswatamcode (approved):
LGTM! Thanks for amazing work! 💪🏻 

A small idea about organizing `newOperator` a bit!

execution/function/histogram.go
yeya24: Umm, actually I meant we can remove `%s` totally since `histogram_quantile` is a constant :)

execution/execution.go
```diff
@@ -92,6 +92,23 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		return newShardedVectorSelector(selector, opts, e.Offset)
 
 	case *parser.Call:
+		hints.Func = e.Func.Name
+		hints.Grouping = nil
+		hints.By = false
+
+		if e.Func.Name == "histogram_quantile" {
```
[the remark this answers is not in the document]
fpetkovski: Thanks, I will look into doing that as a follow up.

fpetkovski:
I will merge this so that we fix the regression and start testing the new operator.

PR #107, opened 2022-11-01
Support topk and bottomK aggregations

alanprot:
Tentative to fix https://github.com/thanos-community/promql-engine/issues/105

This is just a POC to discuss if this is a viable solution.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -207,10 +207,7 @@ func (q *compatibilityQuery) Exec(ctx context.Context) (ret *promql.Result) {
 	}
 
 	series := make([]promql.Series, len(resultSeries))
-	for i := 0; i < len(resultSeries); i++ {
-		series[i].Metric = resultSeries[i]
-		series[i].Points = make([]promql.Point, 0, 121) // Typically 1h of data.
-	}
+	seriesMapping := map[uint64]int{}
 
 loop:
 	for {
@@ -228,7 +225,7 @@ loop:
 
 			// Case where Series call might return nil, but samples are present.
 			// For example scalar(http_request_total) where http_request_total has multiple values.
-			if len(series) == 0 && len(r) != 0 {
+			if len(resultSeries) == 0 && len(r) != 0 {
 				numSeries := 0
 				for i := range r {
 					numSeries += len(r[i].Samples)
@@ -251,7 +248,14 @@ loop:
 
 			for _, vector := range r {
 				for i, s := range vector.SampleIDs {
-					series[s].Points = append(series[s].Points, promql.Point{
+					si, ok := seriesMapping[s]
+					if !ok {
+						si = len(seriesMapping)
+						seriesMapping[s] = si
+						series[si].Metric = resultSeries[s]
+						series[si].Points = make([]promql.Point, 0, 121) // Typically 1h of data.
+					}
+					series[si].Points = append(series[si].Points, promql.Point{
 						T: vector.T,
 						V: vector.Samples[i],
 					})
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -207,8 +207,9 @@ func (q *compatibilityQuery) Exec(ctx context.Context) (ret *promql.Result) {
 	}
 
 	series := make([]promql.Series, len(resultSeries))
-	seriesMapping := map[uint64]int{}
-
+	for i := 0; i < len(resultSeries); i++ {
+		series[i].Metric = resultSeries[i]
+	}
 loop:
 	for {
 		select {
@@ -248,14 +249,10 @@ loop:
 
 			for _, vector := range r {
 				for i, s := range vector.SampleIDs {
-					si, ok := seriesMapping[s]
-					if !ok {
-						si = len(seriesMapping)
-						seriesMapping[s] = si
-						series[si].Metric = resultSeries[s]
-						series[si].Points = make([]promql.Point, 0, 121) // Typically 1h of data.
+					if len(series[s].Points) == 0 {
+						series[s].Points = make([]promql.Point, 0, 121) // Typically 1h of data.
 					}
-					series[si].Points = append(series[si].Points, promql.Point{
+					series[s].Points = append(series[s].Points, promql.Point{
 						T: vector.T,
 						V: vector.Samples[i],
 					})
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1232,7 +1232,8 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-7", series="3"} 11
 						http_requests_total{pod="nginx-8", series="4"} 22
 						http_requests_total{pod="nginx-9", series="4"} 89`,
-			query:        "topk(2, http_requests_total)",
+			query: "topk(2, http_requests_total)",
+			// TODO (alanprot): Top/BottomK series order on result. https://github.com/thanos-community/promql-engine/issues/120
 			sortByLabels: true,
 		},
 		{
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -20,8 +20,9 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
-	"github.com/thanos-community/promql-engine/engine"
 	"go.uber.org/goleak"
+
+	"github.com/thanos-community/promql-engine/engine"
 )
 
 func TestMain(m *testing.M) {
@@ -1072,6 +1073,12 @@ func TestInstantQuery(t *testing.T) {
 		queryTime                time.Time
 		compareSeriesResultOrder bool // if true, the series in the result between the old and new engine should have the same order
 	}{
+		{
+			name:      "scalar",
+			load:      ``,
+			queryTime: time.Unix(160, 0),
+			query:     "12 + 1",
+		},
 		{
 			name: "increase plus offset",
 			load: `load 1s
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1108,11 +1108,11 @@ func TestInstantQuery(t *testing.T) {
 	}
 
 	cases := []struct {
-		load                     string
-		name                     string
-		query                    string
-		queryTime                time.Time
-		compareSeriesResultOrder bool // if true, the series in the result between the old and new engine should have the same order
+		load         string
+		name         string
+		query        string
+		queryTime    time.Time
+		sortByLabels bool // if true, the series in the result between the old and new engine should be sorted before compared
 	}{
 		{
 			name:      "scalar",
@@ -1232,8 +1232,8 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-7", series="3"} 11
 						http_requests_total{pod="nginx-8", series="4"} 22
 						http_requests_total{pod="nginx-9", series="4"} 89`,
-			query:                    "topk(2, http_requests_total)",
-			compareSeriesResultOrder: true,
+			query:        "topk(2, http_requests_total)",
+			sortByLabels: true,
 		},
 		{
 			name: "topk by series",
@@ -1247,8 +1247,8 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-7", series="3"} 11
 						http_requests_total{pod="nginx-8", series="4"} 22
 						http_requests_total{pod="nginx-9", series="4"} 89`,
-			query:                    "topk(2, http_requests_total) by (series)",
-			compareSeriesResultOrder: true,
+			query:        "topk(2, http_requests_total) by (series)",
+			sortByLabels: true,
 		},
 		{
 			name: "bottomK",
@@ -1262,8 +1262,8 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-7", series="3"} 11
 						http_requests_total{pod="nginx-8", series="4"} 22
 						http_requests_total{pod="nginx-9", series="4"} 89`,
-			query:                    "bottomk(2, http_requests_total)",
-			compareSeriesResultOrder: true,
+			query:        "bottomk(2, http_requests_total)",
+			sortByLabels: true,
 		},
 		{
 			name: "bottomk by series",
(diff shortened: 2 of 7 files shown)


execution/aggregate/scalar_table.go
```diff
@@ -163,34 +174,78 @@ func makeAccumulatorFunc(expr parser.ItemType, arg parser.Expr) (newAccumulatorF
...
+							total: v,
+						})
+					}
+				},
+				ValueFunc: func(sId uint64) float64 {
+					return res[sId]
+				},
+				HasValue: func(sId uint64) bool {
+					if res == nil {
+						res = map[uint64]float64{}
```
yeya24, line 222: Can we allocate the map with the heap size?

execution/aggregate/scalar_table.go
```diff
@@ -163,34 +174,78 @@ func makeAccumulatorFunc(expr parser.ItemType, arg parser.Expr) (newAccumulatorF
...
 				Reset: func() {
 					hasValue = false
 					value = 0
 				},
 			}
 		}, nil
+	case "topk":
+		return func() *accumulator {
+			k := int(arg.(*parser.NumberLiteral).Val)
+			samplesHeap := samplesHeap{}
```
yeya24, line 196: Can we preallocate the samples heap size to k?

execution/aggregate/scalar_table.go
```diff
@@ -163,34 +174,78 @@ func makeAccumulatorFunc(expr parser.ItemType, arg parser.Expr) (newAccumulatorF
...
+				HasValue:  func(_ uint64) bool { return hasValue },
 				Reset: func() {
 					hasValue = false
 					value = 0
 				},
 			}
 		}, nil
+	case "topk":
+		return func() *accumulator {
+			k := int(arg.(*parser.NumberLiteral).Val)
```
saswatamcode, line 197: I think `k` here needs to come from a `model.VectorOperator` and be iterated over somehow. This would fail for scalar expressions that contain binary operators or select from storage. For eg `topk(3-2, http_requests_total)` would have an interface conversion error here.

engine/engine_test.go
```diff
@@ -309,6 +309,32 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
...
 			step:  28 * time.Second,
 		},
+		{
+			name: "topk",
+			load: `load 30s
+				http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
+				http_requests_total{pod="nginx-2", series="2"} 2+2.3x50
+				http_requests_total{pod="nginx-4", series="3"} 5+2.4x50
+				http_requests_total{pod="nginx-5", series="1"} 8.4+2.3x50
+				http_requests_total{pod="nginx-6", series="2"} 2.3+2.3x50`,
```
saswatamcode, line 319: For this, maybe having test cases with series that are not monotonically increasing would help better in catching any inaccuracies?

saswatamcode (commented):
Thanks for amazing work!

Some comments. 🙂

fpetkovski:
Thanks for this PR. I think directionally it makes sense, I was also thinking of having something along these lines by adding the metric to the output vector. 

What I would suggest is maybe considering a separate operator for topk and bottomk because they do not act like typical aggregations. I wonder if we can add this operator to the `function` package because these operations behave more like functions. @yeya @saswatamcode wdyt?

engine/engine.go
```diff
@@ -228,6 +227,9 @@ loop:
 
 				for _, vector := range r {
 					for i := range vector.Samples {
+						if len(series[i].Points) == 0 {
```
fpetkovski, line 230: What is the reason for moving the allocation here?

execution/aggregate/scalar_table.go
alanprot: Oh..I see..

This means that we already have this problem here?

https://github.com/thanos-community/promql-engine/blob/1a53dd6b99d8da0502361dcd14437ad18bb098ec/execution/aggregate/scalar_table.go#L286-L287

execution/model/vector.go
```diff
@@ -6,8 +6,9 @@ package model
 import "github.com/prometheus/prometheus/model/labels"
 
 type Series struct {
-	ID     uint64
-	Metric labels.Labels
+	ID      uint64
+	AccHash uint64
```
fpetkovski, line 10: It might be better for the operator itself to calculate this value. in a query with multiple aggregations, each aggregation will likely use different labels to aggregate, which means the hash will be different.

yeya:
I think you mean @yeya24

engine/engine.go
alanprot: As in this case I'm adding all series as output but only filling some, i wanna avoid this memory allocation (so i only allocate when I'm sure there are samples on it).

This reduced a lot the memory allocation on the benchmark.

execution/aggregate/scalar_table.go
saswatamcode: Yeah, nice catch, it's the same problem there too!

saswatamcode:
@fpetkovski I kind of agree that they fit better in function operator, but then we'd also need to account for without/by clauses there then.
Aggregations are of two forms, 
`<aggr-op> [without|by (<label list>)] ([parameter,] <vector expression>)` for eg sum 
and
`<aggr-op>([parameter,] <vector expression>) [without|by (<label list>)]` for eg topk, bottomk, quantile

Perhaps a completely separate operator than can handle aggregations with params would be better here? As @alanprot mentioned quantile has issues too, so those can be handled then.

execution/aggregate/scalar_table.go
alanprot: Created an issue:

https://github.com/thanos-community/promql-engine/issues/108

Idk if we should go ahead with this PR as it is or holt it until this issue get fixed.

WDYT?

fpetkovski:
Got it, I completely forgot about `by`/`without`. Let's then have a new operator in the `hashaggregate` package to avoid adding various exceptions and branches in the normal aggregate.

execution/aggregate/scalar_table.go
saswatamcode: We don't need to block this PR! 🙂  
If this can be fixed here with maybe a new operator that can handle parameterized aggregations, quantile can be moved to that new operator then:  https://github.com/thanos-community/promql-engine/pull/107#issuecomment-1299612683

alanprot:
Lemme see if I understood,

For aggregations that has parameters (topk,botomK,quantiles etc) we should have a brand new operator instead of using the  `NewHashAggregate`,  is that correct?

fpetkovski:
That would be the idea, but feel free to push back if you notice it ends up being too much work or duplication. We can extract and reuse common functions if necessary, and I would also not worry about setting up workers in this new operator (I think that concept will go away). We can do the aggregation sequentially for each step.

alanprot:
Nice. Thanks!!

I will try to raise a PR this week to fix https://github.com/thanos-community/promql-engine/issues/108 with the new operator, and we can continue this one from there.

alanprot:
Ok..
I created a new Operator to the top/bottomK aggregations but it seems we still have a problem (or not).

The old promql engine seems to sort the response `by value` in cases for this aggregation and as we are pre-populating series on the response before calculating the aggregation, we sometimes get the result our of order:

Ex:

```
exp: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:50000, V:89}, Metric:labels.Labels{labels.Label{Name:"__name__", Value:"http_requests_total"}, labels.Label{Name:"pod", Value:"nginx-9"}, labels.Label{Name:"series", Value:"4"}}}, promql.Sample{Point:promql.Point{T:50000, V:22}, Metric:labels.Labels{labels.Label{Name:"__name__", Value:"http_requests_total"}, labels.Label{Name:"pod", Value:"nginx-8"}, labels.Label{Name:"series", Value:"4"}}}}, Warnings:storage.Warnings(nil)}
got: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:50000, V:22}, Metric:labels.Labels{labels.Label{Name:"__name__", Value:"http_requests_total"}, labels.Label{Name:"pod", Value:"nginx-8"}, labels.Label{Name:"series", Value:"4"}}}, promql.Sample{Point:promql.Point{T:50000, V:89}, Metric:labels.Labels{labels.Label{Name:"__name__", Value:"http_requests_total"}, labels.Label{Name:"pod", Value:"nginx-9"}, labels.Label{Name:"series", Value:"4"}}}}, Warnings:storage.Warnings(nil)} 
```

(note the expected response is sorted by value, not by labels - which seems to make sense in this case)

I dont think that this behavior is really documented but seems to make sense.

Thoughts @yeya24 @fpetkovski @saswatamcode ??

alanprot:
Beanchmark: https://gist.github.com/alanprot/95ffc868754df540bd7e0c96987b825c

execution/execution.go
```diff
@@ -165,17 +165,33 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
...
+		if e.Op == parser.TOPK || e.Op == parser.BOTTOMK {
+			next, err = aggregate.NewKHashAggregate(model.NewVectorPool(stepsBatch), next, paramOp, e.Op, !e.Without, e.Grouping)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			next, err = aggregate.NewHashAggregate(model.NewVectorPool(stepsBatch), next, e.Op, e.Param, !e.Without, e.Grouping, stepsBatch)
+			if err != nil {
+				return nil, err
+			}
```
yeya24, line 191: We could move the error check outside

execution/aggregate/khashaggregate.go
```diff
@@ -0,0 +1,221 @@
...
+	defer a.next.GetPool().PutVectors(in)
+
+	var arg float64
+
+	if a.paramOp != nil {
+		p, err := a.paramOp.Next(ctx)
+		if err != nil {
+			return nil, err
+		}
+		arg = p[0].Samples[0]
```
yeya24, line 92: One question, can we put this into once? I am wondering if the topk, bottomk param would change while iterating it.

execution/execution.go
alanprot: yeah! make sense!!

execution/aggregate/khashaggregate.go
alanprot: good catch! :D

Yeah.. we should not put on once as it can change! :D fixed.

execution/aggregate/khashaggregate.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	if in == nil {
+		return nil, nil
+	}
+
+	defer a.next.GetPool().PutVectors(in)
+
+	var args []model.StepVector
+
+	if a.paramOp != nil {
```
fpetkovski, line 87: I think we can skip this check because the parameter will always be set. If we want some validation, we can do it in the constructor method.

execution/execution.go
```diff
@@ -165,17 +165,31 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
...
+		var paramOp model.VectorOperator
 
 		next, err := newOperator(e.Expr, storage, opts, hints)
 		if err != nil {
 			return nil, err
 		}
-		a, err := aggregate.NewHashAggregate(model.NewVectorPool(stepsBatch), next, e.Op, e.Param, !e.Without, e.Grouping, stepsBatch)
+
+		if e.Param != nil {
+			paramOp, err = newOperator(e.Param, storage, opts, hints)
```
fpetkovski, line 176: I think we can push this into the `if e.Op == parser.TOPK || e.Op == parser.BOTTOMK` statement. We shouldn't need to validate if `e.Param` is set, because for topk and bottomk the parser will do the validation for us.

execution/aggregate/khashaggregate.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return fmt.Sprintf("[*kaggregate] %v without (%v)", a.aggregation.String(), a.labels), []model.VectorOperator{a.paramOp, a.next}
+}
+
+func (a *kAggregate) init(ctx context.Context) error {
+	series, err := a.next.Series(ctx)
+	if err != nil {
+		return err
+	}
+	hapsHash := make(map[uint64]*samplesHeap)
```
fpetkovski, line 138: What does this map hold, and can we make it a slice instead? So far we tried to avoid using maps for lookups because they tend to slow things down. This is why we assign a zero-based ID to each series and we use that as a key into a slice.

fpetkovski (commented):
I have some small comments but looks really good 👍 

Maybe we can also amend the benchmarks from the README as part of this PR with a `topk` expression.

execution/aggregate/khashaggregate.go
alanprot: Make sense.. i had the other operator on my mind

execution/execution.go
alanprot: Yeah.. but i think im preparing to the merge conflict about the quantiles PR.
The `NewHashAggregate` will still have a case where it needs a param.

execution/aggregate/khashaggregate.go
alanprot: This has is only to calculate the `by` `without` aggragation.  In other words, basically a need one heap per `by` value.

This is similar to what is done on:

https://github.com/thanos-community/promql-engine/blob/f0c54531a5ee47fcc80d57f97cbb362914b02f9b/execution/aggregate/hashaggregate.go#L185-L194

execution/execution.go
alanprot: I rebase and i think now makes more sense?

execution/aggregate/khashaggregate.go
```diff
@@ -0,0 +1,232 @@
...
+	for i := range a.params {
+		a.params[i] = math.NaN()
+		if i < len(args) {
+			a.params[i] = args[i].Samples[0]
+			a.paramOp.GetPool().PutStepVector(args[i])
+		}
+	}
+	a.paramOp.GetPool().PutVectors(args)
+
+	if len(args) != len(in) {
```
fpetkovski, line 102: I think we can remove this now because we inject `NaN`s when a parameter value is not found.

execution/aggregate/khashaggregate.go
fpetkovski: Got it. I think this can be optimize into a slice. I will look into it as a follow up.

fpetkovski (approved):
Thanks for the work, this looks good to me. Some cleanups can be done as a follow up.

yeya24:
@alanprot Could you resolve the conflict of Readme? Then we are good to merge.

yeya24:
Thanks!

PR #113, opened 2022-11-04
Copy labels before droping metric name

fpetkovski:
This is a short term fix, but long term we should modify the function itself to make a copy of input labels.


GiedriusS (approved):

GiedriusS:
It's not always necessary to copy, for example: https://github.com/thanos-community/promql-engine/pull/113/files#diff-50fb9b14773c87a02f8624df659f3d407d8194332173deb10ce953ee55016ffdL168.

fpetkovski:
Yes, this is true. My idea was to pass multiple labels to drop and make the operation more atomic. In any case, it's just an idea.

PR #115, opened 2022-11-04
Implement matcher propagation

fpetkovski:
This commit implements matcher propagation between two vector selectors in a binary expression.
The feature is implemented as an opt-in optimizer which can be injected when instantiating the engine.

Let's discuss potential options on how to allow opting into certain optimizations.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -39,11 +39,8 @@ const (
 type Opts struct {
 	promql.EngineOpts
 
-	// DisableOptimizers disables Query optimizations using logicalPlan.DefaultOptimizers.
-	DisableOptimizers bool
-
-	// AdditionalOptimizers are optimizers which are run in addition to the default ones.
-	AdditionalOptimizers []logicalplan.Optimizer
+	// LogicalOptimizers are optimizers which are run in addition to the default ones.
+	LogicalOptimizers []logicalplan.Optimizer
 
 	// DisableFallback enables mode where engine returns error if some expression of feature is not yet implemented
 	// in the new engine, instead of falling back to prometheus engine.
@@ -72,12 +69,11 @@ func New(opts Opts) v1.QueryEngine {
 				Help: "Number of PromQL queries.",
 			}, []string{"fallback"},
 		),
-		debugWriter:          opts.DebugWriter,
-		disableFallback:      opts.DisableFallback,
-		disableOptimizers:    opts.DisableOptimizers,
-		logger:               opts.Logger,
-		lookbackDelta:        opts.LookbackDelta,
-		additionalOptimizers: opts.AdditionalOptimizers,
+		debugWriter:       opts.DebugWriter,
+		disableFallback:   opts.DisableFallback,
+		logger:            opts.Logger,
+		lookbackDelta:     opts.LookbackDelta,
+		logicalOptimizers: opts.LogicalOptimizers,
 	}
 }
 
@@ -87,11 +83,10 @@ type compatibilityEngine struct {
 
 	debugWriter io.Writer
 
-	disableFallback      bool
-	disableOptimizers    bool
-	logger               log.Logger
-	lookbackDelta        time.Duration
-	additionalOptimizers []logicalplan.Optimizer
+	disableFallback   bool
+	logger            log.Logger
+	lookbackDelta     time.Duration
+	logicalOptimizers []logicalplan.Optimizer
 }
 
 func (e *compatibilityEngine) SetQueryLogger(l promql.QueryLogger) {
@@ -105,7 +100,7 @@ func (e *compatibilityEngine) NewInstantQuery(q storage.Queryable, opts *promql.
 	}
 
 	lplan := logicalplan.New(expr, ts, ts)
-	if !e.disableOptimizers {
+	if e.logicalOptimizers == nil {
 		lplan = lplan.Optimize(logicalplan.DefaultOptimizers)
 	}
 
@@ -144,8 +139,8 @@ func (e *compatibilityEngine) NewRangeQuery(q storage.Queryable, opts *promql.Qu
 	}
 
 	lplan := logicalplan.New(expr, start, end)
-	if !e.disableOptimizers {
-		optimizers := append(logicalplan.DefaultOptimizers, e.additionalOptimizers...)
+	if e.logicalOptimizers == nil {
+		optimizers := append(logicalplan.DefaultOptimizers, e.logicalOptimizers...)
 		lplan = lplan.Optimize(optimizers)
 	}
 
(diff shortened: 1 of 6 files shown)


engine/engine.go
```diff
@@ -35,6 +35,9 @@ type Opts struct {
 	// DisableOptimizers disables Query optimizations using logicalPlan.DefaultOptimizers.
 	DisableOptimizers bool
 
+	// AdditionalOptimizers are optimizers which are run in addition to the default ones.
+	AdditionalOptimizers []logicalplan.Optimizer
```
GiedriusS, line 39: I would vote to put everything under the `DisableOptimizers` flag. I would guess that a use case like [this](https://github.com/prometheus/prometheus/issues/8053#issuecomment-708677714) is probably rare. At least I haven't seen anything like this in many years of using Prometheus. Another thing is that that use case is really relevant only for remote read in Prometheus i.e. when Prometheus retrieves data from other Prometheus nodes. In the case of Thanos, that can only be hit when push-down happens, when we execute a given query directly on Prometheus. That would be my 2 cents :smile:

yeya24: Yeah I think this one should be safe. But keeping it as an array rather than a Boolean should be fine since the engine is only used as a library? If Cortex/Mimir/Thanos doesn't want to enable that they can just expose that as flag or disable it directly

GiedriusS: 👍 sounds good.

fpetkovski: Hm so I did not fully understand the conclusion. Do we want to have an array of optimizers, or do we want to have a boolean flag for "optional" optimizers?

GiedriusS: As far as I understood it, there should be an array internally so that library users could selectively enable/disable an optimizer but only a boolean flag should be exposed to end-users of software that uses the library like Thanos.

fpetkovski: Is there some value in allowing users to inject optimizers?

logicalplan/propagate_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+	lhSelector, ok := binOp.LHS.(*parser.VectorSelector)
+	if !ok {
+		return
+	}
+	rhSelector, ok := binOp.RHS.(*parser.VectorSelector)
+	if !ok {
+		return
+	}
+	// This case is handled by MergeSelectsOptimizer.
+	if lhSelector.Name == rhSelector.Name {
```
yeya24, line 55: What about queries like `node_filesystem_files{host="$host", mountpoint="/"} - node_filesystem_files_free / node_filesystem_files_free`? Basically LHS or RHS could be another vector

fpetkovski: I tried to cover this case but it's not as simple to implement. Maybe we can iterate on it in a separate PR?

engine/engine.go
GiedriusS: I think it will be useful for promql-engine developers to be able to selectively enable/disable optimizers e.g. when debugging something. We always add the default ones [here](https://github.com/thanos-community/promql-engine/pull/115/files#diff-2e6c4934f63ff9b712c2c346b33036af4724adf70b0801fff9b74f71b37fcd89R140) so I think the current version is OK, we are still able to disable what is not needed when debugging :smile:

GiedriusS (approved):

engine/engine.go
GiedriusS: @fpetkovski ping :point_up:

fpetkovski: Ah missed this. So if logical optimizers is empty, do we just use the default optimizers?

GiedriusS: I think so :smile:

engine/engine.go
```diff
@@ -139,8 +139,9 @@ func (e *compatibilityEngine) NewRangeQuery(q storage.Queryable, opts *promql.Qu
 	}
 
 	lplan := logicalplan.New(expr, start, end)
-	if !e.disableOptimizers {
-		lplan = lplan.Optimize(logicalplan.DefaultOptimizers)
+	if e.logicalOptimizers == nil {
```
GiedriusS, line 142: Shouldn't this be the other way around i.e. `if e.logicalOptimizers != nil { apply them }`? Also, we need to take default optimizers into account. So, it should be something like the following if I am not mistaken:

```
var optimizers []logicalplan.Optimizers

if e.logicalOptimizers == nil {
  optimizers = append(optimizers, logicalplan.DefaultOptimizers)
}

optimizers = append(optimizers, e.logicalOptimizers...)
lplan = lplan.Optimize(optimizers)
``` 

And the same logic is missing in instant query function? Perhaps all of this logic could be moved into a separate function to avoid making a mistake?

engine/engine.go
```diff
@@ -39,8 +39,8 @@ const (
 type Opts struct {
 	promql.EngineOpts
 
-	// DisableOptimizers disables Query optimizations using logicalPlan.DefaultOptimizers.
-	DisableOptimizers bool
+	// LogicalOptimizers are optimizers which are run in addition to the default ones.
```
GiedriusS, line 42: The comment looks wrong? 
```suggestion
	// LogicalOptimizers are optimizers that are run if the value is not nil. If it is nil then the default optimizers are run. Default optimizer list is available in the logicalplan package.
```

logicalplan/plan.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 )
 
+var NoOptimizers = []Optimizer{}
```
GiedriusS, line 13: Do we need to export these two? :thinking:

engine/engine.go
fpetkovski: Good catch, thanks 👍 

I've added a method to determine which logical optimizers to use based on the input and inject them in the query. The query then just applies whatever it has in the logical optimizers variable.

logicalplan/plan.go
fpetkovski: I thought we could export `NoOptimizers` as a helper to disable all optimizers, including default ones. I guess the `optimizer` interface can be private if that's what you had in mind.

bwplotka: I think it's fine for now 👍🏽

logicalplan/plan.go
```diff
@@ -10,6 +10,10 @@ import (
...
+
+var Optimizers = append(DefaultOptimizers, PropagateMatchersOptimizer{})
```
bwplotka, line 15: What's this? Can we add commentary?

bwplotka (approved):
LGTM!

Minor need of comments, but it's fine to do it later. Thanks!

logicalplan/plan.go
fpetkovski: I renamed it to `AllOptimizers` to make it more clear.

PR #116, opened 2022-11-04
Fix aggregations when argument is not a scalar

alanprot:
Fix https://github.com/thanos-community/promql-engine/issues/108

### DIFF
--- a/execution/aggregate/hashaggregate.go
+++ b/execution/aggregate/hashaggregate.go
@@ -116,14 +116,17 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 	}
 	defer a.next.GetPool().PutVectors(in)
 
-	var arg float64
+	var args []model.StepVector
 
 	if a.paramOp != nil {
-		p, err := a.paramOp.Next(ctx)
+		args, err = a.paramOp.Next(ctx)
 		if err != nil {
 			return nil, err
 		}
-		arg = p[0].Samples[0]
+
+		if len(args) != len(in) {
+			return nil, fmt.Errorf("scalar argument not found")
+		}
 	}
 
 	a.once.Do(func() { err = a.initializeTables(ctx) })
@@ -133,7 +136,13 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 
 	result := a.vectorPool.GetVectorBatch()
 	for i, vector := range in {
-		if err := a.workers[i].Send(arg, vector); err != nil {
+		if a.paramOp == nil {
+			err = a.workers[i].Send(0, vector)
+		} else {
+			err = a.workers[i].Send(args[i].Samples[0], vector)
+		}
+
+		if err != nil {
 			return nil, err
 		}
 	}
--- a/execution/aggregate/hashaggregate.go
+++ b/execution/aggregate/hashaggregate.go
@@ -22,7 +22,8 @@ import (
 )
 
 type aggregate struct {
-	next model.VectorOperator
+	next    model.VectorOperator
+	paramOp model.VectorOperator
 
 	vectorPool *model.VectorPool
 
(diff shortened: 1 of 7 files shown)


alanprot:
Hi..

First i tried to create a new Operator as suggested [here](https://github.com/thanos-community/promql-engine/pull/107#issuecomment-1299614682) but i was getting log of duplicated code.

So i think for this change we should keep the same operator.. and i maybe revisit it when implementing the `topk` and `bottomK` aggregations.

@fpetkovski @saswatamcode WDYT?

yeya24 (approved):
Thanks! LGTM

execution/aggregate/hashaggregate.go
```diff
@@ -130,13 +134,25 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+	if a.paramOp != nil {
+		p, err := a.paramOp.Next(ctx)
+
+		if err != nil {
+			return err
+		}
+		arg = p[0].Samples[0]
+	}
+
+	a.newAccumulator, err = makeAccumulatorFunc(a.aggregation, arg)
```
saswatamcode, line 154: Would be good to check the error here.

execution/aggregate/hashaggregate.go
```diff
@@ -130,13 +134,25 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+func (a *aggregate) initialize(ctx context.Context) error {
 	var (
 		tables []aggregateTable
 		series []labels.Labels
+		arg    float64
 		err    error
 	)
 
+	if a.paramOp != nil {
+		p, err := a.paramOp.Next(ctx)
```
saswatamcode, line 146: Let's call `a.paramOp.Next` within `Next` instead of `initialize`, similar to how we do it for `a.next`. Here we're calling `Next` even during `Series` calls which might not be ideal. 🙂

saswatamcode (commented):
Thanks for great work! 🙂 

Some comments!

execution/aggregate/hashaggregate.go
alanprot: Humm...

Can `Next` be called 2 times? I'm wondering if i need to make sure `a.paramOp.Next` is only called once.

alanprot: Ok...

I change the PR to receive the `args` on the `aggregate` function:

```
aggregate(arg float64, vector model.StepVector)
```

I think it makes sense as the aggregations has indeed 2 parameters (the arg and the vector)

saswatamcode (approved):
LGTM!

execution/aggregate/hashaggregate.go
saswatamcode: Maybe you can look at benchmarks and see if this affects it in anyway?

alanprot: Seems pretty much the same:

Before: https://gist.github.com/alanprot/bb096702ce42550c75b1aef534bcb7f2
After: https://gist.github.com/alanprot/b6981295d587a81763144cdd5e35c879

alanprot:
@fpetkovski WDYT?

execution/aggregate/scalar_table.go
```diff
@@ -39,19 +40,23 @@ func newScalarTables(stepsBatch int, inputCache []uint64, outputCache []*model.S
...
-	t.reset()
+func (t *scalarTable) aggregate(arg float64, vector model.StepVector) {
+	if len(t.accumulators) == 0 {
+		t.accumulators = make([]*accumulator, len(t.outputs))
+		for i := 0; i < len(t.accumulators); i++ {
+			t.accumulators[i] = t.newAccumulator(arg)
+		}
+	} else {
+		t.reset()
+	}
```
alanprot, line 58: I think adding the `arg` as argument to the `accumulator reset` method and keep creating them as before - on the tables initialization)

(In theory i could even use the same accumulator instance to to another quantile calculation using a different value of `q`)

@saswatamcode WDYT?

alanprot:
This PR still have a little issue when we have a range query and the argument is changing the value between steps..

Similar thing was addressed on https://github.com/thanos-community/promql-engine/pull/107.. i will update this PR soon.

execution/aggregate/scalar_table.go
```diff
@@ -39,19 +40,23 @@ func newScalarTables(stepsBatch int, inputCache []uint64, outputCache []*model.S
...
+		inputs:         inputSampleIDs,
+		outputs:        outputs,
+		newAccumulator: newAccumulator,
 	}
 }
 
-func (t *scalarTable) aggregate(vector model.StepVector) {
-	t.reset()
+func (t *scalarTable) aggregate(arg float64, vector model.StepVector) {
+	if len(t.accumulators) == 0 {
```
fpetkovski, line 51: Why do we need to do this here? I thought we initialize tables in the Series call.

alanprot: Yeah.. i just commented about this: https://github.com/thanos-community/promql-engine/pull/116#discussion_r1017399624

The reason to initialize here was basically because the `arg` may not be calculated when calling `Series`.

But i wanna change this and put the argument on the `reset` method... And this make sense as the argument in theory value can change from one step (and so we should update it on every reset).

Ex: if we have a range query like `quantiles(scalar(min(something)),some_other_thing)`

fpetkovski: Ok got it. I think this makes sense 👍

alanprot: Updated :D

fpetkovski:
Thanks, this LGTM. I think there are some potential improvements, but I'll address those in a separate PR.

PR #117, opened 2022-11-05
Add continuous benchmark action for the new engine

yeya24:

Fixes https://github.com/thanos-community/promql-engine/issues/48

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -4,7 +4,6 @@ on:
     branches:
       - main
     tags:
-  pull_request:
 
 permissions:
   contents: write
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,42 @@
+name: Continuous Benchmark
+on:
+  push:
+    branches:
+      - main
+    tags:
+  pull_request:
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Continuous Benchmark
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


yeya24:
Seems a token is required to push this repo.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,42 @@
+name: Continuous Benchmark
+on:
+  push:
+    branches:
+      - main
+    tags:
+  pull_request:
```
fpetkovski, line 7: I don't know if this can work, it will mean that any contributor will be able to get write access to the repo. Maybe we should only run this on main?

yeya24: Yeah I need to separate that. If it is a pr then having a benchmark result on pr comment is still nice, but shouldn't update the page

fpetkovski: Maybe we can remove this for now and see what the result looks like on main.

fpetkovski (approved):

fpetkovski:
Let's see how this works.

yeya24:
Seems the workflow succeeded. @fpetkovski Could you help enable Github pages for this repo and use config like that. We should be able to see it then

image

fpetkovski:
Worked https://thanos-community.github.io/promql-engine/dev/bench/

yeya24:
@fpetkovski Cool, can we add this to the Readme?

GiedriusS:
Nice! Maybe I can try working on adding some dedicated GH actions runners because I'm afraid that the results will be a bit random now depending on how loaded the shared workers are :joy:

PR #118, opened 2022-11-10
.github: use self-hosted runner

GiedriusS:
I started a small runner on Equinix using CNCF's resources (c3.small.x86). Let's use it to have consistent results.


GiedriusS:
Let me try if this works

PR #121, opened 2022-11-10
Adding type on compatibilityQuery struct

alanprot:
Just to make things a little more clear :D

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -29,6 +29,13 @@ import (
 	"github.com/thanos-community/promql-engine/logicalplan"
 )
 
+type QueryType int
+
+const (
+	InstantQuery QueryType = 1
+	RangeQuery   QueryType = 2
+)
+
 type Opts struct {
 	promql.EngineOpts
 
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -29,6 +29,13 @@ import (
 	"github.com/thanos-community/promql-engine/logicalplan"
 )
 
+type QueryType int
+
+const (
+	InstantQuery QueryType = 1
```
yeya24, line 35: Is it really needed for clarity? I think Prometheus does the same thing to check time.
And I am not sure whether we want to expose this or not. Any usecase?

alanprot: Checks like this is just very confusing, and the proof of that is that the comment in the code is wrong! hahah at least i was confused and had to run a unit test to figure out.

https://github.com/thanos-community/promql-engine/blob/8afe40d8945907e7ba1f15b148c625803f2a22c3/engine/engine.go#L255-L256

https://github.com/thanos-community/promql-engine/blob/8afe40d8945907e7ba1f15b148c625803f2a22c3/engine/engine.go#L177

fpetkovski (approved):

yeya24:
Thanks!

PR #122, opened 2022-11-10
Enabling Scalar Tests

alanprot:
Comparing json allow us to enable those tests.

I wanted to do a change on scalar and having those tests enabled give us more confidence to do the change.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -5,6 +5,7 @@ package engine_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"runtime"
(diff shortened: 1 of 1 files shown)


yeya24 (approved):
LGTM

engine/engine_test.go
```diff
@@ -2132,3 +2128,11 @@ func TestEngineRecoversFromPanic(t *testing.T) {
 	})
 
 }
+
+func jsonEqual(t *testing.T, expected interface{}, actual interface{}) {
+	e, err := json.Marshal(expected)
```
saswatamcode, line 2133: I think this could lead to more problems while checking equality. JSON doesn't support NaN or Inf (https://github.com/golang/go/issues/3480). There's also stuff with ordering (https://github.com/golang/go/issues/6244) that could end up as confusing bugs, as labels are marshaled as a map.

Maybe not handling all those complications here and just comparing structs recursively would be better? I raised https://github.com/efficientgo/core/pull/3 a few days ago to add support for `NaN`. I'll ping @bwplotka for review, and maybe that can be used here instead? WDYT? 🙂

fpetkovski: If we are blocked by that PR, maybe there's a way to add these cases in a separate test and use a custom comparison function?

saswatamcode: Yup! We can do that.

fpetkovski: Now that https://github.com/efficientgo/core/pull/3 is merged, can we enable the tests for `NaN` values?

MichaHoffmann: Couldnt we just compare `oldResult.String()` and `newResult.String()` for string equality?

PR #124, opened 2022-11-10
Simplify Scalar Calculation

alanprot:
Seems we have a special case on the engine to handle the scalar function.

The reason for that is because currently this for this function, the operator returns `nil` for get series where in reality it can return always a single series with an empty labelset.

This was built on top of: https://github.com/thanos-community/promql-engine/pull/122 (so the scalars testes are running on the PR).

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -216,29 +216,6 @@ loop:
 				break loop
 			}
 
-			// Case where Series call might return nil, but samples are present.
-			// For example scalar(http_request_total) where http_request_total has multiple values.
-			if len(series) == 0 && len(r) != 0 {
-				numSeries := 0
-				for i := range r {
-					numSeries += len(r[i].Samples)
-				}
-
-				series = make([]promql.Series, numSeries)
-
-				for _, vector := range r {
-					for i := range vector.Samples {
-						series[i].Points = append(series[i].Points, promql.Point{
-							T: vector.T,
-							V: vector.Samples[i],
-						})
-					}
-					q.Query.exec.GetPool().PutStepVector(vector)
-				}
-				q.Query.exec.GetPool().PutVectors(r)
-				continue
-			}
-
 			for _, vector := range r {
 				for i, s := range vector.SampleIDs {
 					series[s].Points = append(series[s].Points, promql.Point{
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -5,6 +5,7 @@ package engine_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"runtime"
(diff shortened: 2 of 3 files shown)


engine/engine.go
```diff
@@ -216,29 +216,6 @@ loop:
 				break loop
 			}
 
-			// Case where Series call might return nil, but samples are present.
-			// For example scalar(http_request_total) where http_request_total has multiple values.
-			if len(series) == 0 && len(r) != 0 {
```
saswatamcode, line 221: So I think this case was added specifically for handling cases where `series==nil`. With this change, if you compare structs in the `scalar func with NaN` case you'd see that they are unequal not only due to `NaN` but also due to labels as one in `labels.Labels(nil)` and one is `labels.Labels{}`. 

```
exp: &promql.Result{Err:error(nil), Value:promql.Matrix{promql.Series{Metric:labels.Labels(nil), Points:[]promql.Point{promql.Point{T:0, V:NaN}, promql.Point{T:30000, V:NaN}, promql.Point{T:60000, V:NaN}, promql.Point{T:90000, V:NaN}, promql.Point{T:120000, V:NaN}, promql.Point{T:150000, V:NaN}, promql.Point{T:180000, V:NaN}, promql.Point{T:210000, V:NaN}, promql.Point{T:240000, V:NaN}}}}, Warnings:storage.Warnings(nil)}
        
got: &promql.Result{Err:error(nil), Value:promql.Matrix{promql.Series{Metric:labels.Labels{}, Points:[]promql.Point{promql.Point{T:0, V:NaN}, promql.Point{T:30000, V:NaN}, promql.Point{T:60000, V:NaN}, promql.Point{T:90000, V:NaN}, promql.Point{T:120000, V:NaN}, promql.Point{T:150000, V:NaN}, promql.Point{T:180000, V:NaN}, promql.Point{T:210000, V:NaN}, promql.Point{T:240000, V:NaN}}}}, Warnings:storage.Warnings(nil)}
```

As we compare JSON here, it's not caught. Which might be fine, just wondering if this breaks anything?

alanprot: Hummm.. I see.. 

Just some context why I was looking into this:  https://github.com/thanos-community/promql-engine/issues/120

what I was thinking is if the ‘series’ call returns an empty array, we would keep the order returned by the ‘next’

but the the currently behavior when series returns an empty array and next returns something is to not add the labels at all.. :(

alanprot: @fpetkovski wdyt?

alanprot: Ok… so we should move the discussion to the issue.. but u think we should return another value on the “series” call to hint the engine ?

fpetkovski: I was thinking either that, or make the engine inspect the query independently and decide how to sort the end result. The latter seems to be simpler to implement because there is less code to change, but not sure if it leads to some messy code.

fpetkovski: @alanprot can we add a `sortByValues` flag to the `compatibilityEngine` which can toggle sorting the end result by values and not series?

We can set this flag to true only for instant queries based on whether the `lplan` here is `topk/bottomk`: https://github.com/thanos-community/promql-engine/blob/main/engine/engine.go#L110

alanprot: Only `sortByValues` does not work for the cases:
```
topk(2, http_requests_total) by (series)
```

In this case the result is sorted by value but per series.. =/


If i respect the order returned by the next call it would work.

PR #125, opened 2022-11-11
Update prometheus/prometheus to v0.40.1

fpetkovski:
This commit updates the prometheus/prometheus dependency to v0.40.1 which adds support for native histograms in TSDB and PromQL. The commit also unblocks updating this dependency in upstream Thanos.

Since this engine does not yet support querying native histograms, those cases will return an error for the time being. Unfortunately the error has to be emitted during query execution since the histogram_quantile function is now overloaded to work with both native and counter-based histograms.

Support for native-histograms in this engine will be added as a follow up.


pedro-stanaka (approved):
LGTM, small nit only.

engine/engine_test.go
```diff
@@ -2067,16 +2068,32 @@ type slowIterator struct {
...
-	return true
+	return chunkenc.ValFloat
 }
-func (d *slowIterator) Seek(t int64) bool {
+
+func (d *slowIterator) Seek(t int64) chunkenc.ValueType {
 	<-time.After(10 * time.Millisecond)
 	d.ts = t
-	return true
+	return 1
```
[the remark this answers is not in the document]
fpetkovski: Good point, modified.

fpetkovski:
Looks like `faillint` is also failing in the latest prometheus/prometheus. Not sure how we can resolve this, maybe there is a way to relax these checks? @saswatamcode maybe you have some ideas.

saswatamcode:
I think is might already be fixed in `faillint`, can you try upgrading it? `bingo get github.com/fatih/faillint@latest`

fpetkovski:
Thianks, this seems to have solved it 👍

README.md
```diff
@@ -13,7 +13,7 @@ The following table shows operations which are currently supported by the engine
 | Type                   | Supported                                                                 | Priority |
 |------------------------|---------------------------------------------------------------------------|----------|
 | Binary expressions     | Full support                                                              |          |
-| Histogram quantile     | Full support                                                              |          |
+| Histogram quantile     | Partial support (no support for native histograms)                        | High     |
```
yeya24, line 16: Priority High means it is high for native histogram? Or what else are left for Histogram quantile?

saswatamcode (approved):

README.md
fpetkovski: It means high for native histograms because now we will fail queries on timeseries that have native histograms in their chunks.

yeya24: I see. I don't know if we need to make it high because even Thanos doesn't support the native histogram feature and almost no client has supported instrumenting  native histograms. But I think it is fine to mark high in the doc.

yeya24 (approved):
Thanks

PR #128, opened 2022-11-24
Fix join in binary-operator

fpetkovski:
The binary operator assumes that only one low-cardinality series can join any given high-cardinality series since PromQL does not allow many-to-many joins. However, this rule is true only within individual steps. Two low-cardinality time-series can still join the same high-cardinality series if they do not have overlapping samples inside evaluation steps. This can happen during process restarts, where one time-series ends and new one starts.

image

This commit fixes the issue by properly calculating the output buffer for the low-cardinality side in a binop. The case is hard to test right now because the testing framework does not allow for defining offsets for time-series. I tested this in one of our environments and I could not reproduce the original issue anymore.

Fixes https://github.com/thanos-community/promql-engine/issues/127

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -65,17 +65,6 @@ func TestVectorSelectorWithGaps(t *testing.T) {
 
 }
 
-func storageWithSeries(series storage.Series) *storage.MockQueryable {
-	seriesSet := &testSeriesSet{series: series}
-	return &storage.MockQueryable{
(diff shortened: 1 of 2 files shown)


fpetkovski:
I managed to add a repro test case with more verbose mocks. The test fails on main but succeeds on this branch. This should also help us with keeping things correct while working on https://github.com/thanos-community/promql-engine/issues/32

PR #131, opened 2022-11-30
Implement error reporting in binary operator for many-to-many joins

fpetkovski:
The engine currently does not report many-to-many matching errors and will likely silently return an incorrect result for such binary operations.

This commit adds support for this detection, similar to how the Prometheus engine reports these errors.

Fixes https://github.com/thanos-community/promql-engine/issues/32

It is not simple to directly compare error messages in tests because the order of the series can be different. For example, we can get failures like this:

expected:
<code>
&errors.errorString{s:"found duplicate series for the match group {code=\"200\"} on the right hand-side of the operation: [{__name__=\"bar\", code=\"200\", method=\"post\"}, {__name__=\"bar\", code=\"200\", method=\"get\"}];many-to-many matching not allowed: matching labels must be unique on one side"}
</code>

got:
<code>
&errors.errorString{s:"found duplicate series for the match group {code=\"200\"} on the right hand-side of the operation: [{__name__=\"bar\", code=\"200\", method=\"get\"}, {__name__=\"bar\", code=\"200\", method=\"post\"}];many-to-many matching not allowed: matching labels must be unique on one side"}
</code>

### DIFF
--- a/execution/binary/table.go
+++ b/execution/binary/table.go
@@ -4,7 +4,6 @@
 package binary
 
 import (
-	"fmt"
 	"math"
 
 	"github.com/prometheus/prometheus/promql/parser"
@@ -24,10 +23,6 @@ type errManyToManyMatch struct {
 	sampleID          uint64
 	duplicateSampleID uint64
 	side              binOpSide
-}
-
-func (err errManyToManyMatch) Error() string {
-	return fmt.Sprintf("many-to-many match on %s", err.side)
 }
 
 func newManyToManyMatchError(sampleID, duplicateSampleID uint64, side binOpSide) *errManyToManyMatch {
--- a/execution/binary/table.go
+++ b/execution/binary/table.go
@@ -4,6 +4,7 @@
 package binary
 
 import (
+	"fmt"
 	"math"
 
 	"github.com/prometheus/prometheus/promql/parser"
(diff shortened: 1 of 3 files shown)


execution/binary/table.go
```diff
@@ -12,9 +13,37 @@ import (
...
+	rhBinOpSide binOpSide = "right"
+)
+
+type errManyToManyMatch struct {
+	sampleID          uint64
+	duplicateSampleID uint64
+	side              binOpSide
+}
+
+func (err errManyToManyMatch) Error() string {
```
saswatamcode, line 29: Are we using this anywhere?

execution/binary/vector.go
```diff
@@ -163,9 +169,26 @@ func (o *vectorOperator) Next(ctx context.Context) ([]model.StepVector, error) {
...
+
+			var sampleID, duplicateSampleID labels.Labels
+			switch err.side {
+			case lhBinOpSide:
+				sampleID = o.lhSampleIDs[err.sampleID]
+				duplicateSampleID = o.lhSampleIDs[err.duplicateSampleID]
+			case rhBinOpSide:
+				sampleID = o.rhSampleIDs[err.sampleID]
+				duplicateSampleID = o.rhSampleIDs[err.duplicateSampleID]
+			}
```
saswatamcode, line 187:
```suggestion
			sampleID = o.lhSampleIDs[err.sampleID]
			duplicateSampleID = o.lhSampleIDs[err.duplicateSampleID]
			if err.side == rhBinOpSide {
				sampleID = o.rhSampleIDs[err.sampleID]
				duplicateSampleID = o.rhSampleIDs[err.duplicateSampleID]
			}
```

saswatamcode (approved):
Thanks for fixing! LGTM! 🙂 

Some small suggestions.

execution/binary/vector.go
fpetkovski: I did have this before, but I find the switch statement a bit more explicit and easier to read

execution/binary/table.go
fpetkovski: Good point. Initially this struct was implementing the error interface, but I don't think it's needed in the current version. Removed.

execution/binary/vector.go
saswatamcode: Yeah I was thinking in "left-right" terms, but you're right, the switch is better for readability here!

saswatamcode (approved):

PR #133, opened 2022-12-01
execution: add support for KeepBool

GiedriusS:
Add support for "KeepBool"
https://prometheus.io/docs/prometheus/latest/querying/operators/:

tl;dr of the rules:
- If `bool` modifier is provided between scalars then the result is either 0 or 1
- If `bool` modifier is provided between vectors/scalars then the result is 0 if the metric would've been dropped; `__name__` is removed.
- If `bool` modifier is provided between vectors then the result is 0 if the metric would've been dropped; `__name__` is removed.


fpetkovski (approved):
TIL 😄 

Thanks, lgmt.

yeya24 (approved):
Thanks!

PR #135, opened 2022-12-07
engine: make histogram_quantile resilient against duplicate output IDs

GiedriusS:
`o.seriesBuckets[i]` can be `nil` if multiple input series IDs map to the output series ID. In such cases, we need to skip over them.

### DIFF
--- a/execution/function/histogram.go
+++ b/execution/function/histogram.go
@@ -134,8 +134,12 @@ func (o *histogramOperator) processInputSeries(vectors []model.StepVector) ([]mo
 
 		step := o.pool.GetStepVector(vector.T)
 		for i, stepBuckets := range o.seriesBuckets {
-			// We only have one bucket so it needs to be NaN.
-			// Or, if we are after how many scalar points we have.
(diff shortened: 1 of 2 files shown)


execution/function/histogram.go
```diff
@@ -134,8 +134,12 @@ func (o *histogramOperator) processInputSeries(vectors []model.StepVector) ([]mo
 
 		step := o.pool.GetStepVector(vector.T)
 		for i, stepBuckets := range o.seriesBuckets {
-			// We only have one bucket so it needs to be NaN.
-			// Or, if we are after how many scalar points we have.
+			// It could be zero if multiple input series map to the same output series ID.
+			if len(stepBuckets) == 0 {
```
fpetkovski, line 138: Ah so this is what we miss by removing the check for`len(stepBuckets) < 2` :)

fpetkovski (approved):

execution/function/histogram.go
GiedriusS: Yeah :smile:

PR #136, opened 2022-12-07
Fix topk/bottomk on empty inputs

fpetkovski:
TopK and bottomK can take in empty input vectors in certain cases, which can lead to validation failing in the kHashAggregate operator.

This commit changes the validation to check if the input vector batch has less steps than the scalar vector batch.


alanprot:
Thx :D

alanprot (approved):

saswatamcode (approved):
Thanks!

PR #137, opened 2022-12-08
Fetch Series in parallel for vector operators

csmarchbanks:
Fixes a TODO in the code, just fetching the Series can be slow in a PromQL server I run and running the operations in parallel can halve the response time.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -8,8 +8,10 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"reflect"
 	"runtime"
 	"sort"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


csmarchbanks:
Hmm, test failure seems like it could be real, but I am not exactly sure why. I will take a look more tomorrow but any pointers are also welcome.

fpetkovski:
Interesting, I thought there was some race condition with how we propagate hints. Feel free to post here if you hit a wall.

csmarchbanks:
I'm not sure I love it, but I just update the test to not care about the order in which the hints are sent since the Select statements will be done in any order. Let me know what you think, my other thought was to change the hints to a map by matcher, but passing the matcher for every test case would get quite tedious.

fpetkovski (approved):
Thanks, I think this is good enough for now. 

We can improve the test in a follow up PR.

GiedriusS (approved):

PR #139, opened 2023-01-03
Implement distributed execution

fpetkovski:
This PR allows creating an instance of the engine capable of distributed execution.

The detailed proposal can be found here: https://github.com/thanos-io/thanos/pull/6012.
Deduplication will be added as a follow up.


logicalplan/plan_test.go
```diff
@@ -132,3 +136,128 @@ func TestMatcherPropagation(t *testing.T) {
...
+		{
+			name: "avg",
+			expr: `avg by (pod) (http_requests_total)`,
+			expected: `
+avg by (pod) (
+  coalesce(
+    remote(http_requests_total),
+    remote(http_requests_total)
+  )
+)`,
```
yeya24, line 166: I am wondering for this query, do we want to do distributed execution or not? To me this seems better to just execute locally without querying remote engines. Since we are querying raw series, remote engines are just stores in this case

fpetkovski: I also thought about this, and the reason I thought it's beneficial to distribute the query is because PromQL is done over a limited number of steps. So by distributing the query we would "downsample" on the fly. When querying longer ranges, the number of samples going to the central querier will be much smaller than if we directly fetch chunks from storage. 

For example, the resolution for a 30d query is 10368s. So we will remove a significant amount of samples because remote engines will sample the data.

yeya24: Got it that makes sense to me. But one step further I am wondering if it is better to do this when retrieving series from stores. Like https://github.com/thanos-io/thanos/pull/4857

GiedriusS: The way it is currently implemented we would get the benefit of downsampling even if there is no function involved so it should be much better.

fpetkovski: > But for remote engines, all the raw samples are queried into memory, right? 

Yes, this is correct. However solving that at the storage level seems a bit more complicated and maybe something to tackle independently in the future.

fpetkovski:
Should we merge this and continue iterating on main? The newly added code should be unused, so it won't affect anyone using the engine.

GiedriusS (approved):
:+1:

PR #141, opened 2023-01-05
function: add support for time()

GiedriusS:
Add support for time() function. Add special handling for functions that take no input and just return something for each step. It will be used in the future for other date functions.


fpetkovski (approved):
Looks great, thanks!

saswatamcode (approved):
Thanks!

PR #142, opened 2023-01-05
function: add support for more functions

GiedriusS:
Add various math functions and tests for them. Fix fuzzing test to ignore functions that do not consume a matrix.


fpetkovski (approved):
Thanks, lgtm!

I think we might want to reorganize tests a bit in the future, we seem to have accumulated lots of repetitive code and it's hard to know what is now tested and what isnt.

GiedriusS:
Good point, I've created https://github.com/thanos-community/promql-engine/issues/144 for this.

PR #151, opened 2023-01-16
Implement distributed deduplication

fpetkovski:
This commit adds support for deduplicating series/samples and applies it to the distributed query engine. This is done by implementing a Deduplication operator that removes samples with the same sample IDs from individual step vectors. Samples are deduplicated by a last-sample-wins strategy.

In order to properly prioritize non-stale samples in favor of stale ones, remote engines are sorted by their max time. This way if samples come from both older and newer engines, the ones from newer engines will be pprioritized. As a result, stale samples will be deduplicated in favour of non-stale ones, but will be preserved if non-stale samples are not present in a step vector.


logicalplan/distribute.go
```diff
@@ -62,6 +79,14 @@ type DistributedExecutionOptimizer struct {
 
 func (m DistributedExecutionOptimizer) Optimize(plan parser.Expr) parser.Expr {
 	engines := m.Endpoints.Engines()
+
+	// The Deduplicate operator will deduplicate samples using a last-sample-wins strategy.
+	// Sorting engines by max times ensures that samples produced due to staleness will be
+	// overwritten and corrected by samples coming from engines with a higher max time.
+	sort.Slice(engines, func(i, j int) bool {
+		return engines[i].MaxT() < engines[j].MaxT()
```
GiedriusS, line 87: I wonder if there's a way to move this somehow into the deduplication operator itself because I had to look at this to understand what is even the purpose of the deduplication operator :thinking: probably not but just throwing the idea out here

execution/exchange/dedup.go
```diff
@@ -0,0 +1,132 @@
...
+			outputSampleID := d.outputIndex[inputSampleID]
+			d.outputCache[outputSampleID].t = vector.T
+			d.outputCache[outputSampleID].v = vector.Samples[i]
+		}
+
+		out := d.pool.GetStepVector(vector.T)
+		for outputSampleID, sample := range d.outputCache {
+			if sample.t == vector.T {
+				out.SampleIDs = append(out.SampleIDs, uint64(outputSampleID))
+				out.Samples = append(out.Samples, sample.v)
```
GiedriusS, line 68: Don't we need to `break` here? :thinking:

logicalplan/distribute.go
```diff
@@ -99,7 +126,37 @@ func (m DistributedExecutionOptimizer) Optimize(plan parser.Expr) parser.Expr {
...
+	}
+
+	groupingLabels := make([]string, 0, len(groupingSet))
+	for lbl := range groupingSet {
+		groupingLabels = append(groupingLabels, lbl)
+	}
+
+	sort.Strings(groupingLabels)
+
+	remoteAggregation := *aggr
```
GiedriusS, line 154: Nit: maybe this can be simplified to:
```go
	aggr.Grouping = groupingLabels
	return aggr
```

I think the result should be the same.

logicalplan/distribute_test.go
```diff
@@ -0,0 +1,166 @@
...
+    remote(http_requests_total)
+  ))
+)))`,
+		},
+		{
+			name: "binary operation in the operand path",
+			expr: `max by (pod) (sort(metric_a / metric_b))`,
+			expected: `
+max by (pod) (sort(
+  dedup(coalesce(remote(metric_a), remote(metric_a)))
```
GiedriusS, line 98: In the future, it would be nice to add `Name() string` or a similar method to `api.RemoteEngine` so that we could use it in Explain() and other places.

GiedriusS (approved):
LGTM, just some comments/suggestions

logicalplan/distribute.go
fpetkovski: I agree, and I also noticed myself that there is now coupling between the logical plan and the deduplication logic. If we don't sort the result here, dedup can return stale results from an older engine. I'll look into how we can remove this dependency.

GiedriusS: Another thing is that I think this introduces a race between multiple Optimize() because `engines` is a slice and multiple `Engines()` will return the same slice :smile:

fpetkovski: That's also a good point. One thing to note is that stale samples from an "older" engine (like one attached to store-gw) will have the same timestamps as samples from a newer engine (prometheus/receiver). So in the dedup operator it's already too late to figure out which sample has priority because the information we need is lost. 

We can add some information to the samples themselves to associate them with an engine, but I'm concerned this will lead to memory bloat. Maybe we can add a field to the entire step vector, or even the vector batch and use it to prioritize samples in dedup.

fpetkovski: I've moved this sorting to `execution.go`, right before we create the dedup operator. This way it's closer to where dedup is happening.

logicalplan/distribute.go
fpetkovski: That would modify the input parameter which is used for the root aggregation. To avoid that, `remoteAggregation` should be a copy instead.

execution/exchange/dedup.go
fpetkovski: Hm that would cause only one sample from the dedup cache to be added to the output vector. I wanted to avoid using a new dedup slice for each step vector, so this `t` field helps determine whether a sample from the current step vector should be added to the output. We set the `t` field [here](https://github.com/thanos-community/promql-engine/pull/151/files#diff-15c9fb08ea6d60219b7b77d429edda47df09c757ee3ba2262c84226f13b4cafcR65). If `t` does not equal to the current time for the output vector, we skip the sample because it's from a previous step.

I added a comment to help explain this.

GiedriusS (approved):
LGTM, thanks!

PR #153, opened 2023-01-17
Avoid panic in query stats

yeya24:

If query stats is required, now the engine would just panic. This pr initializes required field for this API but doesn't implement the query stats feature, just to avoid panic for now.


fpetkovski (approved):
Thanks!

PR #154, opened 2023-01-18
add support for a few functions

kama910:
Signed-off-by: Kama Huang <kamatogo13@gmail.com>
Support date/time related functions.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1055,6 +1055,118 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			end:   time.Unix(60000, 0),
 		},
 		{
+			name: "days_in_month with input",
+			load: `load 30s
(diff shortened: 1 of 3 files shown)


fpetkovski:
This originates here: https://github.com/thanos-community/promql-engine/blob/3e3e0b3c510bae274b2af7bcf74eea40282c0425/execution/execution.go#L114-L116

I think you could even remove that check and see what happens.

yeya24:
@fpetkovski I see those functions are Variadic functions indeed. Shall we support Variadic function in our framework first and then tackle those promql functions?

fpetkovski:
I was hoping that we already support this because the function operator can take an arbitrary number of input operators. But I am not 100% sure if this would cover variadic functions.

yeya24 (approved):
Thanks for your contribution!

GiedriusS (approved):

PR #155, opened 2023-01-20
Enable NaN tests

fpetkovski:
This commit enables NaN tests by using the latest version of efficientgo/core.


saswatamcode (approved):
LGTM! 

But maybe we can split the NaN test cases into a separate testing function that uses the go-cmp option?

Also the test seems to timeout after 1m in CI

fpetkovski:
What would be the advantage of splitting the tests? Don't we always want to treat NaNs as equal?

saswatamcode:
The advantage I was thinking of was that we'd be aware that those cases return NaN and aren't comparing any numerical data. If not separate test function, then maybe just comments to highlight which test case returns NaN? 🙂 
I don't feel too strongly about this, so not a blocker!

fpetkovski:
Hm, so with native histograms we will have cases where we return a mix of NaNs and real data, so the result won't always be binary. Maybe we can detect in the test case whether there are NaNs and print this out? I would like to avoid another test function where we have to repeat things. Even the separation between range and instant queries is getting a bit hard to manage.

saswatamcode:
I think detecting and printing might be best!

fpetkovski:
Updated 👍

saswatamcode (approved):
Thanks! LGTM!

yeya24:
Just curious, I am seeing some NaN related errors in my tests. The old engine result has NaN while the new engine doesn't. Is this a bug of the new engine or a flaw of the NaN comparison code?

```
vector(--scalar(atan(irate({__name__="http_requests_total"}[1m] offset 4m21s))))
    engine_test.go:81: Applying comparison with NaN equality.
    testutil.go:91: engine_test.go:82: ""
        
        	exp: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:1677955629940, V:NaN, H:(*histogram.FloatHistogram)(nil)}, Metric:labels.Labels{}}}, Warnings:storage.Warnings(nil)}
        
        	got: &promql.Result{Err:error(nil), Value:promql.Vector{}, Warnings:storage.Warnings(nil)}
```

fpetkovski:
Yes, this is a bug probably in atan function but worth double checking. We can haven an issue to investigate.

PR #156, opened 2023-01-20
Fix instant query result sort for topk/bottomk

fpetkovski:
There is an implicit expectation that topk and bottomk instant queries should return values in sorted order. When a grouping is used for these aggregations, results should be sorted within individual groups.

The prometheus engine does not sort the groups themselves, which makes it hard to do exact comparisons. However, we can now compare results directly for these aggregations when no grouping is used.

Fixes https://github.com/thanos-community/promql-engine/issues/120.

cc @alanprot

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -6,6 +6,7 @@ package engine
 import (
 	"context"
 
+	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
Thanks! This makes sense! 💪🏻

alanprot:
Thanks!

PR #162, opened 2023-01-27
Handle no arg function labels

yeya24:

I am currently looking into https://github.com/thanos-community/promql-engine/pull/154 and see if I can help unblock it.

I found out those variadic functions can be supported, but there is some nuance here in the upstream promql implementation.

For functions that return scalar, `scalar`, `pi` and `time` will return samples with `Metric` set to nil, for example:
1. https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L431
2. https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L741
3. https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L57

For other functions added in https://github.com/thanos-community/promql-engine/pull/154, it is using an empty `Metric`, but not nil so the test will fail. https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L1165

I don't know why promql implements scalar functions this way but this PR should help unblock #154.

### DIFF
--- a/execution/function/operator.go
+++ b/execution/function/operator.go
@@ -42,21 +42,23 @@ type noArgFunctionOperator struct {
 	funcExpr    *parser.Call
 	call        FunctionCall
 	vectorPool  *model.VectorPool
+	series      []labels.Labels
+	sampleIDs   []uint64
 }
 
 func (o *noArgFunctionOperator) Explain() (me string, next []model.VectorOperator) {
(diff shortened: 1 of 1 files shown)


yeya24:
Kindly ping for review.

fpetkovski (approved):
Thanks, lgtm 👍

PR #163, opened 2023-01-27
Fixing problem on FilteredSelector when applying filter in series without filtered labels

pedro-stanaka:
## Summary

This PR fixes a problem where a label matcher would match a series if the label did not exist in the series.

So, for example:

```
metric{}
```

Would not match the selector `{foo~=".*"}`, where it should to be compatible with Prometheus.

See this [added test case](https://github.com/pedro-stanaka/promql-engine/blob/4e4a6b09bd2ecb2bce2279e31f1a8ed923935240/execution/storage/filter_test.go#L46-L51) for more details.

### DIFF
--- a/execution/storage/filter.go
+++ b/execution/storage/filter.go
@@ -45,14 +45,13 @@ func (f filter) Matches(series storage.Series) bool {
 	if len(f.matcherSet) == 0 {
 		return true
 	}
-	hasMatch := true
+
 	for name, m := range f.matcherSet {
 		label := series.Labels().Get(name)
 		if !m.Matches(label) {
-			hasMatch = false
-			break
+			return false
 		}
 	}
 
-	return hasMatch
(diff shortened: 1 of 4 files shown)


fpetkovski (approved):
Thanks!

execution/storage/filter.go
```diff
@@ -45,16 +45,14 @@ func (f filter) Matches(series storage.Series) bool {
...
-			continue
-		}
-		if !m.Matches(l.Value) {
-			return false
+	hasMatch := true
+	for name, m := range f.matcherSet {
+		label := series.Labels().Get(name)
+		if !m.Matches(label) {
+			hasMatch = false
+			break
```
fpetkovski, line 53: Maybe we can return `false` immediately here.

GiedriusS (approved):

PR #164, opened 2023-01-27
remove fmt.Sprintf() when writing to hash

yeya24:

Fixes https://github.com/thanos-community/promql-engine/issues/159 and use `strconv.Format` instead.

### DIFF
--- a/execution/storage/pool.go
+++ b/execution/storage/pool.go
@@ -4,7 +4,6 @@
 package storage
 
 import (
-	"fmt"
 	"strconv"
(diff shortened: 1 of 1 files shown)


yeya24:
There is no much different in benchmark result, probably we don't have much series. 
But from some other benchmarks we can see strconv.formatInt has better performance. https://gist.github.com/evalphobia/caee1602969a640a4530

fpetkovski (approved):
Thanks, TIL.

PR #165, opened 2023-01-30
Add link to continuous benchmark

yeya24:

This PR adds the continuous benchmark result to our readme. I don't think current solution is good enough, but okay as a temporary solution.

Would be great to integrate some bots like https://github.com/prometheus/test-infra/tree/master/funcbench for benchmarking.


fpetkovski (approved):
Good idea, thanks!

PR #169, opened 2023-02-02
Add tests for distributed instant queries

fpetkovski:
This commit adds tests for distributed instant queries. Since an instant query can be expressed as a range query with the same start and end timestamp, this commit also removes the NewInstantQuery method from the remote engine interface in order to keep the API smaller.


saswatamcode (approved):
LGTM!

PR #170, opened 2023-02-06
Update Prometheus to v0.42.0 and Prometheus common to latest

rabenhorst:
As preparation for updating Thanos to Prometheus v0.42.0 we need to update the promql engine first. Additionally prometheus common is updated to latest version.

> Note: It was necessary to add a workaround for literal tests since the original Prometheus engine returns inconsistent results depending on the literal type (nil labels vs. empty labels).


engine/engine_test.go
```diff
@@ -3262,3 +3264,15 @@ func roundValues(r *promql.Result) {
 		}
 	}
 }
+
+// nilEmptyLabels sets empty labelsets to nil to work around inconsistent
+// results from the old engine depending on the literal type (e.g. number vs. compare).
+func nilEmptyLabels(result *promql.Result) {
```
fpetkovski, line 3270:
```suggestion
func emptyLabelsToNil(result *promql.Result) {
```

fpetkovski (commented):
Thanks, just one small suggestion.

engine/bench_test.go
```diff
@@ -42,7 +42,7 @@ func BenchmarkChunkDecoding(b *testing.B) {
 			ss := querier.Select(false, nil, matcher)
 			series := make([]chunkenc.Iterator, 0)
 			for ss.Next() {
-				series = append(series, ss.At().Iterator())
+				series = append(series, ss.At().Iterator(nil))
```
fpetkovski, line 45: Is there a way we can take advantage of reusing these iterators? Should we open an issue to investigate?

rabenhorst: I guess it's something worth looking into. I will create an issue and link details from Prom repo.

PR #176, opened 2023-02-13
Add query timeout Support

### LINKED ISSUE #146: Support query timeout
The original promql engine supports specifying query timeout https://github.com/prometheus/prometheus/blob/main/promql/engine.go#L249.

It would be nice to have that feature in this engine as well.

PradyumnaKrishna:
Fixes #146 

Current engine doesn't support query timeout, which was a part of original promql-engine.
This Pull Request adds timeout option in `compatibilityEngine`, as our engine can't access
private data member `timeout` of the `promql.Engine`.

Testcase are updated as well.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -128,8 +128,7 @@ func New(opts Opts) *compatibilityEngine {
 	}
 
 	return &compatibilityEngine{
-		prom:    promql.NewEngine(opts.EngineOpts),
-		timeout: opts.Timeout,
+		prom: promql.NewEngine(opts.EngineOpts),
 		queries: promauto.With(opts.Reg).NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "promql_engine_queries_total",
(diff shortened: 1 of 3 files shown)


engine/engine.go
```diff
@@ -145,6 +146,7 @@ func New(opts Opts) *compatibilityEngine {
 
 type compatibilityEngine struct {
 	prom    *promql.Engine
+	timeout time.Duration
```
fpetkovski, line 149: Let's maybe move the timeout bellow the `logicalOptimizers` variable to group query parameters in one place?

engine/engine_test.go
```diff
@@ -2538,6 +2538,34 @@ func TestQueryCancellation(t *testing.T) {
 	testutil.Equals(t, context.Canceled, newResult.Err)
 }
 
+func TestQueryTimeout(t *testing.T) {
+
```
fpetkovski, line 2542:
```suggestion
```

engine/engine_test.go
```diff
@@ -2538,6 +2538,34 @@ func TestQueryCancellation(t *testing.T) {
...
+	end := time.Unix(120, 0)
+
```
fpetkovski, line 2544:
```suggestion
```

fpetkovski (commented):
This looks great, mostly code style comments from my side. 

Just to confirm, will all queries fail when timeout is set to 0? And is this how the Prometheus engine behaves as well?

PradyumnaKrishna:
> Just to confirm, will all queries fail when timeout is set to 0? And is this how the Prometheus engine behaves as well?

Yes, I have confirmed this by creating some testcase with Prometheus engine as well.

fpetkovski (approved):
Thanks, LGTM!

PR #177, opened 2023-02-14
Add support for histogram aggregates

fpetkovski:
This commit adds support for aggregating native histograms. The behavior is currently defined for the `sum`, `count` and `group` aggregations.

Aggregations like `avg`, `max`, `min`, etc.. currently return incorrect results in the Prometheus engine.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -3193,7 +3193,7 @@ func TestNativeHistogram(t *testing.T) {
 			query: "count by (foo) (native_histogram_series)",
 		},
 		// TODO(fpetkovski): The Prometheus engine returns an incorrect result for this case.
-		// Uncomment once it gets fixed.
+		// Uncomment once it gets fixed: https://github.com/prometheus/prometheus/issues/11973.
 		//{
 		//	name:  "max",
 		//	query: "max (native_histogram_series)",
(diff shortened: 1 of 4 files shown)


execution/aggregate/scalar_table.go
```diff
@@ -120,17 +138,29 @@ func makeAccumulatorFunc(expr parser.ItemType) (newAccumulatorFunc, error) {
...
+						return
+					}
+					histSum = histSum.Add(h)
+				},
+				ValueFunc: func() (float64, *histogram.FloatHistogram) {
+					return value, histSum
 				},
-				ValueFunc: func() float64 { return value },
-				HasValue:  func() bool { return hasValue },
+				// Sum returns an empty result when floats are histograms are aggregated.
```
GiedriusS, line 160: Do you mean this here?:
```suggestion
				// Sum returns an empty result when floats and histograms are aggregated.
```

engine/engine_test.go
```diff
@@ -3150,29 +3150,68 @@ func TestNativeHistogram(t *testing.T) {
...
+		{
+			name:  "count",
+			query: "count (native_histogram_series)",
+		},
+		{
+			name:  "count by (foo)",
+			query: "count by (foo) (native_histogram_series)",
+		},
+		// TODO(fpetkovski): The Prometheus engine returns an incorrect result for this case.
+		// Uncomment once it gets fixed.
```
GiedriusS, line 3196: Maybe there's some Prometheus issue for this? We could link it here.

GiedriusS (approved):
Just two suggestions otherwise LGTM

engine/engine_test.go
fpetkovski: Good point, let me reach out in the native-histograms-dev slack channel and see what the consensus is on these edge cases.

GiedriusS (approved):

PR #179, opened 2023-02-15
Add support for native histogram functions

fpetkovski:
This commit adds support for all native histogram functions which currently have a sane behavior in Prometheus. These include:
* histogram_sum
* histogram_count
* histogram_fraction
* histogram_quantile

These functions should be sufficient to allow the engine to work with native histograms.


saswatamcode (approved):
Thanks! LGTM!

PR #180, opened 2023-02-16
Fix failing benchmark tests

PradyumnaKrishna:
Recent merge #176 Add query timeout support broke the benchmark test. A timeout must be provided to initialize the new engine. It was breaking the tests with error "context deadline exceeded". Providing timeout as option in engine fixes this problem.


fpetkovski (approved):
Thanks for the quick fix 👍

PR #181, opened 2023-02-17
Allow configuration of query shards

douglascamata:
Taking a chance at solving https://github.com/thanos-community/promql-engine/issues/168.


engine/engine_test.go
```diff
@@ -2985,6 +2985,42 @@ func TestQueryStats(t *testing.T) {
 	stats.NewQueryStats(q.Stats())
 }
 
+func TestQueryConcurrency(t *testing.T) {
```
douglascamata, line 2988: This is a small test to ensure there's no difference in the results of queries depending on the number of shards used.

PR #172, opened 2023-02-09
add absent function

siddhikhapare:
Signed-off-by: siddhikhapare <siddhikhapare66@bvcoenm.edu.in>

Fixes: #138

PR #152, opened 2023-01-16
function: add support for absent()

zeus2611:
Signed-off-by: Nischay [itsnischay2604@gmail.com](mailto:itsnischay2604@gmail.com)

add support for absent(), related issue https://github.com/thanos-community/promql-engine/issues/138

Issue #138, opened 2022-12-14
Add support for all functions


fpetkovski:
This issue tracks which functions are left to be implemented before we have complete support for all PromQL functions.
The list of available functions in PromQL is here: https://prometheus.io/docs/prometheus/latest/querying/functions/

- [x] abs()
- [x] absent()
- [x] absent_over_time()
- [x] ceil()
- [x] changes()
- [x] clamp()
- [x] clamp_max()
- [x] clamp_min()
- [x] day_of_month()
- [x] day_of_week()
- [x] day_of_year()
- [x] days_in_month()
- [x] delta()
- [x] deriv()
- [x] exp()
- [x] floor()
- [x] histogram_count() and histogram_sum()
- [x] histogram_fraction()
- [x] histogram_quantile()
- [ ] holt_winters()
- [x] hour()
- [x] idelta()
- [x] increase()
- [x] irate()
- [x] label_join()
- [x] label_replace()
- [x] ln()
- [x] log2()
- [x] log10()
- [x] minute()
- [x] month()
- [x] predict_linear()
- [x] pi()
- [x] rate()
- [x] resets()
- [x] round()
- [x] scalar()
- [x] sgn()
- [x] sort()
- [x] sort_desc()
- [x] sqrt()
- [x] time()
- [x] timestamp()
- [x] vector()
- [x] year()
- [x] <aggregation>_over_time()
- [x] [Trigonometric Functions](https://prometheus.io/docs/prometheus/latest/querying/functions/#trigonometric-functions)

GiedriusS:
Why did you mark the function "rate()" as unfinished? :thinking:

zeus2611:
hey, i would like to start contributing to this as well. Can you guide me how i can get started.

fpetkovski:
@zeus2611 maybe you can look into adding support for the `absent` function? Start by adding a test case, and work your way to making the test case pass.

yeya24:
To support more functions, we probably need to refactor the function part a little bit. Now Variadic is not supported and I can see for functions like `predict_linear` and `holt_winters`, since the first parameter is a range vector, we don't pass the scalar parameters when executing the function.

fpetkovski:
Since the function operator takes a slice of operands, do we need anything special for variadic functions?

For the second issue, I agree that we need a better way to handle matrix selectors outside of a rate et al.

Althaf66:
[...]

PR #184, opened 2023-02-22
execution: add absent function

MichaHoffmann:
partially addresses #138 
duplicates #172 and #152 

I wanted to check out the execution model and started out implementing this function as excercise; this PR can be closed in favor of the other two, but since i implemented it already i thought i push it. Tests and implementation of label logic were basically taken from prometheus upstream docs and code. 

Code contains a bit of duplication with noArgsFunctionOperator.

```
RangeQuery/absent_and_exists-8                              123.54m ± ∞ ¹   14.15m ± ∞ ¹   -88.55% (p=0.008 n=5)
RangeQuery/absent_and_doesnt_exist-8                         191.0µ ± ∞ ¹   719.3µ ± ∞ ¹  +276.65% (p=0.008 n=5)
```

### DIFF
--- a/execution/function/operator.go
+++ b/execution/function/operator.go
@@ -90,6 +90,12 @@ func NewFunctionOperator(funcExpr *parser.Call, call FunctionCall, nextOps []mod
 			next:     nextOps[0],
 			funcExpr: funcExpr,
 		}, nil
+	case "absent":
+		return &absentOperator{
+			next:     nextOps[0],
+			pool:     model.NewVectorPool(stepsBatch),
+			funcExpr: funcExpr,
+		}, nil
 	}
 
 	// Short-circuit functions that take no args. Their only input is the step's timestamp.
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -260,6 +260,16 @@ func BenchmarkRangeQuery(b *testing.B) {
 			query: `sort_desc(http_requests_total)`,
 			test:  sixHourDataset,
 		},
+		{
+			name:  "absent and exists",
+			query: `absent(http_requests_total)`,
+			test:  sixHourDataset,
+		},
+		{
+			name:  "absent and doesnt exist",
+			query: `absent(nonexistent)`,
+			test:  sixHourDataset,
+		},
 	}
 
 	for _, tc := range cases {
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1768,6 +1768,17 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 				storage_info{storage_info="Swap", storage_index="30"} 1x20`,
 			query: `avg by (storage_info) (storage_used * on (instance, storage_index) group_left(storage_info) (sum by (instance, storage_index, storage_info) (storage_info)))`,
 		},
+		{
+			name: "absent with partial data in range",
+			load: `load 30s
+				existent{job="myjob"} 1 1 1`,
+			query: `absent(existent{job="myjob"})`,
+		},
+		{
+			name:  "absent with no data in range",
+			load:  `load 30s`,
+			query: `absent(nonexistent{job="myjob"})`,
+		},
 	}
 
 	disableOptimizerOpts := []bool{true, false}
@@ -3418,6 +3429,38 @@ func TestInstantQuery(t *testing.T) {
 				http_requests_total{pod="nginx-2", series="1"} -10+1x50
 				http_requests_total{pod="nginx-3", series="1"} NaN`,
 			query: "sgn(http_requests_total)",
+		},
+		{
+			name:  "absent and series does not exist",
+			load:  `load 30s`,
+			query: `absent(nonexistent{job="myjob"})`,
+		},
+		{
+			name: "absent and series exists",
+			load: `load 30s
+				existent{job="myjob"} 1`,
+			query: `absent(existent{job="myjob"})`,
+		},
+		{
+			name:  "absent and regex matcher",
+			load:  `load 30s`,
+			query: `absent(nonexistent{job="myjob", instance=~".*"})`,
+		},
+		{
+			name:  "absent and duplicate matchers",
+			load:  `load 30s`,
+			query: `absent(nonexistent{job="myjob", job="yourjob", foo="bar"})`,
+		},
+		{
+			name:  "absent and nested function",
+			load:  `load 30s`,
+			query: `absent(sum(nonexistent{job="myjob"}))`,
+		},
+		{
+			name: "absent and nested absent with existing series",
+			load: `load 30s
+				existent{job="myjob"} 1`,
+			query: `absent(absent(existent{job="myjob"}))`,
 		},
 	}
 
--- a/execution/function/absent.go
+++ b/execution/function/absent.go
@@ -1,0 +1,95 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package function
+
+import (
+	"context"
+	"sync"
+
+	"github.com/prometheus/prometheus/model/labels"
+
+	"github.com/thanos-io/promql-engine/execution/model"
+	"github.com/thanos-io/promql-engine/parser"
+)
+
+type absentOperator struct {
+	once     sync.Once
+	funcExpr *parser.Call
+	series   []labels.Labels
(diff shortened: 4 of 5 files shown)


execution/function/operator.go
```diff
@@ -33,6 +33,96 @@ type functionOperator struct {
...
+	funcExpr    *parser.Call
+	series      []labels.Labels
+	once        sync.Once
+
+	next       model.VectorOperator
+	vectorPool *model.VectorPool
+}
+
+func (o *absentOperator) Explain() (me string, next []model.VectorOperator) {
+	return fmt.Sprintf("[*absentOperator] %v()", o.funcExpr.Func.Name), []model.VectorOperator{}
```
yeya24, line 51: We can just use %s

execution/function/operator.go
```diff
@@ -33,6 +33,96 @@ type functionOperator struct {
...
+		var lm []*labels.Matcher
+		switch n := o.funcExpr.Args[0].(type) {
+		case *parser.VectorSelector:
+			lm = n.LabelMatchers
+		case *parser.MatrixSelector:
+			lm = n.VectorSelector.(*parser.VectorSelector).LabelMatchers
+		default:
+			o.series = []labels.Labels{labels.EmptyLabels()}
+			return
+		}
```
yeya24, line 72: The argument itself is the operator next IIUC as it can be nested expressions.

execution/function/operator.go
MichaHoffmann: :heavy_check_mark:

execution/function/operator.go
MichaHoffmann: Thats a good comment! I tested it out briefly but that would fail `absent(absent(exists{foo="bar"}))`; in this case prometheus returns `{} 1` but if i use nested labels it would be `{foo="bar"} 1` ; 

```
                    	exp: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:50000, V:1, H:(*histogram.FloatHistogram)(nil)}, Metric:labels.Labels{}}}, Warnings:storage.Warnings(nil)}

                    	got: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{Point:promql.Point{T:50000, V:1, H:(*histogram.FloatHistogram)(nil)}, Metric:labels.Labels{labels.Label{Name:"job", Value:"myjob"}}}}, Warnings:storage.Warnings(nil)}
```
let me actually push that test.

MichaHoffmann: Or did i misunderstand something? This is relatively new to me

yeya24: I haven't tested this myself but in theory, the inner `absent(exists{foo="bar"})` will give empty series if the series exists. So the outer `absent` will return `{}`. Probably there is edge case in the new engine implementation?

fpetkovski (commented):
Thanks for giving this a stab 👍  

I wonder if we can reuse or expand the existing functions to handle this case. Are some parameters missing from `FunctionArgs` to help us make the decision in the existing function operator?

MichaHoffmann:
> Thanks for giving this a stab +1
> 
> I wonder if we can reuse or expand the existing functions to handle this case. Are some parameters missing from `FunctionArgs` to help us make the decision in the existing function operator?

Current functions are operating on the points of the samples directly whereas `absent` would need special handling in the `if len(vectors)  == 0` clause. Its a bit more similar to `noArgsFunctionOperator` since it also constructs its result and doesnt modify the output of `Next` from its arguments with some special handling of labels. It feels a bit cleaner to have some dedicated operator for this to me; wdyt?

execution/function/operator.go
MichaHoffmann: Ah i forgot to say: I baseically reused upstream [promql implementation](https://github.com/prometheus/prometheus/blob/666f61a4d5d57e2a4ea2adbd27a3dc25c599e132/promql/functions.go#L1385) for the labels logic. If the expression is not a selector it will use empty labels every time. So we should ignore `next` here.

MichaHoffmann:
> > Thanks for giving this a stab +1
> > I wonder if we can reuse or expand the existing functions to handle this case. Are some parameters missing from `FunctionArgs` to help us make the decision in the existing function operator?
> 
> Current functions are operating on the points of the samples directly whereas `absent` would need special handling in the `if len(vectors) == 0` clause. Its a bit more similar to `noArgsFunctionOperator` since it also constructs its result and doesnt modify the output of `Next` from its arguments with some special handling of labels. It feels a bit cleaner to have some dedicated operator for this to me; wdyt?

Thinking about it, this also implements `absent_over_time` right?

execution/function/operator.go
```diff
@@ -33,6 +33,96 @@ type functionOperator struct {
...
+	}
+
+	vectors, err := o.next.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(vectors) > 0 {
+		return nil, nil
+	}
```
yeya24, line 113: We probably can optimize a little bit for `o.next.Next(ctx)` call as we don't need to retrieve the whole vectors back, just need to know whether `len(vectors) > 0` or not. But that's not something blocking this pr.

yeya24 (approved):
Nice. I love how this is implemented. The regression on absent not existent is worth investigating in the future.

yeya24:
> Its a bit more similar to noArgsFunctionOperator since it also constructs its result and doesnt modify the output of Next from its arguments with some special handling of labels.

I feel the same issue when I was working on other functions that requires to return different labels. Probably we can have a single operator and make `series` implementation configurable?

execution/function/operator.go
```diff
@@ -33,6 +33,96 @@ type functionOperator struct {
...
+
+	vectors, err := o.next.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(vectors) > 0 {
+		return nil, nil
```
fpetkovski, line 112: This in theory should cause the entire execution to stop. So if a series is present and then goes absent, we will never emit a `1` value. Is this how Prometheus works too?

MichaHoffmann: Ah really good point, 
```
		{
			name: "absent",
			load: `load 30s
				existent{job="myjob"} 1 1 1`,
			query: `absent(existent{job="myjob"})`,
		},
```
in `TestQueriesAgainstOldEngine` is failing exactely as you describe 

```
                    Diff:
                    --- Expected
                    +++ Actual
                    @@ -1,6 +1,2 @@
                    -(*promql.Result)({job="myjob"} =>
                    -1 @[150000]
                    -1 @[180000]
                    -1 @[210000]
                    -1 @[240000])
                    +(*promql.Result)()
```

fpetkovski: I assume the same would apply for the individual batches, we probably need to check on a per-step basis whether the metric is present or not.

mhoffm-aiven: Yeah i think we need to step through the vector at timestamps and if we have no samples we need to add a sample for absent; if we hvae samples we need to drop them. Or something like that

execution/function/operator.go
mhoffm-aiven: My mistake, we probably need whole vector, i added test that fails if we do a range query and some data is there.

execution/function/operator.go
MichaHoffmann: Ok i fixed the test in the latest version!

execution/function/operator.go
MichaHoffmann: weirdly enough with latest changes performance has improved drastically

```
RangeQuery/absent_and_exists-8                              132.40m ± ∞ ¹   14.03m ± ∞ ¹  -89.40% (p=0.008 n=5)
RangeQuery/absent_and_doesnt_exist-8                         184.4µ ± ∞ ¹   268.6µ ± ∞ ¹  +45.67% (p=0.008 n=5)
```

fpetkovski:
Let me take another look at this in the next few days. This is a commonly used function and I want to make sure we don't introduce some subtle bug.

fpetkovski (approved):
Nice work 👍

fpetkovski:
Let's rebase this and we can merge it

MichaHoffmann:
rebased and fixed some import errors ( thanos-community => thanos-io )

fpetkovski (approved):
🚀

PR #185, opened 2023-02-23
Fixing index out of range error in scaler operations

sharadgaur:
Seeing following error while running example query `time()-10`

image

Also noticed time() function only return 10 data points 
image

### DIFF
--- a/execution/function/operator.go
+++ b/execution/function/operator.go
@@ -63,7 +63,7 @@ func (o *noArgFunctionOperator) Next(_ context.Context) ([]model.StepVector, err
 		return nil, nil
 	}
 	ret := o.vectorPool.GetVectorBatch()
-	for i := 0; i < o.stepsBatch && o.currentStep <= o.maxt; i++ {
+	for i := 0; o.currentStep <= o.maxt; i++ {
 		sv := o.vectorPool.GetStepVector(o.currentStep)
 		result := o.call(FunctionArgs{
 			StepTime: o.currentStep,
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -849,6 +849,16 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			query: "time()",
 		},
 		{
+			name:  "binary operation on time function",
+			load:  "",
+			query: "time() - 20",
+		},
+		{
+			name:  "reverse binary operaton on time function",
+			load:  "",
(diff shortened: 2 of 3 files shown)


execution/function/operator.go
```diff
@@ -63,7 +63,7 @@ func (o *noArgFunctionOperator) Next(_ context.Context) ([]model.StepVector, err
 		return nil, nil
 	}
 	ret := o.vectorPool.GetVectorBatch()
-	for i := 0; i < o.stepsBatch && o.currentStep <= o.maxt; i++ {
```
sharadgaur, line 66: Why do we need this `i < o.stepsBatch`?

fpetkovski:
This doesn't look like the right fix. If you run just `time()` over a 30m range you'll notice gaps in the result, which is probably causing the panic.

MichaHoffmann:
> This doesn't look like the right fix. If you run just `time()` over a 30m range you'll notice gaps in the result, which is probably causing the panic.

I added the failing test and printed out `in` and `scalarIn`: 

```
{0 [] [0] [] []} {0 [0] [20] [] []}
{30000 [] [30] [] []} {30000 [0] [20] [] []}
{60000 [] [60] [] []} {60000 [0] [20] [] []}
{90000 [] [90] [] []} {90000 [0] [20] [] []}
{120000 [] [120] [] []} {120000 [0] [20] [] []}
{150000 [] [150] [] []} {150000 [0] [20] [] []}
{180000 [] [180] [] []} {180000 [0] [20] [] []}
{210000 [] [210] [] []} {210000 [0] [20] [] []}
{240000 [] [240] [] []} {240000 [0] [20] [] []}
```
it seems that every timestamp has a match on both sides

MichaHoffmann:
```
		switch funcExpr.Func.Name {
		case "pi", "time", "scalar":
		default:
			// Other functions require non-nil labels.
			op.series = []labels.Labels{{}}
			op.sampleIDs = []uint64{0}
		}
```
seems by design that "time()" has empty sampleIDs ? that feels weird.

fpetkovski:
Yeah that's one of the reasons, but we had other subtle issues which were not caught due to the short default range we use in tests. https://github.com/thanos-community/promql-engine/pull/186 should fix this.

sharadgaur:
Thank you guys for quick response. I going to close this as fix is already merged.

PR #186, opened 2023-02-24
Fix scalar and noArg functions

fpetkovski:
Looking at PR #185, I noticed that the functions which produce nil labels are not properly handled by the engine. They are handled as an exception in the engine and unwinded separately. This code has a bug where it only takes the last batch that comes out of the operator.

The reason why this was not caught by tests is because our default query range is very short and is covered by a single batch of 10 steps. This PR increases the default range to help uncover the issues. It also fixes the  tests that fail due to the increased range.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -363,24 +363,8 @@ loop:
 
 			// Case where Series call might return nil, but samples are present.
 			// For example scalar(http_request_total) where http_request_total has multiple values.
-			if len(resultSeries) == 0 && len(r) != 0 {
-				numSeries := 0
-				for i := range r {
-					numSeries += len(r[i].Samples)
-				}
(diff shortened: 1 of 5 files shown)


mhoffm-aiven (approved):
LGTM

fpetkovski:
@mhoffm-aiven you might want to bump the range in your PR for `absent` to make sure similar issues are not present there.

mhoffm-aiven:
> @mhoffm-aiven you might want to bump the range in your PR for `absent` to make sure similar issues are not present there.

I thought I'll rebase on main once this is merged and get it for free then. That should work right?

fpetkovski:
Yes, that should work too.

saswatamcode (approved):
Nice catch! Thanks for fixing!

PR #188, opened 2023-02-24
engine,execution: add sort functions

MichaHoffmann:
partially addresses #138

### DIFF
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -203,6 +203,14 @@ func BenchmarkRangeQuery(b *testing.B) {
 			name:  "histogram_quantile",
 			query: `histogram_quantile(0.9, http_response_seconds_bucket)`,
 		},
+		{
+			name:  "sort",
+			query: `sort(http_response_seconds_bucket)`,
+		},
+		{
+			name:  "sort_desc",
+			query: `sort_desc(http_response_seconds_bucket)`,
+		},
 	}
 
 	for _, tc := range cases {
@@ -294,6 +302,14 @@ func BenchmarkOldEngineInstant(b *testing.B) {
 		{
 			name:  "vector and scalar comparison",
 			query: `http_requests_total > 10`,
+		},
+		{
+			name:  "sort",
+			query: `sort(http_requests_total)`,
+		},
+		{
+			name:  "sort_desc",
+			query: `sort_desc(http_requests_total)`,
 		},
 	}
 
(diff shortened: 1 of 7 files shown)


MichaHoffmann:
Method is a bit dirty but we cannot sort in the operator. Maybe `resultSort` should be refactored slightly?

engine/bench_test.go
```diff
@@ -203,6 +203,14 @@ func BenchmarkRangeQuery(b *testing.B) {
 			name:  "histogram_quantile",
 			query: `histogram_quantile(0.9, http_response_seconds_bucket)`,
 		},
+		{
+			name:  "sort",
+			query: `sort(http_response_seconds_bucket)`,
+		},
+		{
+			name:  "sort_desc",
+			query: `sort(http_response_seconds_bucket)`,
```
yeya24, line 212: Small nit. Should be `sort_desc`.

MichaHoffmann: :heavy_check_mark:

MichaHoffmann: Thanks!

MichaHoffmann: Though i wonder if that is even a great benchmark since semantics of `sort` for a range query are not clear to me right now

fpetkovski:
I love the approach of sorting the end-result only. In principle, sort shouldn't even be a function but some sort of a formatter that you apply on the end result. If you want to take this further, you can also implement an optimizer that strips all non-root `sort` and `sort_desc` calls since their impact will be a no-op.

MichaHoffmann:
> 

Yeah sounds good, can you give me a pointer how to start with the optimizer? I'm still finding my footing with the engine.

fpetkovski:
You can take a look at the ones we have here: https://github.com/thanos-community/promql-engine/tree/main/logicalplan.
This one is the simplest one to understand: https://github.com/thanos-community/promql-engine/blob/main/logicalplan/sort_matchers.go.

MichaHoffmann:
Ok, i gave it a go; it seems alright, only caveat is that the engine now expects the `TrimSort` optimizer to always be present so that no function fallback is performed.

engine/bench_test.go
yeya24: Yeah, I think it is not clear at all what's the expected behavior in this case.

fpetkovski (approved):
Great work, lgtm 👍

mhoffm-aiven:
> Great work, lgtm +1

thank you!

yeya24 (approved):

Issue #166, opened 2023-01-31
Aggregate without one label produces incorrect metric labels


yeya24:
```
		{
			name: "aggregate without",
			load: `load 30s
				       http_requests_total{pod="nginx-1"} 1+1.1x1
				       http_requests_total{pod="nginx-2"} 2+2.3x1`,
			start: time.Unix(0, 0),
			end:   time.Unix(60, 0),
			step:  30 * time.Second,
			query: "avg without (pod) (http_requests_total)",
		}
```

The test case below producess different results for new and old engine.

```
exp: &promql.Result{Err:error(nil), Value:promql.Matrix{promql.Series{Metric:labels.Labels{}, Points:[]promql.Point{promql.Point{T:0, V:1.5, H:(*histogram.FloatHistogram)(nil)}, promql.Point{T:30000, V:3.2, H:(*histogram.FloatHistogram)(nil)}, promql.Point{T:60000, V:3.2, H:(*histogram.FloatHistogram)(nil)}}}}, Warnings:storage.Warnings(nil)}
        
got: &promql.Result{Err:error(nil), Value:promql.Matrix{promql.Series{Metric:labels.Labels{labels.Label{Name:"__name__", Value:"http_requests_total"}}, Points:[]promql.Point{promql.Point{T:0, V:1.5, H:(*histogram.FloatHistogram)(nil)}, promql.Point{T:30000, V:3.2, H:(*histogram.FloatHistogram)(nil)}, promql.Point{T:60000, V:3.2, H:(*histogram.FloatHistogram)(nil)}}}}, Warnings:storage.Warnings(nil)}
```

The old engine expects to drop the label `__name__` while the new engine still keeps it.

yeya24:
Closed by #190

PR #190, opened 2023-02-27
engine,execution: fix aggregation issues

MichaHoffmann:
Addresses #166 

Found an issue with `stddev` and `stdvar` when aggregating over a label with cardinality 1; upstream promql zeros out the first value (https://github.com/prometheus/prometheus/blob/ae72c752a175733a1150c04d68ea2d485d0a37ad/promql/engine.go#L2424).

The changes to fuzzing were just from me playing around; i can drop them from the PR if needed.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -100,6 +100,30 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 		step  time.Duration
 	}{
 		{
+			name: "stddev with NaN 1",
+			load: `load 30s
+				       http_requests_total{pod="nginx-1", route="/"} NaN
+				       http_requests_total{pod="nginx-2", route="/"} 1`,
+			query: "stddev by (route) (http_requests_total)",
+		},
+		{
+			name: "stddev with NaN 2",
+			load: `load 30s
+				       http_requests_total{pod="nginx-1", route="/"} NaN
+				       http_requests_total{pod="nginx-2", route="/"} 1`,
+			query: "stddev by (pod) (http_requests_total)",
+		},
+		{
+			name: "aggregate without",
+			load: `load 30s
+				       http_requests_total{pod="nginx-1"} 1+1.1x1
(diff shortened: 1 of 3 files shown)


execution/aggregate/hashaggregate.go
```diff
@@ -225,7 +226,8 @@ func (a *aggregate) initializeScalarTables(ctx context.Context) ([]aggregateTabl
 
 	series = make([]labels.Labels, len(outputCache))
 	for i := 0; i < len(outputCache); i++ {
-		series[i] = outputCache[i].Metric
+		lbls, _ := function.DropMetricName(outputCache[i].Metric)
```
fpetkovski, line 229: I wonder why this wasn't a problem before.

MichaHoffmann: That is a really good question. Ill investigate a bit.

MichaHoffmann: With `by` we aggregate away the `__name__` label i guess

MichaHoffmann: Yeah i think its precisely that: https://github.com/prometheus/prometheus/blob/5ec1b4baaf01dae0b22567db53886c83b74e0e76/promql/engine.go#L2389

MichaHoffmann: `without` implicitely drops the MetricName label

fpetkovski (approved):
Thanks, lgtm!

MichaHoffmann:
> Thanks, lgtm!

Should i keep the fuzzing change? I wonder if we can use the comparer there for other assertions too. I think it handles floats correctly

fpetkovski:
The fuzzing change makes sense to me, we just might want to break it down in multiple Fuzz functions. @GiedriusS wrote the fuzzing so it would be nice to hear his feedback too.

GiedriusS (approved):
LGTM, I think we can grow the fuzzing naturally for a bit until we will want to make something more sophisticated instead of the function accepting simple parameters

PR #191, opened 2023-02-28
Scope distributed queries to the min-time for each engine

fpetkovski:
The current algorithm for distributing queries only takes into account
the max-time advertised by each engine. When a remote engine is deployed
in front of Prometheus or Receivers, those stores can apply retention
periods at different times, leading to remote aggregations getting calculated
on a subset of the data. Since samples from newer engines are prioritized
over samples from older engines, these incomplete aggregates will end up in the end-result.

To work around the issue, this commit introduces a `MinTime` method to the
remote engine interface. The `MinTime` is used to scope remote queries at the lower
end of the range so that partial data is never queried and aggregated. 
The correct value for MinTime should be determined by the remote engine itself. 
In the case of Thanos, we can use the highest min-time from all TSDB backed by a given store,
and bubble it up to the querier responsible for querying those TSDBs.

Fixes https://github.com/thanos-community/promql-engine/issues/187


fpetkovski:
@GiedriusS any thoughts on this one?

logicalplan/distribute.go
```diff
@@ -148,20 +152,52 @@ func newRemoteAggregation(rootAggregation *parser.AggregateExpr, engines []api.R
 	return &remoteAggregation
 }
 
-func (m DistributedExecutionOptimizer) makeSubQueries(current *parser.Expr, engines []api.RemoteEngine) Deduplicate {
-	remoteQueries := make(RemoteExecutions, len(engines))
-	for i := 0; i < len(engines); i++ {
-		remoteQueries[i] = RemoteExecution{
-			Engine: engines[i],
-			Query:  (*current).String(),
+func (m DistributedExecutionOptimizer) makeSubQueries(expr *parser.Expr, engines []api.RemoteEngine, opts *Opts) Deduplicate {
```
saswatamcode, line 155: Let's add a detailed comment for this method, on how it helps create the subquery?

logicalplan/plan.go
```diff
@@ -20,34 +20,43 @@ var DefaultOptimizers = []Optimizer{
 	MergeSelectsOptimizer{},
 }
 
+type Opts struct {
```
saswatamcode, line 23: Can we include `expr` here too?

saswatamcode (approved):
Thanks for awesome work! LGTM as far as I understand this.

Some small suggestions/questions.

logicalplan/distribute.go
fpetkovski: Good point, comment should be there now. 

I also renamed the function to better indicate what it's doing.

logicalplan/plan.go
fpetkovski: Since each optimizer takes the `expr` as the first argument, do we need to add it to the query options as well?

saswatamcode (approved):

GiedriusS (approved):
LGTM, I think just the README could be improved a bit in the same PR :smile:

execution/execution.go
```diff
@@ -255,13 +255,19 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
...
+		qry, err := e.Engine.NewRangeQuery(&promql.QueryOpts{}, e.Query, e.QueryRangeStart, opts.End, opts.Step)
 		if err != nil {
 			return nil, err
 		}
 
-		return exchange.NewConcurrent(remote.NewExecution(qry, model.NewVectorPool(stepsBatch), opts), 2), nil
-
+		// The selector uses the original query time to make sure that steps from different
+		// operators have the same timestamps.
+		// We need to set the lookback for the selector to 0 since the remote query already applies one lookback.
```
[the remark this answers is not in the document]
GiedriusS: Good point only that it looks like Prometheus itself doesn't support setting lookback delta via individual requests https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go#L461-L465 :disappointed: but perhaps we could still push lookback delta through https://github.com/thanos-community/promql-engine/pull/191/files#diff-ac999eb9f6193331831ad85e398328d9b516bcfa685c5cc9282ca8761fe60d51R259 and use it if possible?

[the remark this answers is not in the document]
fpetkovski: Sounds good, the lookback should be forwarded in the query now.

fpetkovski:
Thanks, I added some more information in the README. Let me know if you think we need to mention anything else.

README.md
```diff
@@ -79,6 +77,33 @@ The current implementation uses goroutines very liberally which means the query
 
 The current implementation creates a physical plan directly from the PromQL abstract syntax tree. Plan optimizations not yet implemented and would require having a logical plan as an intermediary step.
 
+## Distributed execution mode
+
+The supports a distributed mode where aggregations can be delegated to multiple remote engines, each responsible for an independent dataset. This mode is currently implemented through an optimizer which rewrites a query as a combination of multiple remote and one local aggregation. For example, when two remote engines are available, a query like:
```
MichaHoffmann, line 82: Randomly noticed, but it probably should read `The engine supports`.

fpetkovski: Nice catch, thanks.

README.md
```diff
@@ -79,6 +77,33 @@ The current implementation uses goroutines very liberally which means the query
...
+  coalesce(
+    sum(rate(http_request_total[4m])) # remote engine 1
+    sum(rate(http_request_total[4m])) # remote engine 2
+  )
+)
+```
+
+The inner aggregations are forwarded to remote engines and the global result is completed in memory.
+
+An engine using the distributed mode can be created through the `NewDistributedEngine` function. The user is expected to pass an implementation of `RemoteEndpoints` with have a single `Engines()` method. When invoked, `Engines()` should return all remote engines that can be used for a single query. The `Engines()` method is called separately for each individual query which allows the `RemoteEndpoints` implementation to do continuous service discovery and inject engines as they become available.
```
saswatamcode, line 101:
```suggestion
An engine using the distributed mode can be created through the `NewDistributedEngine` function. The user is expected to pass an implementation of `RemoteEndpoints` which has a single `Engines()` method. When invoked, `Engines()` should return all remote engines that can be used for a single query. The `Engines()` method is called separately for each individual query which allows the `RemoteEndpoints` implementation to do continuous service discovery and inject engines as they become available.
```

README.md
```diff
@@ -79,6 +77,33 @@ The current implementation uses goroutines very liberally which means the query
...
+
+The interfaces used for remote execution can be found at [link](https://github.com/thanos-community/promql-engine/blob/main/api/remote.go). Note that the `RemoteEngine` interface has a `NewRangeQuery` method, similar to the one in the Prometheus [v1.QueryEngine](https://pkg.go.dev/github.com/prometheus/prometheus@v0.42.0/web/api/v1#QueryEngine) interface. It is up to the user of the library to implement this method as they see fit. An example implementation could be to forward the query to an HTTP `/api/v1/query_range` endpoint of a Prometheus instance. In Thanos, this method is implemented as a gRPC call to a Thanos Querier.
```
saswatamcode, line 103: Maybe using goDoc for all code references would be better?
```suggestion
The interfaces used for remote execution can be found in [api](https://pkg.go.dev/github.com/thanos-community/promql-engine/api) package. Note that the `RemoteEngine` interface has a `NewRangeQuery` method, similar to the one in the Prometheus [v1.QueryEngine](https://pkg.go.dev/github.com/prometheus/prometheus@v0.42.0/web/api/v1#QueryEngine) interface. It is up to the user of the library to implement this method as they see fit. An example implementation could be to forward the query to an HTTP `/api/v1/query_range` endpoint of a Prometheus instance. In Thanos, this method is implemented as a gRPC call to a Thanos Querier.
```

saswatamcode (approved):
LGTM!

Some small readme suggestions!

README.md
fpetkovski: Great point.

saswatamcode:
Seems like test timed out. Re-running.

README.md
```diff
@@ -79,6 +77,33 @@ The current implementation uses goroutines very liberally which means the query
...
+
+The interfaces used for remote execution can be found in [api](https://pkg.go.dev/github.com/thanos-community/promql-engine/api) package. Note that the `RemoteEngine` interface has a `NewRangeQuery` method, similar to the one in the Prometheus [v1.QueryEngine](https://pkg.go.dev/github.com/prometheus/prometheus@v0.42.0/web/api/v1#QueryEngine) interface. It is up to the user of the library to implement this method as they see fit. An example implementation could be to forward the query to an HTTP `/api/v1/query_range` endpoint of a Prometheus instance. In Thanos, this method is implemented as a gRPC call to a Thanos Querier.
+
+For more details on the overall design, please refer to the [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals-accepted/202301-distributed-query-execution.md) in the Thanos project
```
GiedriusS, line 105:
```suggestion
For more details on the overall design, please refer to the [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals-accepted/202301-distributed-query-execution.md) in the Thanos project.
```

fpetkovski:
I will proceed with merging this and we can keep iterating.

PR #193, opened 2023-02-28
execution: add round function

MichaHoffmann:
Partially addresses #138

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1820,6 +1820,28 @@ func TestInstantQuery(t *testing.T) {
 			query:     "increase(http_requests_total[1m] offset 1m)",
 		},
 		{
+			name: "round",
+			load: `load 1s
+				       http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
+				       http_requests_total{pod="nginx-2", series="2"} 2+2.3x50
+				       http_requests_total{pod="nginx-4", series="3"} 5+2.4x50
+				       http_requests_total{pod="nginx-5", series="1"} 8.4+2.3x50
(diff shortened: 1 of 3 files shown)


engine/engine_test.go
```diff
@@ -1819,6 +1819,28 @@ func TestInstantQuery(t *testing.T) {
 			queryTime: time.Unix(160, 0),
 			query:     "increase(http_requests_total[1m] offset 1m)",
 		},
+		{
+			name: "round with argument",
```
saswatamcode, line 1823: I think the test case name should be switched here?
```suggestion
			name: "round without argument",
```

engine/engine_test.go
```diff
@@ -1819,6 +1819,28 @@ func TestInstantQuery(t *testing.T) {
...
+				       http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
+				       http_requests_total{pod="nginx-2", series="2"} 2+2.3x50
+				       http_requests_total{pod="nginx-4", series="3"} 5+2.4x50
+				       http_requests_total{pod="nginx-5", series="1"} 8.4+2.3x50
+				       http_requests_total{pod="nginx-6", series="2"} 2.3+2.3x50`,
+			queryTime: time.Unix(0, 0),
+			query:     "round(http_requests_total)",
+		},
+		{
+			name: "round without argument",
```
saswatamcode, line 1834:
```suggestion
			name: "round with argument",
```

saswatamcode (approved):
LGTM, mod some suggestions! 🙂

engine/engine_test.go
MichaHoffmann: thanks, addressed it!

fpetkovski (approved):
Thanks

PR #196, opened 2023-03-02
Fix distributed exec for no arg functions

fpetkovski:
Functions without arguments can be calculated in memory and do not need to be distributed. This commit changes the distributing optimizer to always run such functions using the local engine.


saswatamcode (approved):
Thanks! LGTM!

PR #197, opened 2023-03-02
Add support for float histograms

fpetkovski:
Float histograms are currently not handled by the engine. They are less frequently returned from storage, but they will be returned by a remote engine.

This commit adds support for querying this histogram type and extends existing tests to include queries against them.


saswatamcode (approved):
LGTM as far as I understand!

PR #199, opened 2023-03-03
Add benchmarks for native histograms

fpetkovski:
This commit adds benchmarks for queries on native histograms. The benchmark results should be added to the same output as the ones for queries on floats.


execution/aggregate/vector_table.go
```diff
@@ -129,10 +129,10 @@ func newVectorAccumulator(expr parser.ItemType) (vectorAccumulator, error) {
 
 func histogramSum(histograms []*histogram.FloatHistogram) *histogram.FloatHistogram {
 	if len(histograms) == 1 {
-		return histograms[0]
+		return histograms[0].Copy()
```
saswatamcode, line 132: Is there a reason for this change? Did this fix the test somehow? 🙂

fpetkovski: Good question, it did in fact fix the test after I changed the number of generated histograms.

When applying a lookback, the selector can return the same instance of histogram as in previous steps. So if we don't copy the histogram for the sum result, we will end up modifying it in place when adding new histograms to it. Then when we apply a lookback, the returned histogram will actually be the `sum` result.

This is where the lookback is done: https://github.com/thanos-community/promql-engine/blob/main/execution/scan/vector_selector.go#L189.

saswatamcode: Thanks! That makes sense

engine/engine_test.go
```diff
@@ -3399,7 +3402,12 @@ func testNativeHistograms(t *testing.T, cases []histogramTestCase, opts promql.E
 
 						sortByLabels(promResult)
 						sortByLabels(newResult)
-						testutil.Equals(t, promVector, newVector)
+						if hasNaNs(promResult) {
+							t.Log("Applying comparison with NaN equality.")
+							testutil.WithGoCmp(cmpopts.EquateNaNs()).Equals(t, promVector, newVector)
+						} else {
+							require.Equal(t, promVector, newVector)
```
saswatamcode, line 3409: Let's maybe keep single testing lib
```suggestion
							testutil.Equals(t, promVector, newVector)
```

engine/engine_test.go
```diff
@@ -3423,18 +3431,25 @@ func testNativeHistograms(t *testing.T, cases []histogramTestCase, opts promql.E
 							testutil.Assert(t, len(expected) == 0)
 						}
 						testutil.Equals(t, len(expected), len(actual))
-						testutil.Equals(t, expected, actual)
+						if hasNaNs(res) {
+							t.Log("Applying comparison with NaN equality.")
+							testutil.WithGoCmp(cmpopts.EquateNaNs()).Equals(t, expected, actual)
+						} else {
+							require.Equal(t, expected, actual)
```
saswatamcode, line 3438:
```suggestion
							testutil.Equals(t, expected, actual)
```

saswatamcode (approved):
LGTM! Thanks for adding this! Maybe we can add the results to README too?

Small suggestion.

fpetkovski:
Good point, updated 👍

saswatamcode (approved):
LGTM

PR #205, opened 2023-03-06
engine: add test for binary operators with bool modifiers

MichaHoffmann:
If i read https://prometheus.io/docs/prometheus/latest/querying/operators/ correctly we drop the name label for all arithmetic operations and for comparisons that have the bool modifier set.


fpetkovski (approved):
Thanks, lgtm.

PR #207, opened 2023-03-07
Fix topk aggregate

fpetkovski:
The topk aggregate appends one step vector per group. This causes vector steps to get misaligned and leads to panics in the engine.

This commit fixes the issue by appending results from all groups to the same step and adds a test to prevent regressions.

Fixes https://github.com/thanos-community/promql-engine/issues/178.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1434,6 +1434,17 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			step:  2 * time.Second,
 		},
 		{
+			name: "topk wrapped by another aggregate",
+			load: `load 30s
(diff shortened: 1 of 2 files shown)


MichaHoffmann:
Test crashes on main. LGTM

MichaHoffmann (approved):

MichaHoffmann:
Of note: i tried to run this as InstantQuery and it didnt crash but produce wrong results on main:
```
                    Diff:
                    --- Expected
                    +++ Actual
                    @@ -1,2 +1,2 @@
                    -(*promql.Result)({} => 10.7 @[50000])
                    +(*promql.Result)({} => 4.3 @[50000])

```

This is fixed on this PR too.

GiedriusS (approved):

PR #210, opened 2023-03-07
execution: fix NaN handling in max aggregation

MichaHoffmann:
upstream handles NaN slightly different in max aggregations: https://github.com/prometheus/prometheus/blob/ae72c752a175733a1150c04d68ea2d485d0a37ad/promql/engine.go#L2497

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1976,6 +1976,20 @@ func TestInstantQuery(t *testing.T) {
 			query: "sum by (pod) (http_requests_total)",
 		},
 		{
+			name: "sum with NaN",
+			load: `load 30s
+						http_requests_total{pod="nginx-1"} 1+1x4
+						http_requests_total{pod="nginx-2"} NaN`,
+			query: "sum(http_requests_total)",
+		},
+		{
+			name: "sum by with NaN",
+			load: `load 30s
+						http_requests_total{pod="nginx-1"} 1+1x4
+						http_requests_total{pod="nginx-2"} NaN`,
+			query: "sum by (pod) (http_requests_total)",
+		},
+		{
 			name: "count",
 			load: `load 30s
 						http_requests_total{pod="nginx-1"} 1+1x15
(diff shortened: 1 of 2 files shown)


engine/engine_test.go
```diff
@@ -2061,6 +2061,16 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-2"} 1`,
 			query: "max(http_requests_total) by (pod)",
 		},
+		{
+			name: "max by with NaN",
```
fpetkovski, line 2065: Do we need to fix this for other aggregations too, like `sum`, `min` etc?

Also can we have a test for aggregations without grouping?

MichaHoffmann: I added couple tests; the other cases ( no grouping, NaN in sum and min ) seem to be ok!

engine/engine_test.go
```diff
@@ -1975,6 +1975,13 @@ func TestInstantQuery(t *testing.T) {
 						http_requests_total{pod="nginx-2"} 1+2x4`,
 			query: "sum by (pod) (http_requests_total)",
 		},
+		{
+			name: "sum with NaN",
+			load: `load 30s
+						http_requests_total{pod="nginx-1"} 1+1x4
+						http_requests_total{pod="nginx-2"} NaN`,
+			query: "sum(http_requests_total)",
```
fpetkovski, line 1983: Can we also add `sum by` with NaN since there's a different code path for grouping aggregates.

MichaHoffmann: added a test for `sum by` with NaN

fpetkovski (approved):
Thanks!

execution/aggregate/scalar_table.go
```diff
@@ -185,8 +185,8 @@ func makeAccumulatorFunc(expr parser.ItemType) (newAccumulatorFunc, error) {
 				AddFunc: func(v float64, _ *histogram.FloatHistogram) {
 					if !hasValue {
 						value = v
-					} else {
-						value = math.Max(value, v)
+					} else if value < v || math.IsNaN(value) {
```
GiedriusS, line 188: Shall we do the same for `min`?

yeya24: +1. I feel we need to apply the same for min

mhoffm-aiven:
This is superseded by #237 which fixes more issues

PR #212, opened 2023-03-08
logicalplan: propagate matchers should not be applied to atan2

MichaHoffmann:


logicalplan/propagate_selectors.go
```diff
@@ -21,8 +21,8 @@ func (m PropagateMatchersOptimizer) Optimize(expr parser.Expr) parser.Expr {
 			return
 		}
 
-		// The optimizer cannot be applied to comparison operations.
-		if binOp.Op.IsComparisonOperator() {
+		// The optimizer cannot be applied to comparison operations or 'atan2'.
+		if binOp.Op.IsComparisonOperator() || binOp.Op.String() == "atan2" {
```
fpetkovski, line 25: Interesting, what about other trig functions?

MichaHoffmann: I think atan2 is the only one that can be used as a binary expression!

fpetkovski (approved):
Thanks for the fix 👍

PR #213, opened 2023-03-09
Fix capitalisation in error message that differs from Prometheus' error message.

charleskorn:
Very minor fix for an issue that I noticed while comparing Prometheus' engine with this one.


yeya24 (approved):
Nice catch. Thanks!

PR #214, opened 2023-03-09
execution: verify parameters of topk function

### LINKED ISSUE #211: topk panics when param < 1
```
		{
			name: "bbb",
			load: `load 30s
http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
http_requests_total{pod="nginx-2", series="2"} 2+2.3x50
http_requests_total{pod="nginx-3", series="3"} 6+0.8x60
http_requests_total{pod="nginx-4", series="3"} 5+2.4x50
http_requests_total{pod="nginx-5", series="1"} 8.4+2.3x50
http_requests_total{pod="nginx-6", series="2"} 2.3+2.3x50
`,
			query: `topk without () (0.5, vector(time()))`,
		}
```

Using float value for the `k` param is actually valid in the old Prometheus engine. Same query won't hit panic in the old engine but panics in the new engine.

Old engine: https://prometheus.demo.do.prometheus.io/graph?g0.expr=topk%20without%20()%20(0.5%2C%20vector(time()))&g0.tab=1&g0.stacked=0&g0.range_input=1h

```
=== RUN   TestInstantQuery/disableOptimizers=true/bbb/disableFallback=false
panic: runtime error: index out of range [0] with length 0

goroutine 108 [running]:
github.com/thanos-community/promql-engine/execution/aggregate.(*kAggregate).aggregate(0x14000546a90, 0xc350, 0x14000077de8, 0x0, {0x1400032ce30, 0x1, 0x1?}, {0x140007d00e0, 0x1, 0x0?})
	/Users/benye/hub/gowork/promql-engine/execution/aggregate/khashaggregate.go:167 +0x414
github.com/thanos-community/promql-engine/execution/aggregate.(*kAggregate).Next(0x14000546a90, {0x1012b3ec8, 0x14000115f80})
	/Users/benye/hub/gowork/promql-engine/execution/aggregate/khashaggregate.go:114 +0x594
github.com/thanos-community/promql-engine/execution/exchange.(*concurrencyOperator).pull(0x14000801a10, {0x1012b3ec8, 0x14000115f80})
	/Users/benye/hub/gowork/promql-engine/execution/exchange/concurrent.go:80 +0xc0
created by github.com/thanos-community/promql-engine/execution/exchange.(*concurrencyOperator).Next.func1
	/Users/benye/hub/gowork/promql-engine/execution/exchange/concurrent.go:56 +0xa0
```

MichaHoffmann:
fixes #211

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ export GOCACHE=/tmp/cache
 test:
 	@echo ">> running unit tests (without cache)"
 	@rm -rf $(GOCACHE)
-	@go test -v -race -timeout=5m $(shell go list ./...);
+	@go test -v -race -timeout=10m $(shell go list ./...);
 
 .PHONY: deps
 deps: ## Ensures fresh go.mod and go.sum.
--- a/execution/aggregate/khashaggregate.go
+++ b/execution/aggregate/khashaggregate.go
@@ -82,25 +82,28 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 	if err != nil {
 		return nil, err
 	}
-	if in == nil {
-		return nil, nil
-	}
-
-	defer a.next.GetPool().PutVectors(in)
-
 	args, err := a.paramOp.Next(ctx)
 	if err != nil {
 		return nil, err
 	}
-	for i := range a.params {
-		a.params[i] = math.NaN()
-		if i < len(args) {
-			a.params[i] = args[i].Samples[0]
-			a.paramOp.GetPool().PutStepVector(args[i])
+
+	for i := range args {
+		a.params[i] = args[i].Samples[0]
+		a.paramOp.GetPool().PutStepVector(args[i])
+
+		val := a.params[i]
+		if val > math.MaxInt64 || val < math.MinInt64 || math.IsNaN(val) {
+			return nil, errors.Newf("scalar value %v overflows int64", val)
+		}
+		if int(val) == 0 {
+			return nil, nil
 		}
 	}
 	a.paramOp.GetPool().PutVectors(args)
 
+	if in == nil {
+		return nil, nil
+	}
 	if len(args) < len(in) {
 		return nil, errors.New("scalar argument not found")
 	}
@@ -115,6 +118,7 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 		a.aggregate(vector.T, &result, int(a.params[i]), vector.SampleIDs, vector.Samples)
 		a.next.GetPool().PutStepVector(vector)
 	}
+	a.next.GetPool().PutVectors(in)
 
 	return result, nil
 }
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1462,6 +1462,74 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			step:  2 * time.Second,
 		},
 		{
+			name: "topk with float64 parameter",
+			load: `load 30s
+				http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
+				http_requests_total{pod="nginx-2", series="1"} 2+2.3x50
+				http_requests_total{pod="nginx-4", series="2"} 5+2.4x50
+				http_requests_total{pod="nginx-5", series="2"} 8.4+2.3x50
+				http_requests_total{pod="nginx-6", series="2"} 2.3+2.3x50`,
+			query: "topk(3.5, http_requests_total)",
+			start: time.Unix(0, 0),
+			end:   time.Unix(3000, 0),
+			step:  2 * time.Second,
+		},
+		{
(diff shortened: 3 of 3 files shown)


execution/aggregate/khashaggregate.go
```diff
@@ -91,12 +91,16 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+	for i := range args {
+		val := args[i].Samples[0]
+		if val > math.MaxInt64 || val < math.MinInt64 {
+			return nil, errors.Newf("scalar value %v overflows int64", val)
 		}
+		if int(val) == 0 {
+			return nil, nil
+		}
+		a.params[i] = val
+		a.paramOp.GetPool().PutStepVector(args[i])
```
MichaHoffmann, line 103: Should we do this on the error paths too?

execution/aggregate/khashaggregate.go
```diff
@@ -91,12 +91,16 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 	if err != nil {
 		return nil, err
 	}
-	for i := range a.params {
-		a.params[i] = math.NaN()
-		if i < len(args) {
-			a.params[i] = args[i].Samples[0]
-			a.paramOp.GetPool().PutStepVector(args[i])
+	for i := range args {
```
MichaHoffmann, line 94: I wonder: why do we need to iterate args for all steps, shouldnt it be constant?

yeya24: IIRC, the param could change if we use something like `topk(scalar(some vector), xxx)`. I don't remember the details so I might be wrong.

execution/aggregate/khashaggregate.go
fpetkovski: Probably a good idea to always do it.

MichaHoffmann: refactored slightly so that we always return args[i] back to the pool

execution/aggregate/khashaggregate.go
```diff
@@ -91,19 +91,25 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
 	}
-	for i := range a.params {
-		a.params[i] = math.NaN()
-		if i < len(args) {
-			a.params[i] = args[i].Samples[0]
-			a.paramOp.GetPool().PutStepVector(args[i])
-		}
-	}
-	a.paramOp.GetPool().PutVectors(args)
+	defer a.paramOp.GetPool().PutVectors(args)
```
fpetkovski, line 94: Using `defer` to release vectors has lead to some subtle bugs in the past. Could we revert to the previous implementation where we do it explicitly after iterating through parameter vectors?

fpetkovski (commented):
Thanks, just one small suggestion from my side.

execution/aggregate/khashaggregate.go
MichaHoffmann: I thought it would be better like this since we could return early and in that case we would not return to the pool ( or explicitely `Put` on every return statement ). Out of curiosity what was the bug with defer?

MichaHoffmann: reverted to putting the vectors back after the loops

Makefile
```diff
@@ -35,7 +35,7 @@ export GOCACHE=/tmp/cache
 test:
 	@echo ">> running unit tests (without cache)"
 	@rm -rf $(GOCACHE)
-	@go test -v -race -timeout=5m $(shell go list ./...);
+	@go test -v -race -timeout=10m $(shell go list ./...);
```
MichaHoffmann, line 38: I took the liberty to increase this to 10m since the pipeline timed out a few times.

saswatamcode (approved):
Thanks!

execution/aggregate/khashaggregate.go
```diff
@@ -82,25 +82,28 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+		if val > math.MaxInt64 || val < math.MinInt64 || math.IsNaN(val) {
+			return nil, errors.Newf("scalar value %v overflows int64", val)
+		}
+		if int(val) == 0 {
+			return nil, nil
 		}
 	}
 	a.paramOp.GetPool().PutVectors(args)
 
+	if in == nil {
```
fpetkovski, line 104: Why do we need to move this check here?

MichaHoffmann: That is needed for this test to pass: 
```
		{
			name:  "topk with NaN and no matching series",
			query: "topk(NaN, not_there)",
			start: time.Unix(0, 0),
			end:   time.Unix(3000, 0),
			step:  2 * time.Second,
		},

```
upstream prometheus errors out even if there is no matching series to compute topk for, if we return early we dont get that error.

fpetkovski: Got it 👍

PR #215, opened 2023-03-10
Add promqlsmith fuzzer

yeya24:
This PR uses the new PromQLsmith query fuzzer to replace our old fuzzing test.

Fixes https://github.com/thanos-community/promql-engine/issues/209.

Note that the fuzzer already found some bugs at https://github.com/thanos-community/promql-engine/issues/200 but currently we don't have a way to tell the fuzzer to avoid generating such kind of queries. In this case, the CI will fail.

I am also wondering if we should have a dedicated workflow for running the fuzz test. And currently we don't have a way to collect queries bugs from our test run.

### DIFF
--- a/engine/enginefuzz_test.go
+++ b/engine/enginefuzz_test.go
@@ -7,19 +7,24 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"math/rand"
 	"testing"
 	"time"
 
+	"github.com/cortexproject/promqlsmith"
+	"github.com/efficientgo/core/errors"
 	"github.com/efficientgo/core/testutil"
 	"github.com/google/go-cmp/cmp"
+	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql"
-	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


engine/enginefuzz_test.go
```diff
@@ -32,108 +37,141 @@ func FuzzEngineQueryRangeMatrixFunctions(f *testing.F) {
...
+		ps := promqlsmith.New(rnd, seriesSet, psOpts...)
+
+		newEngine := engine.New(engine.Opts{EngineOpts: opts, DisableFallback: true})
+
+		var (
+			q1    promql.Query
+			query string
+		)
+		// Since we disabled fallback, keep trying until we find a query
+		// that can be natively execute by the engine.
```
fpetkovski, line 78:
```suggestion
		// that can be natively executed by the engine.
```

engine/new.txt
```diff
@@ -0,0 +1,18 @@
+goos: darwin
```
fpetkovski, line 1: Is this supposed to be in the commit?

fpetkovski (commented):
Thanks, this is a great addition.

engine/new.txt
yeya24: Ah, sorry I should exclude this.

yeya24:
Fixed @fpetkovski Please take another look.

fpetkovski (approved):
Thanks, lgtm

PR #216, opened 2023-03-10
Implement XFunctions (xRate, xDelta, xIncrease)

mklocke-shopify:
**Description**
This is a first pass at implementing xRate, xDelta and xIncrease in the `promql-engine`. 

A large part of this is inspired by the excellent work done [here](https://github.com/vinted/prometheus/commit/835de212f1022e4be7d18af5a0dd1f1c8589adce). Its been reworked and adapted to fit into the newer engine but the core principles are the same.

Additionally the following resources were also very helpful for inspiration and understanding:
- https://docs.google.com/document/d/1y2Mp041_2v0blnKnZk7keCnJZICeK2YWUQuXH_m4DVc/edit
- https://github.com/free/prometheus/commit/80132a575cba5c4c95b6a2651fc50326b8eafd00

**What is left to do?**
- [x] Add additional missing tests and fine tune other tests that may need adjustment.
- [x] Completeness of non-extrapolating rate function, this is intertwined with adding all of the tests that are required as it will help assess what is missing in the implementation(s).
- [x] Additional comments, cleanups and documentation.
- [x] Bucket histogram support.


engine/engine.go
```diff
@@ -54,6 +54,10 @@ type Opts struct {
 	// If nil, nothing will be printed.
 	// NOTE: Users will not check the errors, debug writing is best effort.
 	DebugWriter io.Writer
+
+	// ExtLookbackDelta specifies what time range to use to determine valid previous time seris for extended range functions.
```
fpetkovski, line 58:
```suggestion
	// ExtLookbackDelta specifies what time range to use to determine valid previous sample for extended range functions.
```

engine/enginefuzz_test.go
```diff
@@ -38,6 +38,11 @@ func FuzzEngineQueryRangeMatrixFunctions(f *testing.F) {
 				continue
 			}
 
+			// Skip xFunctions as the old engine does not support them
```
fpetkovski, line 41:
```suggestion
			// Skip xFunctions as the Prometheus engine does not support them
```

execution/execution.go
```diff
@@ -65,10 +65,10 @@ func New(expr parser.Expr, queryable storage.Queryable, mint, maxt time.Time, st
 		// TODO(fpetkovski): Adjust the step for sub-queries once they are supported.
 		Step: step.Milliseconds(),
 	}
-	return newOperator(expr, selectorPool, opts, hints)
+	return newOperator(expr, selectorPool, opts, hints, extLookbackDelta)
 }
 
-func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.Options, hints storage.SelectHints) (model.VectorOperator, error) {
+func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.Options, hints storage.SelectHints, extLookbackDelta time.Duration) (model.VectorOperator, error) {
```
fpetkovski, line 71: Maybe we can add extLookbackDelta to query.Options? This way we don't have to drag it around as another parameter.

execution/scan/matrix_selector.go
```diff
@@ -288,3 +309,112 @@ loop:
...
+// single time series, with points retrieved from an iterator.
+//
+// As an optimization, the matrix vector may already contain points of the same
+// time series from the evaluation of an earlier step (with lower mint and maxt
+// values). Any such points falling before mint are discarded; points that fall
+// into the [mint, maxt] range are retained; only points with later timestamps
+// are populated from the iterator.
+// TODO(fpetkovski): Add max samples limit.
+func selectExtPoints(it *storage.BufferedSeriesIterator, mint, maxt int64, out []promql.Point, functionName string, extLookbackDelta int64) ([]promql.Point, error) {
+	extRange := function.IsExtFunction(functionName)
```
fpetkovski, line 323: Can we assume we always want the extended range? The function is called `selectExtPoints` already.

fpetkovski (commented):
Thanks, this looks quite at this point.

engine/engine.go
```diff
@@ -128,6 +132,14 @@ func New(opts Opts) *compatibilityEngine {
...
 		level.Debug(opts.Logger).Log("msg", "lookback delta is zero, setting to default value", "value", 5*time.Minute)
 	}
+	if opts.ExtLookbackDelta == 0 {
+		opts.ExtLookbackDelta = 1 * 24 * time.Hour
+	}
+
+	// Set the parser functions for extended functions appropriately as they are not present in prometheus
+	parser.Functions["xdelta"] = parse.Functions["xdelta"]
+	parser.Functions["xincrease"] = parse.Functions["xincrease"]
+	parser.Functions["xrate"] = parse.Functions["xrate"]
```
fpetkovski, line 142: Maybe we want this in an `init` function so we don't have to do it for each instance of the engine.

GiedriusS: We were planning on making it possible to switch between engines using headers to improve the UX i.e. so that a whole restart wouldn't be needed to switch between engines and it would allow faster experimentation with engines. I believe this will interfere with how the Prometheus engine works? :cry: we will probably have to fork the parser :/

fpetkovski: What if we have a custom implementation of `ParseExpr` that injects the new functions, calls Prometheus `ParseExpr` and removes the functions once it is done? We might need to use a mutex but I don't think the impact should be that large in comparison to the cost of executing the query.

Maintaining a fork of the Prometheus parser sound scary 😀

GiedriusS: That sounds very hacky :smile: I mean we wouldn't even have to change anything in the parser, the only reason we'd have that fork is to avoid clashing between engines so it should be very easy to update. Perhaps we could just vendor it inside `vendor/` in this repo with some kind of different name and then use it in `go.mod`?

execution/execution.go
mklocke-shopify: Good call! Refactored here: https://github.com/thanos-community/promql-engine/pull/216/commits/7408ac13ffbced1d1173fc17c07ad47135e71c6b

Much nicer.

execution/scan/matrix_selector.go
mklocke-shopify: Removed `ext` relevant logic in the function here: https://github.com/thanos-community/promql-engine/pull/216/commits/63d82031eff1b14f8312b0613ca8f650826aacba

engine/engine.go
fpetkovski: Thinking about this again, I think vendoring might be the better option. I am concerned we could push ourselves into a corner if we fork Prometheus since we will import two modules with the same name `github.com/prometheus/prometheus`.  

I don't know if it's a valid concern, but maybe it's better to vendor the parser just to be on the safe side.

GiedriusS: Is there a possibility to vendor only one module? `go mod vendor` seems to suggest that it's all or nothing :sweat_smile:

saswatamcode: Something like what we do for Cortex in Thanos: https://github.com/thanos-io/thanos/pull/5504?

fpetkovski: I think the decision to vendor Cortex under an internal package was because of circular dependencies.

I'll need to check what happens if we vendor prometheus using `go mod`. My concern is that Thanos will require prometheus/prometheus from github.com, and promql-engine will require prometheus/prometheus from a local vendor folder. 

In the ideal case this will not cause Thanos to somehow magically use prometheus/prometheus from promql's vendored folder, but I would like to double check that first.

mklocke-shopify: We tried to vendor `prometheus/prometheus` within `promql-engine` but go module system starts running into issues when you reference both `promql-engine` and a separate `prometheus/prometheus` within a third project.

So for example, if we import `thanos-community/promql-engine` and a separate `prometheus/prometheus` within a third project it will just optimise the vendored Prometheus away.

We tried to force it to use separate versions of Prometheus to see if it would avoid the optimisation but it simply pruned the whole `thanos-community/promql-engine` module once it detected a version mismatch of prometheus between the two.

mklocke-shopify: And another update on this, I tried to simply copy the parser code using a make target into the engine but we run into some more issues with compatibility.

For example the Result expects the prom parser to be defined there: https://github.com/prometheus/prometheus/blob/v0.42.0/promql/value.go#L296

Screenshot 2023-03-21 at 11 22 40

GiedriusS: Mhm, so it seems like a full fork of the repository is the only solution? Thanks for trying out all of these options 🙇

mklocke-shopify: Or we just inject the functions as initially implemented 😞 but I realise that leads to some undesirable side effects.

fpetkovski: We can try the fork, but I suspect it will have the same problem as the vendoring solution. The fork's module name will be `prometheus/prometheus` so when we do a require from Thanos, the `go mod` command will have to converge a single module. My guess is that we won't be able to `require` both our fork and github.com/prometheus/prometheus from inside Thanos, even though the fork is a transitive dependency.

mklocke-shopify: Unfortunately I think you're correct, I tried forking [prometheus](https://github.com/mklocke-shopify/prometheus/commit/f2d98bc2f97bfb0c09bb40c499e7da850d6c9054) and then replaced in promql-engine:

`replace github.com/prometheus/prometheus v0.42.0 => github.com/mklocke-shopify/prometheus v0.42.1-0.20230321212756-f2d98bc2f97b`

When both `promql-engine` and `prometheus` are imported into a third go application it optimises the imports in such a manner that my fork in `promql-engine` was ignored and the functions were no longer included in the parser.

GiedriusS: Why not replace the whole import so the module's path will be `github.com/mklocke-shopify/prometheus` instead of `github.com/prometheus/prometheus`?

fpetkovski: Unfortunately this is currently not possible with go modules because the module name is defined in the `go.mod` [file](https://github.com/grafana/mimir/blob/main/go.mod#L245-L246). If we change the module name of the fork, we also have to modify all internal imports that point to `github.com/prometheus/prometheus` to point to `github.com/thanos-io/prometheus`.

Mimir also has a Prometheus fork and they point to it using a replace statement: https://github.com/grafana/mimir/blob/main/go.mod#L245-L246. 

Maybe we should address the root cause here is that the parser is not configurable and relies on the global `Functions` variable. I've raised a PR to address this: https://github.com/prometheus/prometheus/pull/12170.

mklocke-shopify:
A couple screenshots of this running locally:
Screenshot 2023-03-22 at 09 03 47

Screenshot 2023-03-22 at 09 04 03

engine/engine.go
fpetkovski: It seems that Prometheus might in fact be willing to take in a contribution for xrate/xincrease, but Julien is reluctant to change the parser to allow injecting additional functions: https://github.com/prometheus/prometheus/pull/12170#issuecomment-1479320537. 

Any suggestions on what else we can do here?

mklocke-shopify: If they are considering accepting xFunctions behind a feature flag some day (probably not soon?) can we accept this as is with injecting it into the parser? I realise this causes potential unwanted side-affects when importing both this and original prometheus engine.

GiedriusS: > If we change the module name of the fork, we also have to modify all internal imports which point to github.com/prometheus/prometheus to point to github.com/thanos-io/prometheus.

Why isn't this an option? 🤔 what are the blockers?

yeya24: What I think we could do is to have separate go module for https://github.com/prometheus/prometheus/promql/parser part. Or we extract that part of code to a new repo and we make changes.
Cortex uses both Prometheus and promql-engine, but we don't want to use Prometheus fork.

fpetkovski: > Why isn't this an option? 🤔 what are the blockers?

I would be concerned with the maintenance burden. Once we change all internal references in the fork, we won't be able to update it through the UI and we will need some automation to pull latest changes from Prometheus and also resolve conflicts with imports. It might be doable but I am concerned it can get quite error prone.

However, this could be a way to vendor in the Prometheus parser: https://github.com/thanos-community/promql-engine/pull/223.

engine/engine.go
```diff
@@ -155,6 +159,10 @@ func New(opts Opts) *compatibilityEngine {
 		opts.LookbackDelta = 5 * time.Minute
 		level.Debug(opts.Logger).Log("msg", "lookback delta is zero, setting to default value", "value", 5*time.Minute)
 	}
+	if opts.ExtLookbackDelta == 0 {
+		opts.ExtLookbackDelta = 1 * 24 * time.Hour
+		level.Debug(opts.Logger).Log("msg", " externallookback delta is zero, setting to default value", "value", 1*24*time.Hour)
```
GiedriusS, line 164: Spurious space:
```suggestion
		level.Debug(opts.Logger).Log("msg", "externallookback delta is zero, setting to default value", "value", 1*24*time.Hour)
```

execution/function/functions.go
```diff
@@ -355,6 +355,48 @@ var Funcs = map[string]FunctionCall{
...
+		return promql.Sample{
+			Metric: f.Labels,
+			Point: promql.Point{
+				T: f.StepTime,
+				V: v,
+				H: h,
+			},
+		}
+	},
+	"xdelta": func(f FunctionArgs) promql.Sample {
```
GiedriusS, line 372: I wonder if it is worth adding this behind an option in the engine? Perhaps not and it will be available to everyone who knows about this :smile:

engine/engine_test.go
```diff
@@ -1763,6 +1763,575 @@ func TestBinopEdgeCases(t *testing.T) {
 	testutil.Equals(t, oldResult, newResult)
 }
 
+func TestXFunctions(t *testing.T) {
+	defaultQueryTime := time.Unix(50, 0)
+	fmt.Println(defaultQueryTime.String())
```
GiedriusS, line 1768: Shall we remove this or convert to `t.Log`?

GiedriusS (approved):
LGTM, just a few nits :+1:

execution/function/functions.go
fpetkovski: Having an explicit option makes sense to me as well.

engine/engine.go
```diff
@@ -65,6 +65,10 @@ type Opts struct {
 	// If nil, nothing will be printed.
 	// NOTE: Users will not check the errors, debug writing is best effort.
 	DebugWriter io.Writer
+
+	// ExtLookbackDelta specifies what time range to use to determine valid previous sample for extended range functions.
+	// Defaults to 1 day if not specified.
+	ExtLookbackDelta time.Duration
```
fpetkovski, line 71: This might be too high as a default because the function will buffer 1 day of data for each series. Maybe we could reduce it to 1h and we can still extend it by explicitly setting the parameter.

engine/engine.go
mklocke-shopify: Thanks! Update here: 7425161cf6b2b641ac1fd107e376bed9177a9373

engine/engine_test.go
mklocke-shopify: Good point - updated it here: 7425161cf6b2b641ac1fd107e376bed9177a9373

fpetkovski (approved):
Thanks, this looks good 👍

saswatamcode (approved):
Thanks for awesome work on this!

GiedriusS (approved):

PR #218, opened 2023-03-15
WIP: Add metrics to engine

PradyumnaKrishna:
**Description**
Adds metrics to thanos promql engine.

**Implemented metrics**
* `currentQueries`: View no of queries executing through thanos promql engine.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -35,6 +35,10 @@ import (
 
 type QueryType int
 
+type engineMetrics struct {
+	currentQueries prometheus.Gauge
+}
+
 const (
 	InstantQuery QueryType = 1
 	RangeQuery   QueryType = 2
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -129,6 +133,19 @@ func New(opts Opts) *compatibilityEngine {
...
+		}),
+	}
+
+	if opts.EngineOpts.Reg != nil {
+		opts.EngineOpts.Reg.MustRegister(metrics.currentQueries)
+	}
+
 	return &compatibilityEngine{
 		prom: promql.NewEngine(opts.EngineOpts),
 		queries: promauto.With(opts.Reg).NewCounterVec(
```
GiedriusS, line 151: Maybe this can be moved into the new `engineMetrics`?

engine/engine.go
```diff
@@ -129,6 +133,19 @@ func New(opts Opts) *compatibilityEngine {
...
+		currentQueries: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "thanos",
+			Subsystem: "engine",
+			Name:      "queries",
+			Help:      "The current number of queries being executed or waiting.",
+		}),
+	}
+
+	if opts.EngineOpts.Reg != nil {
+		opts.EngineOpts.Reg.MustRegister(metrics.currentQueries)
```
GiedriusS, line 146: Let's use `promauto.With(opts.Reg)`? It seems like `opts.Reg` is a mandatory parameter.

engine/engine.go
```diff
@@ -129,26 +136,38 @@ func New(opts Opts) *compatibilityEngine {
...
+			prometheus.GaugeOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      "queries",
+				Help:      "The current number of queries being executed or waiting.",
+			},
+		),
 		queries: promauto.With(opts.Reg).NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "promql_engine_queries_total",
```
saswatamcode, line 150: I wonder if the name should be changed to `thanos_engine_queries_total` as well? Wdyt @GiedriusS?

GiedriusS: Makes sense, we can re-use the same pattern as in `currentQueries` :+1:

GiedriusS (approved):

saswatamcode (approved):

PR #219, opened 2023-03-17
Fix grouping by metric name

fpetkovski:
The `labels.HashWithoutLabels` function from Prometheus does not take into account the `__name__` label and explicitly excludes it from the input into the hashing function. This causes grouping expressions by `__name__` to produce an error message in the form of:
```
vector cannot contain metrics with the same labelset.
```

It also interferes with deduplication in the distributed engine since series with different __name__ labels can be treated as duplicates of each other.

This commit replaces the use of `HashWithoutLabels` with `BytesWithoutLabels` in the places where the engine is affected.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -6,6 +6,7 @@ package engine
 import (
 	"context"
 
+	"github.com/cespare/xxhash/v2"
 	"github.com/prometheus/prometheus/model/labels"
 	v1 "github.com/prometheus/prometheus/web/api/v1"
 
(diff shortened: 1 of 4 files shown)


engine/engine.go
```diff
@@ -539,7 +540,7 @@ func containsDuplicateLabelSet(series []labels.Labels) bool {
 	seen := make(map[uint64]struct{}, len(series))
 	for i := range series {
 		buf = buf[:0]
-		h, buf = series[i].HashWithoutLabels(buf)
+		h = xxhash.Sum64(series[i].BytesWithoutLabels(buf))
```
GiedriusS, line 543: Why BytesWithoutLabels instead of Bytes? :thinking:

fpetkovski: Good point, I completely missed that `Bytes` exists :) Should be updated now.

GiedriusS (approved):

MichaHoffmann:
Oh, my bad! Good catch!

PR #220, opened 2023-03-17
Add label based pruning in distributed optimizer

fpetkovski:
The distributed optimizer currently creates a remote query for each available engine. This can increase latency for queries that can be scoped to a subset of engines based on the external labels announced by each engine.

This commit adds support for pruning engines based on their announced external labels. If the selectors for a query do not match any external engine, the query will not be executed at all.

This commit also modifies the distributed optimizer to distribute binary expressions where one operand is a constant value.

### DIFF
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -283,7 +283,6 @@ func isDistributive(expr *parser.Expr) bool {
 }
 
 // matchesExternalLabels returns false if given matchers are not matching external labels.
-// If true, matchesExternalLabels also returns Prometheus matchers without those matching external labels.
 func matchesExternalLabelSet(expr parser.Expr, externalLabelSet []labels.Labels) bool {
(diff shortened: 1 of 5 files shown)


logicalplan/distribute.go
```diff
@@ -222,3 +246,51 @@ func isDistributive(expr *parser.Expr) bool {
 
 	return true
 }
+
+// matchesExternalLabels returns false if given matchers are not matching external labels.
+// If true, matchesExternalLabels also returns Prometheus matchers without those matching external labels.
```
GiedriusS, line 251: Where/how does this happen? This only returns `bool` :smile:

fpetkovski: You're right, I used the store filtering function we have in Thanos as a starting point, but forgot to remove this line.

I don't really see the see the need to override selectors and it also makes the code a bit more complicated.

GiedriusS (approved):

PR #222, opened 2023-03-22
Ignore stale histogram samples

fpetkovski:
Selector operators currently do not ignore native histogram samples. This leads to the sum aggregation producing gaps when a NaN histogram is present in the step vector.


saswatamcode (approved):
LGTM!

PR #223, opened 2023-03-23
Vendor prometheus parser

fpetkovski:
This commit vendors in the promql parser from prometheus/prometheus. The purpose of this is to allow enable parsing new functions without interfering with the promql engine from Prometheus.

This code should ideally be temporary until upstream Prometheus adds support for xrate/xincrease and xdelta.


logicalplan/plan.go
```diff
@@ -4,10 +4,12 @@
 package logicalplan
 
 import (
+	"fmt"
+	"github.com/prometheus/prometheus/model/timestamp"
+	"github.com/prometheus/prometheus/promql"
 	"time"
```
saswatamcode, line 10: This might need to be `goimports`ed

logicalplan/plan.go
```diff
@@ -131,6 +133,112 @@ func traverseBottomUp(parent *parser.Expr, current *parser.Expr, transform func(
 	return true
 }
 
+// preprocessExpr wraps all possible step invariant parts of the given expression with
+// StepInvariantExpr. It also resolves the preprocessors.
+func preprocessExpr(expr parser.Expr, start, end time.Time) parser.Expr {
```
saswatamcode, line 138: I'm guessing this had to be copied over, to work with `internal/prometheus/parser`? Might be worth adding a comment for that! 🙂

fpetkovski:
All good points 👍 . Any idea how we can exclude the internal folder from faillint?

saswatamcode (approved):
LGTM!

Might need to fix the faillint command in Makefile to ignore the vendored parser.

logicalplan/plan.go
fpetkovski: Added

GiedriusS (approved):
:medal_military:

PR #227, opened 2023-03-27
Workaround for Prometheus sampleRing iterator bug 

rabenhorst:
Signed-off-by: Sebastian Rabenhorst <sebastian.rabenhorst@shopify.com>

There is a bug in Prometheus [sampleRing iterator](https://github.com/prometheus/prometheus/blob/211ae4f1f0a2cdaae09c4c52735f75345c1817c6/storage/buffer.go#L204-L378) for mixed native histogram types that repeats the last seen floatHistogram when followed by histograms and using [the `AtFloatHistogram` fn](https://github.com/prometheus/prometheus/blob/211ae4f1f0a2cdaae09c4c52735f75345c1817c6/storage/buffer.go#L280-L285). That results in zero rate when for example querying a native histogram over a downsampled (always type floatHistograms) and non downsampled block (type histogram.)

In this PR we add a workaround in the engine. It will be followed by PR for Prometheus.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -3340,6 +3340,42 @@ func TestNativeHistograms(t *testing.T) {
 	t.Run("float_histograms", func(t *testing.T) {
 		testNativeHistograms(t, cases, opts, generateFloatHistogramSeries)
 	})
+	t.Run("mixed_histograms", func(t *testing.T) {
+		testMixedNativeHistogramTypes(t, opts)
+	})
+}
+
+func testMixedNativeHistogramTypes(t *testing.T, opts promql.EngineOpts) {
+	test, err := promql.NewTest(t, "")
+	testutil.Ok(t, err)
+	defer test.Close()
+
+	var samples []tsdbutil.Sample
+
+	histograms := tsdbutil.GenerateTestHistograms(2)
+
+	samples = append(samples, sample{t: 0, fh: histograms[0].ToFloat()})
+	samples = append(samples, sample{t: 30_000, h: histograms[1]})
+
+	series := storage.NewListSeries(labels.Labels{{Name: "__name__", Value: "native_histogram_series"}}, samples)
+
+	engine := engine.New(engine.Opts{
+		EngineOpts:        opts,
+		DisableFallback:   true,
+		LogicalOptimizers: logicalplan.AllOptimizers,
+	})
+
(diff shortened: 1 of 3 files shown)


engine/engine_test.go
```diff
@@ -3340,6 +3340,49 @@ func TestNativeHistograms(t *testing.T) {
...
+		if i < 1 {
+			samples = append(samples, sample{t: ts, fh: th.ToFloat()})
+		} else {
+			samples = append(samples, sample{t: ts, h: th})
+		}
+	}
+
+	series := storage.NewListSeries(labels.Labels{{Name: "__name__", Value: "native_histogram_series"}}, samples)
+
+	// New Engine
```
fpetkovski, line 3368:
```suggestion
```

engine/engine_test.go
```diff
@@ -3340,6 +3340,49 @@ func TestNativeHistograms(t *testing.T) {
...
+func testMixedNativeHistogramTypes(t *testing.T, opts promql.EngineOpts) {
+	test, err := promql.NewTest(t, "")
+	testutil.Ok(t, err)
+	defer test.Close()
+
+	var samples []tsdbutil.Sample
+
+	histograms := tsdbutil.GenerateTestHistograms(2)
+
+	for i, th := range histograms {
```
fpetkovski, line 3357: Let's use explicit indexes to access histograms. The `for` statement obfuscates what the code tries to do.

fpetkovski (commented):
Looks good, just a few comments from my side.

execution/scan/matrix_selector.go
```diff
@@ -256,13 +256,21 @@ loop:
 		switch buf.Next() {
 		case chunkenc.ValNone:
 			break loop
-		case chunkenc.ValHistogram, chunkenc.ValFloatHistogram:
-			t, h := buf.AtFloatHistogram()
+		case chunkenc.ValHistogram:
+			t, h := buf.AtHistogram()
```
fpetkovski, line 260: We need the same fix here: https://github.com/thanos-community/promql-engine/blob/main/execution/scan/matrix_selector.go#L281

rabenhorst: This should

> Looks good, just a few comments from my side.

I addressed the comments. Can you ptal.

fpetkovski (approved):
LGTM

rabenhorst:
PR for error in distributed engine fuzz test (happens also in main): https://github.com/thanos-community/promql-engine/issues/229

fpetkovski (approved):
Thanks, this looks good now 👍

PR #228, opened 2023-03-28
Add engine param to thanos engine constructor

PradyumnaKrishna:
Prometheus and Thanos engine both uses same registerer to register the metrics. A error raises when we try to register a metric multiple time in promql engine. ~To avoid them I seperated thanos promql engine metrics with prefix `compatibility_engine_`.~

Edit: Added `engine` for fallback engine in EngineOpts. If engine is not provided then creates one otherwise uses the same.
Also, change `prom` type to `v1.QueryEngine` from `*promql.Engine` in thanos `CompatibilityEngine`.

Related: thanos-io/thanos#6234

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -73,6 +73,9 @@ type Opts struct {
 	// EnableXFunctions enables custom xRate, xIncrease and xDelta functions.
 	// This will default to false.
 	EnableXFunctions bool
+
+	// FallbackEngine
+	Engine v1.QueryEngine
 }
 
 func (o Opts) getLogicalOptimizers() []logicalplan.Optimizer {
@@ -193,8 +196,15 @@ func New(opts Opts) *compatibilityEngine {
 		),
 	}
 
+	var engine v1.QueryEngine
+	if opts.Engine == nil {
+		engine = promql.NewEngine(opts.EngineOpts)
+	} else {
+		engine = opts.Engine
+	}
+
 	return &compatibilityEngine{
-		prom: promql.NewEngine(opts.EngineOpts),
+		prom: engine,
 
 		debugWriter:       opts.DebugWriter,
 		disableFallback:   opts.DisableFallback,
@@ -208,7 +218,7 @@ func New(opts Opts) *compatibilityEngine {
 }
 
 type compatibilityEngine struct {
-	prom *promql.Engine
+	prom v1.QueryEngine
 
 	debugWriter io.Writer
 


saswatamcode (approved):
LGTM! Thanks!

engine/engine.go
```diff
@@ -175,6 +175,10 @@ func New(opts Opts) *compatibilityEngine {
 		),
 	}
 
+	if opts.Reg != nil {
+		opts.EngineOpts.Reg = prometheus.WrapRegistererWithPrefix("compatibility_engine_", opts.Reg)
```
fpetkovski, line 179: I am not sure if `compatibility_engine_` is the because this is an implementation detail of how we manage to support all promql functions. Can we use `thanos_engine_` or whatever the prefix is for the main engine?

saswatamcode: So while working with https://github.com/thanos-io/thanos/pull/6234, we noticed that we would basically have 3 engines: the new engine, the new engine's compatibility prometheus engine, and the regular prometheus engine in Thanos. The metrics between the compatibility prometheus engine, and the prometheus engine run in Thanos clash when registering, so the `compatibility_engine` prefix is only added to the compatibility prometheus engine. And we reserve the `thanos_engine` prefix for metrics exported by new engine.

I wonder if there is some better way of handling this. Seems like we are running duplicative engines.

fpetkovski: Right, but what does `compatibility_engine` mean to someone who is using the library?

fpetkovski: Alternatively, we can have a constructor where a user injects a `QueryEngine` so we don't need to instantiate one internally.

saswatamcode: Hmm. For users of this library, I think knowing which queries hit fallback engine, might be somewhat desirable, but we already have a metric for that. 

I think the latter constructor approach might be preferable here.

fpetkovski: 👍 let's try the new constructor then and see if it fixes the problem.

PradyumnaKrishna: Does that mean adding this prefix at `QueryEngineFactory` in thanos, not here?

GiedriusS: If I understood correctly, let's pass an instance of the original Prometheus PromQL engine to the Thanos PromQL engine via the constructor. We can re-use the same, old one, no need to create a new one with a prefix this way.

GiedriusS (commented):
Since it's still early days for this project and we don't have any major versions, my suggestion would be to edit the original functions instead of creating new ones. Thoughts, @fpetkovski @saswatamcode?

engine/engine.go
```diff
@@ -148,6 +152,10 @@ func (l distributedEngine) NewRangeQuery(q storage.Queryable, opts *promql.Query
 }
 
 func New(opts Opts) *compatibilityEngine {
+	return Create(opts, promql.NewEngine(opts.EngineOpts))
+}
+
+func Create(opts Opts, engine *promql.Engine) *compatibilityEngine {
```
PradyumnaKrishna, line 158: Does the `engine` param need to be a `promql.Engine` type? I guess we can make `prom` in `compatibilityEngine` a `v1.QueryEngine` type. Because when thanos creates a prometheus engine it gets type casted into `v1.QueryEngine`.

saswatamcode:
Let's edit the original function 🙂

saswatamcode (approved):
Thanks! This looks good to me!

fpetkovski (approved):
Thanks, lgtm!

PR #231, opened 2023-04-06
Benchmark improvements

charleskorn:
This PR includes a number of small improvements to the benchmarks:
* it renames the instant query benchmark to follow the same naming pattern as the range query benchmark
* it ensures that engine creation is not included in the instant query benchmark
* it adds a benchmark for a query involving a large vector selector
* it adds a benchmark for queries involving a very large number of series

Note to reviewers: it's probably easiest to review each commit individually.

### DIFF
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -21,7 +21,7 @@ import (
 )
 
 func BenchmarkChunkDecoding(b *testing.B) {
-	test := setupStorage(b, 1000, 3)
+	test := setupStorage(b, 1000, 3, 720)
 	defer test.Close()
 
 	start := time.Unix(0, 0)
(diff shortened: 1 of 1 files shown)


yeya24 (approved):
Great work! Really nice improvement.

fpetkovski:
This is indeed great and helps us find cases where we can make further improvements. The only caveat I see is that benchmark takes longer to run and I suspect we might start to see timeouts in the CI. For reference, we track performance over time here: https://thanos-community.github.io/promql-engine/dev/bench.

Is there a way to have an extended mode which will include the large 6 hour and 7 day datasets?

fpetkovski (approved):
Sounds good! 

For local benchmarks we can tweak the code temporarily if we need a faster run.

saswatamcode (approved):
Thanks for great improvements!

PR #232, opened 2023-04-06
Enable all optimizer in fuzz test

yeya24:
Just found out that the propogator optimizer is not enabled in fuzz tests so adding it in this pr.


fpetkovski (approved):
Thanks

PR #235, opened 2023-04-09
Add support for string literals and `label_join` function

Abhra303:
Promql engine does not have a support for string literals (it falls back to the old engine). Because it doesn't, functions like `label_join` and `label_replace` can't be implemented. As promql doesn't deal with strings for calculations, this PR uses the strings directly with functions. This PR adds support for label_join function. The StepInvariantExpr is no longer wrapped over the string literals (because fixed strings do not need it).

Below is the `label_join` function operator tree:

```
[*functionOperator] label_join(http_requests_total, "label", "-", "pod", "series"):
└──[*coalesce]:
   ├──[*concurrencyOperator(buff=2)]:
   │  └──[*vectorSelector] {[__name__="http_requests_total"]} 0 mod 4
   ├──[*concurrencyOperator(buff=2)]:
   │  └──[*vectorSelector] {[__name__="http_requests_total"]} 1 mod 4
   ├──[*concurrencyOperator(buff=2)]:
   │  └──[*vectorSelector] {[__name__="http_requests_total"]} 2 mod 4
   └──[*concurrencyOperator(buff=2)]:
      └──[*vectorSelector] {[__name__="http_requests_total"]} 3 mod 4
```

### DIFF
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -159,13 +159,17 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		}
 
 		// Does not have matrix arg so create functionOperator normally.
-		nextOperators := make([]model.VectorOperator, len(e.Args))
+		nextOperators := make([]model.VectorOperator, 0, len(e.Args))
 		for i := range e.Args {
+			// Strings don't need an operator
+			if e.Args[i].Type() == parser.ValueTypeString {
+				continue
+			}
 			next, err := newOperator(e.Args[i], storage, opts, hints)
 			if err != nil {
 				return nil, err
 			}
-			nextOperators[i] = next
+			nextOperators = append(nextOperators, next)
 		}
 
 		return function.NewFunctionOperator(e, call, nextOperators, stepsBatch, opts)
@@ -181,7 +185,7 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 			return nil, err
 		}
 
-		if e.Param != nil {
+		if e.Param != nil && e.Param.Type() != parser.ValueTypeString {
 			paramOp, err = newOperator(e.Param, storage, opts, hints)
 			if err != nil {
 				return nil, err
@@ -211,8 +215,7 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		return newOperator(e.Expr, storage, opts, hints)
 
 	case *parser.StringLiteral:
-		// TODO(saswatamcode): This requires separate model with strings.
-		return nil, errors.Wrapf(parse.ErrNotImplemented, "got: %s", e)
+		return nil, errors.Newf("string literals don't support operators: %s", e)
 
 	case *parser.UnaryExpr:
 		next, err := newOperator(e.Expr, storage, opts, hints)
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -214,9 +214,6 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 	case *parser.ParenExpr:
 		return newOperator(e.Expr, storage, opts, hints)
 
-	case *parser.StringLiteral:
-		return nil, errors.Newf("string literals don't support operators: %s", e)
-
 	case *parser.UnaryExpr:
 		next, err := newOperator(e.Expr, storage, opts, hints)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


Abhra303:
> I am just not completely sure we need an operator for string literals. Aren't they constant values that are always known and fixed before/during execution?

I also thought about it. But I think string operators would help us visualise the operator tree better for such functions and also, there would be no exceptions in the design of this engine (i.e. handling code for string as non-operator and handling code for other operators). Moreover, I was unsure if it (i.e. if string literals are not operators) would affect future function integrations that may require string literals as operators.

So I decided to make it as operator. I would love to hear your thoughts on this.

mhoffm-aiven:
Can we measure it somehow? If the additional pointer is not that much of a problem this would help since we could also carry a slice for timestamps to implement the `timestamp`  function maybe.

fpetkovski:
It probably won't be a large regression (we can check the benchmarks), but I would really like to keep the step vector object small and focused. I am concerned that we can easily overload it with various nullable fields which can make things harder to debug or cause subtle bugs down the line. 

We should try to infer as much information ahead of time and encode it as static parameters in the query plan instead of adding branching during execution. The latter also makes it harder to work on the engine.

execution/function/operator.go
```diff
@@ -285,10 +292,62 @@ func (o *functionOperator) loadSeries(ctx context.Context) error {
...
+		if o.funcExpr.Func.Name == "label_join" {
+			l := len(o.funcExpr.Args)
+			if l < 3 {
+				err = errors.New("invalid label_join function arguments: atleast 3 arguments required")
+				return
+			}
+
+			labelJoinDstLiteral, ok := o.funcExpr.Args[1].(*parser.StringLiteral)
+			if !ok {
+				err = errors.New("invalid label_join function arguments: dst label is not a string")
```
mhoffm-aiven, line 308: can you write a test with those error scenarios?

mhoffm-aiven:
Ah i understand now what was requested; using the parser here to get the argument is pretty cool; nice PR!

execution/function/operator.go
Abhra303: Yeah, of course I can! Generally, queries like `label_join(foo{}, "label", http_total_requests, 23)`  (i.e. if the arguments are not strings) should trigger these errors.
I just need to create a dedicated `TestLabelJoin` function for that. Because `TestInstantQuery` doesn't check for error string equality.

Abhra303: Seems like these type of invalid cases are handled by parser already. So these checks shouldn't be triggered here. Should I remove these checks?

mhoffm-aiven: Oh i think `TestQueriesAgainstOldEngine` tests also the error messages so that would be a quick way to check. If that is already handled by the parser and we have tests in case this should change then removing the checks seems like a better option!

Abhra303: Yeah, I tested one invalid function `label_join(http_requests_total{}, "label", http_requests{series="3"}, 12)` and the returned error was `unexpected error: 1:44: parse error: expected type string in call to function "label_join", got instant vector` which is coming from the parser.

Abhra303:
Hey @saswatamcode @fpetkovski , could you please approve the workflows?

execution/function/operator.go
```diff
@@ -131,10 +133,12 @@ func NewFunctionOperator(funcExpr *parser.Call, call FunctionCall, nextOps []mod
 	}
 
 	// Check selector type.
-	// TODO(saswatamcode): Add support for string and matrix.
+	// TODO(saswatamcode): Add support for matrix.
 	switch funcExpr.Args[f.vectorIndex].Type() {
 	case parser.ValueTypeVector, parser.ValueTypeScalar:
 		return f, nil
+	case parser.ValueTypeString:
```
fpetkovski, line 140: I think the parser already does validation. Do we still need to handle this case?

execution/execution.go
```diff
@@ -211,8 +215,7 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		return newOperator(e.Expr, storage, opts, hints)
 
 	case *parser.StringLiteral:
-		// TODO(saswatamcode): This requires separate model with strings.
-		return nil, errors.Wrapf(parse.ErrNotImplemented, "got: %s", e)
+		return nil, errors.Newf("string literals don't support operators: %s", e)
```
fpetkovski, line 218: Is it possible to hit this code path? Maybe we can remove the `*parser.StringLiteral` case completely now that we handle strings in the engine itself.

fpetkovski (commented):
Thanks, this looks great. Just some small nits from my side, otherwise looks good.

fpetkovski (approved):
Thanks, lgtm! 👍

yeya24 (approved):
Thanks!

PR #237, opened 2023-04-11
Improve query fuzzing test

yeya24:
Now the query fuzz test just executes once, which is bad in terms of test coverage as it might not execute enough interesting queries for us.

Fixes #237

This pr improves the existing query fuzzing test by:
1. Better test case result collection and display
2. Run fuzzing test for 100 times each

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1651,6 +1651,17 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 				bar 2+2x5`,
 			query: `count by (__name__) ({__name__=~".+"})`,
 		},
+		{
+			name: "scalar with bool",
+			load: `load 30s
+				http_requests_total{pod="nginx-1", series="1"} 1+1.1x40
+				http_requests_total{pod="nginx-2", series="2"} 2+2.3x50
+				http_requests_total{pod="nginx-3", series="3"} 6+0.8x60
+				http_requests_total{pod="nginx-4", series="3"} 5+2.4x50
+				http_requests_total{pod="nginx-5", series="1"} 8.4+2.3x50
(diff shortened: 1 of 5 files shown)


fpetkovski (approved):

fpetkovski:
Looks like we hit some more edge cases with extended fuzzing now, tests seem to be failing.

yeya24:
Yeah I believe it is a known issue https://github.com/thanos-community/promql-engine/issues/200#issuecomment-1456416450.
Do we have any plan to maybe disable `bool modifier` for the new engine and fallback?

fpetkovski:
Some of the failures were definitely real and I tried to fix most of them in https://github.com/yeya24/promql-engine/pull/45.

Others seem to be a bit flaky. For example, `bottomk(1, metric)` can return different series if series have the same values. The result depends on the order in which series will come into the aggregate.

There are some other failures with `atan2` because it sometimes can return either -pi or +pi non-deterministically.

fpetkovski (approved):
Thanks, this is a nice improvement.

Is https://github.com/thanos-community/promql-engine/issues/200 now resolved?

yeya24:
I pushed another commit to sort series before comparing.
I can close https://github.com/thanos-community/promql-engine/issues/200 for now. When I tested the new commits locally I can see those tests passing, but there are some other failed tests, I will try to open an issue once I figure out what happened.

mhoffm-aiven:
Since issues were solved, you could also incorporate the testcases from #210 probably ( though fuzzing on its own does a good enough job probably )

yeya24:
@mhoffm-aiven I will probably merge this pr first. We can still have the tests you added. Please rebase your pr and we can merge yours.

PR #240, opened 2023-04-13
Revert workaround for vector-selector selectPoint 

rabenhorst:
In this PR we revert an unnecessary fix for the vector `selectPoint` fn, which introduced a performance regression for vector select and was part of https://github.com/thanos-community/promql-engine/pull/227. It lead to an additional conversion for a already decoded histogram to float histogram, while before it was a direct decoding to a float histogram.

image (9)


fpetkovski (approved):
Thanks

PR #242, opened 2023-04-18
Allow Engines method to take additional parameters

yeya24:
This allows `RemoteEndpoints` implementation to use different engines based on the query itself.
For example, in my usecase I will select different engines based on query, start, end and lookback delta value.

### DIFF
--- a/api/remote.go
+++ b/api/remote.go
@@ -10,8 +10,16 @@ import (
 	"github.com/prometheus/prometheus/promql"
 )
 
+type Opts struct {
+	Query         string
+	Start         time.Time
+	End           time.Time
+	Step          time.Duration
+	LookbackDelta time.Duration
+}
+
 type RemoteEndpoints interface {
-	Engines() []RemoteEngine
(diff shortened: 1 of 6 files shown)


yeya24:
Kindly ping for review.

fpetkovski:
Is this needed because we cannot add new logical optimizers from outside the engine?

yeya24:
No, even if we can define own optimizer, the engines interface is still unaware of the query itself so it cannot know which remote engines it should pick.

fpetkovski:
But could you then prune engines in the optimizer itself? The same parameters are present in the logical plan optimizer interface.

yeya24:
Got it, I think this might work yeah. But since we cannot define external optimizer it is still not doable for now. This is the only way to extend it.

fpetkovski:
I think it's fine to expose the parser publicly because optimizers are an extension point for the engine. Feel free to raise a PR exposing what you need if that is blocking you right now.

PR #243, opened 2023-04-18
function: capture metric appearance for xrate/xincrease/xdelta

GiedriusS:
Current check for zero value injection doesn't make sense - we need to check it for since the beginning of when the metric appears.

### DIFF
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -405,11 +405,17 @@ loop:
 	case chunkenc.ValHistogram, chunkenc.ValFloatHistogram:
 		t, fh := it.AtFloatHistogram()
 		if t == maxt && !value.IsStaleNaN(fh.Sum) {
+			if *metricAppearedTs == nil {
+				*metricAppearedTs = &t
+			}
 			out = append(out, promql.Sample{T: t, H: fh})
 		}
 	case chunkenc.ValFloat:
 		t, v := it.At()
 		if t == maxt && !value.IsStaleNaN(v) {
+			if *metricAppearedTs == nil {
+				*metricAppearedTs = &t
+			}
 			out = append(out, promql.Sample{T: t, F: v})
 		}
 	}
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -368,13 +368,6 @@ loop:
 		case chunkenc.ValNone:
 			break loop
 		case chunkenc.ValHistogram, chunkenc.ValFloatHistogram:
-			t, h := buf.AtHistogram()
-			if *metricAppearedTs == nil {
-				*metricAppearedTs = &t
-			}
-			if t >= mint {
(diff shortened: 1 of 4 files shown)


GiedriusS:
@mklocke-shopify what do you think about this as you've worked on the original implementation in this engine?

execution/scan/matrix_selector.go
```diff
@@ -32,12 +32,13 @@ type matrixScanner struct {
...
-	storage  engstore.SeriesSelector
-	call     function.FunctionCall
-	scanners []matrixScanner
-	series   []labels.Labels
-	once     sync.Once
+	funcExpr       *parser.Call
+	storage        engstore.SeriesSelector
+	call           function.FunctionCall
+	scanners       []matrixScanner
+	metricAppeared []int64
```
fpetkovski, line 39: Should we put this in the scanner itself? Also does it denote the time when the metric first appeared?

execution/scan/matrix_selector.go
```diff
@@ -155,13 +156,14 @@ func (o *matrixSelector) Next(ctx context.Context) ([]model.StepVector, error) {
...
 			// https://github.com/thanos-community/promql-engine/issues/39
-			result := o.call(function.FunctionArgs{
-				Labels:      series.labels,
-				Samples:     rangeSamples,
-				StepTime:    seriesTs,
-				SelectRange: o.selectRange,
-				Offset:      o.offset,
-			})
+			fa := function.NewFunctionsArgs()
+			fa.Labels = series.labels
```
fpetkovski, line 160: Does it make sense to have a constructor with all parameters so we don't accidentally omit something?

mklocke-shopify (approved):
This looks good! Thanks a lot for adding this, it makes sense.

GiedriusS:
Thx for the review @fpetkovski, I have refactored the code to be more Go-like:

- Instead of `-1` to indicate the absence of a value, let's use a pointer to int64; this makes the zero value of FunctionArgs and others make sense;
- Moved `metricAppeared` into scanners;
- Appended `ts` suffix everywhere so that it would be clear that it stores a timestamp.

execution/scan/matrix_selector.go
```diff
@@ -366,10 +368,20 @@ loop:
 		case chunkenc.ValNone:
 			break loop
 		case chunkenc.ValHistogram, chunkenc.ValFloatHistogram:
+			t, h := buf.AtHistogram()
```
fpetkovski, line 371: Hmm, we call `AtFloatHistogram` bellow, so I wonder why we need to call `AtHistogram` here. We also seem to now append to the output before we check for staleness, so we might end up picking up stale samples. Is this the correct behavior?

fpetkovski (approved):

execution/scan/matrix_selector.go
GiedriusS: You are correct, this was an artifact of rebasing. I somehow missed this part. Thank you for a diligent review :bow:

PR #244, opened 2023-04-22
update Prometheus and add context

yeya24:
Update Prometheus to main and bring https://github.com/prometheus/prometheus/commit/fd3630b9a3c7accecbf7fd438706c43612cfc568.


fpetkovski (approved):
Thanks. How is the context supposed to be used now? Should we propagate it to the query, or will the query get its own context?

yeya24:
In my usecase, I am thinking I can use the context to propagate user ID and use that as part of the query optimizing logic.

PR #245, opened 2023-04-22
Expose parser package

yeya24:
This pr moves the `internal/prometheus/parser` package into `parser` package. This allows external projects able to define their own logicalplan optimizer.


Makefile
```diff
@@ -6,7 +6,7 @@ MDOX_VALIDATE_CONFIG ?= .mdox.validate.yaml
 # if macos, use gsed
 SED ?= $(shell which gsed 2>/dev/null || which sed)
 
-LINT_DIRS = $(shell go list ./... | grep -v "internal/prometheus")
+LINT_DIRS = $(shell go list ./... | grep -v "parser")
```
fpetkovski, line 9: Maybe we want to enable regexps and match for `^parser` to make sure we don't accidentally exclude some nested packages.

Makefile
```diff
@@ -6,7 +6,7 @@ MDOX_VALIDATE_CONFIG ?= .mdox.validate.yaml
 # if macos, use gsed
 SED ?= $(shell which gsed 2>/dev/null || which sed)
 
-LINT_DIRS = $(shell go list ./... | grep -v "internal/prometheus")
+LINT_DIRS = $(shell go list ./... | grep -v "^parser")
```
fpetkovski, line 9: Looks like we need to match the entire package, so maybe something like `grep -v "$(go list -m)/parser"`?

yeya24: It worked thanks!

fpetkovski (approved):

PR #246, opened 2023-04-26
distribute: Add offset to engines to account for range selectors

fpetkovski:
Range selectors in PromQL need to select points from before the start of the range query. When running the distributed optimizer, certain engines might not have sufficient range in the past to correctly calculate the result for certain steps in the range.

To avoid returning incorrect results, this commit calculates the offset that needs to be added to the beginning of a query for each individual remote engine. This PR also adds detection for whether a sufficient time-based overlap exists between engines for queries to be safely distributed. If the overlap is not sufficient, the distributed optimizer will not modify the query and the engine will execute it in memory.

This addresses https://github.com/thanos-community/promql-engine/issues/195. Ideally we want to remove the fallback in the future and figure out how to distribute these queries as well.

### DIFF
--- a/engine/distributed_test.go
+++ b/engine/distributed_test.go
@@ -83,26 +83,23 @@ func TestDistributedAggregations(t *testing.T) {
 	}{
 		{
 			name: "base case",
-			seriesSets: []partition{
-				{
-					extLset: []labels.Labels{labels.FromStrings("zone", "east-1")},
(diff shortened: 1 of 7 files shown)


saswatamcode (approved):
LGTM as far as I understand this!

fpetkovski:
Yeah I think we should move this to a separate workflow and make it not required for now.

PR #252, opened 2023-04-28
Add method to explain query vector

PradyumnaKrishna:
Adds ExplainOutputNode struct with OperatorName and Children, A method `explainVector` generates `ExplainOutputNode` for the provided `VectorOperator`.

This `explainVector` method is used by `Explain` method of query to get an explaination of the query.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -334,8 +334,13 @@ func (e *compatibilityEngine) NewRangeQuery(ctx context.Context, q storage.Query
 type ExplainableQuery interface {
 	promql.Query
 
-	Explain() string
+	Explain() *ExplainOutputNode
 	Profile()
+}
+
+type ExplainOutputNode struct {
+	OperatorName string
+	Children     []ExplainOutputNode
 }
 
 var _ ExplainableQuery = &compatibilityQuery{}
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -346,15 +351,29 @@ type Query struct {
...
 }
 
 func (q *Query) Profile() {
 	// TODO(bwplotka): Return profile.
 }
 
+func explainVector(v model.VectorOperator) *ExplainOutputNode {
+	name, vectors := v.Explain()
+
+	var childs []ExplainOutputNode
```
fpetkovski, line 366:
```suggestion
	var children []ExplainOutputNode
```

engine/engine.go
```diff
@@ -346,15 +351,29 @@ type Query struct {
...
+	for _, vector := range vectors {
+		childs = append(childs, *explainVector(vector))
```
fpetkovski, line 368: Can `explainVector` return `nil`?

GiedriusS (commented):
Maybe we could add some tests for this e.g. `Explain()` on a query that hasn't been executed yet, `Explain()` on a query that has one operator, `Explain()` on a "regular" query like `sum(up) by (job)`.

PradyumnaKrishna:
I guess the query `Explain()` method produce output according to no of processors available (`concurrencyOperator `). That's the cause of failing test.

GiedriusS:
> I guess the query `Explain()` method produce output according to no of processors available (`concurrencyOperator `). That's the cause of failing test.

I think we could use `runtime.GOMAXPROCS(0) / 2` to calculate the number of concurrency operators.

GiedriusS (approved):

PR #255, opened 2023-05-03
Speeding up benchmark tests

alanprot:
Creating the load string in the benchmark test is taking huge time.

Using a string builder now to create that string.


yeya24 (approved):
Thanks! Do we know how much could this improve the test?

yeya24:
I will merge this change and see how much it improves :)

alanprot:
On my computer it took more than 10 min just to create the string... after the change was quite fast.

PR #257, opened 2023-05-11
Add comprehensive test and support for timestamp function

### LINKED ISSUE #206: Add support for timestamp function
The timestamp function is currently broken and we've decided to revert it in https://github.com/thanos-community/promql-engine/pull/204.

The correct way to implement it is to return each sample's timestamp instead of the step timestamp. This will make sure that samples returned due to the engine lookback have their original timestamp in the output. 

One way to do it is to detect the presence of a the `timestamp` function in a query (maybe through an optimizer), and push it down into the vector selector. The selector will then store timestamps directly into the values of output vectors.

nishchay-veer:
According to my understanding of the codebase, promql.Point is not defined in the Prometheus Go client library. Here's the definition of promql.Point that needs to be added in value.go file:
type Point struct {
    T int64
    V float64
}
also we can add a list of all the Points.  It then loops over all the filtered samples and appends a promql.Point struct to its Points field with its timestamp and value (converted to seconds).
Resolve #206

### DIFF
--- a/execution/scan/vector_selector.go
+++ b/execution/scan/vector_selector.go
@@ -192,6 +192,8 @@ func selectPoint(it *storage.MemoizedSeriesIterator, ts, lookbackDelta, offset i
 	refTime := ts - offset
 	var p point
 
+	returnTimestampAsValue := true // bool flag to indicate whether to return the timestamp as the point value.
+
 	valueType := it.Seek(refTime)
 	switch valueType {
 	case chunkenc.ValNone:
@@ -203,7 +205,13 @@ func selectPoint(it *storage.MemoizedSeriesIterator, ts, lookbackDelta, offset i
 		p = point{t: t, fh: fh}
 	case chunkenc.ValFloat:
 		t, v := it.At()
-		p = point{t: t, v: v}
+		if returnTimestampAsValue {
+			p = point{t: t, v: float64(t)}
+
+		} else {
+			p = point{t: t, v: v}
+		}
+
 	default:
 		panic(errors.Newf("unknown value type %v", valueType))
 	}


execution/function/functions.go
```diff
@@ -120,6 +120,18 @@ var Funcs = map[string]FunctionCall{
...
 		}
 	},
+	"timestamp": func(f FunctionArgs) promql.Sample {
+		if len(f.Samples) == 0 {
+			return InvalidSample
+		}
+		result := promql.Sample{}
+		for _, sample := range f.Samples {
+			if sample.H == nil { // only consider float samples
+				result.Points = append(result.Points, promql.Point{T: sample.T, V: float64(sample.T) / 1000})
```
MichaHoffmann, line 130: Hey, this cannot be implemented as a function i think, it will return the step timestamp for nested functions for example.

execution/scan/vector_selector.go
```diff
@@ -165,35 +187,38 @@ func (o *vectorSelector) loadSeries(ctx context.Context) error {
 }
 
 // TODO(fpetkovski): Add max samples limit.
-func selectPoint(it *storage.MemoizedSeriesIterator, ts, lookbackDelta, offset int64) (int64, float64, *histogram.FloatHistogram, bool, error) {
+// To push down the timestamp function into this file and store the timestamp in the value for each series, you can modify the selectPoint function.
+func selectPoint(it *storage.MemoizedSeriesIterator, ts, lookbackDelta, offset int64) (point, bool, error) {
```
fpetkovski, line 191: My suggestion would be to roll back the changes and use a `bool` flag here to indicate whether to return the timestamp as the point value. Then for the `case chunkenc.ValFloat` case in the switch statement we can return `t` as the point value instead of `v`.

nishchay-veer: Is it possible to return the timestamp as the point value in` case chunkenc.ValFloat `without the need to set a boolean flag for it? This would allow us to handle non-histogram data directly.

fpetkovski: Yes, the return types would remain as they are. The idea is to use a `bool` as an argument to the function in order to decide whether the `chunkenc.ValFloat` case would return `t` or `v`.

I guess we would need to do the same for histograms, based on this boolean we would either return the histogram itself or `t` as the `float64` return value.

nishchay-veer: Okay I got the idea behind it !

execution/scan/vector_selector.go
```diff
@@ -49,6 +49,12 @@ type vectorSelector struct {
 	numShards int
 }
 
+type point struct {
```
fpetkovski, line 52: Do we still need this new struct?

nishchay-veer: well I made this just to have a good relation between timestamp and values in our data.

nishchay-veer: what changes do we need now ?

PR #261, opened 2023-05-14
#138 : Added absent_over_time support function

AryanSharma9917:


execution/function/functions.go
```diff
@@ -87,6 +87,13 @@ var Funcs = map[string]FunctionCall{
 	"deg": simpleFunc(func(v float64) float64 {
 		return v * 180 / math.Pi
 	}),
+	"absent_over_time": simpleFunc(func(v float64) float64 {
```
MichaHoffmann, line 90: i dont think the implementation works; this operator actually needs to be a source of new samples ( if there are no samples in the window ). only implemented as simpleFunc it can only transform preexisting samples. Can you take a look at my previous PR for the `absent` function? I had to introduce a new operator that actually produces new samples in that case. It might be possible to adapt this for the "_over_time" variant too, but it might be a bit harder because functions are pushed down into the matrix selector

AryanSharma9917: Okey sure!

PR #265, opened 2023-05-19
execution: fix scalar function

MichaHoffmann:
this should fix issues with `scalar(non_existent)` problems in fuzzing.

### DIFF
--- a/execution/function/scalar.go
+++ b/execution/function/scalar.go
@@ -1,0 +1,59 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package function
+
+import (
+	"context"
+	"math"
+
+	"github.com/prometheus/prometheus/model/labels"
+
+	"github.com/thanos-community/promql-engine/execution/model"
+)
+
+type scalarFunctionOperator struct {
+	pool *model.VectorPool
+	next model.VectorOperator
+}
+
+func (o *scalarFunctionOperator) Explain() (me string, next []model.VectorOperator) {
+	return "[*scalarFunctionOperator]", []model.VectorOperator{}
+}
+
+func (o *scalarFunctionOperator) Series(ctx context.Context) ([]labels.Labels, error) {
+	return nil, nil
+}
+
+func (o *scalarFunctionOperator) GetPool() *model.VectorPool {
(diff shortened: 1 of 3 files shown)


execution/function/scalar.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, nil
+	}
+
+	in, err := o.next.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	result := o.GetPool().GetVectorBatch()
+	if len(in) == 0 {
```
fpetkovski, line 53: We should not need this anymore right?

MichaHoffmann: ok rebased, it works as intended :tada:

execution/function/scalar.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(in) == 0 {
+		return nil, nil
+	}
+
+	result := o.GetPool().GetVectorBatch()
+	for _, vector := range in {
+		sv := o.GetPool().GetStepVector(vector.T)
```
fpetkovski, line 53: I noticed that in the original implementation we modify the input vector in place: https://github.com/thanos-io/promql-engine/pull/265/files#diff-45b283ce3f8010b8cd1c8c50b568e54e58a974d8b4ba75b14e8821680641b0fcL219.

Can we do the same here?

MichaHoffmann: i think we could do that yes

execution/function/scalar.go
```diff
@@ -0,0 +1,64 @@
...
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+	in, err := o.next.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if len(in) == 0 {
```
fpetkovski, line 47:
```suggestion
	if in == nil {
```

fpetkovski (approved):
Looks great 👍

execution/function/scalar.go
MichaHoffmann: note: tried and it did fail some range query tests; decided to go back to using the pool from the scalar operator itself.

PR #266, opened 2023-05-19
Adding regex matcher optimizer

sharadgaur:
Optimizing the regular expression if input string does not contain any regex literal. This optimization step will convert the input query from up{app=~"something"} to up{app="something"}

### DIFF
--- a/logicalplan/regex_matchers.go
+++ b/logicalplan/regex_matchers.go
@@ -1,0 +1,38 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package logicalplan
+
+import (
+	"strings"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/thanos-community/promql-engine/parser"
+)
+
+const hasRegexChars = ".+|[]{}^$*?+()\\"
+
+// RegexMatchersFunctions function will optimize the query by avoiding unnecessary regex match.
+// For example: up{app=~"something"} will be optimized to up{app="something"}.
+type RegexMatchersFunctions struct {
+}
+
+func (RegexMatchersFunctions) Optimize(expr parser.Expr, _ *Opts) parser.Expr {
+	traverse(&expr, func(node *parser.Expr) {
+		switch e := (*node).(type) {
+		case *parser.VectorSelector:
+			for matcherIndex, lm := range e.LabelMatchers {
+				if lm.Type == labels.MatchRegexp {
+					if !strings.ContainsAny(lm.Value, hasRegexChars) {
+						nm, err := labels.NewMatcher(labels.MatchEqual, lm.Name, lm.Value)
+						if err != nil {
+							continue
+						}
+						e.LabelMatchers[matcherIndex] = nm
+					}
+				}
+			}
+		}
+	})
+	return expr
+}
--- a/logicalplan/regex_matchers.go
+++ b/logicalplan/regex_matchers.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 
 	"github.com/prometheus/prometheus/model/labels"
-	"github.com/thanos-community/promql-engine/parser"
+	"github.com/thanos-io/promql-engine/parser"
 )
 
 const hasRegexChars = ".+|[]{}^$*?+()\\"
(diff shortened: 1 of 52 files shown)


MichaHoffmann:
I think prometheus already does this: https://github.com/prometheus/prometheus/blob/92d69803606620505ed8f0226abb640a121ec1cd/tsdb/querier_test.go#L1840

fpetkovski:
> I think prometheus already does this

The problem is that Thanos does not, and the Querier and all Stores end up parsing and compiling regular expressions for these cases.

yeya24:
Can you share the code? @fpetkovski I think store gateway has the same optimization in the posting group code, right?

fpetkovski:
We see high CPU usage here for `NewFastRegexMatcher` which is called from here: https://github.com/thanos-io/thanos/blob/main/pkg/store/storepb/custom.go#L396.

yeya24:
Got it. I think we saw same thing in Prometheus. A lot of time spend on `NewFastRegexMatcher`. Ideally it would be good to cache the matchers.
Then this optimization makes sense to me.

logicalplan/regex_matchers.go
```diff
@@ -0,0 +1,37 @@
...
+
+package logicalplan
+
+import (
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/thanos-community/promql-engine/parser"
+	"regexp"
+)
+
+var hasRegexString = regexp.MustCompile(`^[^\\\\[{()^$.*?|+]*$`)
```
MichaHoffmann, line 12: Is the regular expression here necessary? Probably it would be safer to use `regexp/syntax` here, parse the regular expression and check that the topmost opcode is "OpLiteral"?

mhoffm-aiven: Or is parsing the expensive part?

sharadgaur: compiling a regular expression is generally faster than parsing the regular expression because `regexp.MustCompile` function caches the compiled regular expression, allowing subsequent uses of the regular expression to be faster.

mhoffm-aiven: I meant to check if the incoming regular expression is a literal; but probably parsing here is just making the problem worse

mhoffm-aiven: couldnt we do this without the regex and a simple `strings.ContainsAny(regexFromLabel, specialRegexChars)` ?

sharadgaur: No, we don't want change the query if input string contains valid expression. If you have better idea please feel free to share.

mhoffm-aiven: Wouldn't that be equivalent though? I meant if it doesn't contain special regex chars it's probably a literal and can be optimized with this, right?

sharadgaur: ah yes

yeya24:
Do we have any benchmark? I am wondering how much could this help by using equal matcher rather than regex matcher?

MichaHoffmann:
Ah i finally understand; probably its grafana sending lots of regex matchers with literals!

MichaHoffmann:
thsi is in prometheus now https://github.com/prometheus/prometheus/pull/12434

logicalplan/regex_matchers.go
```diff
@@ -0,0 +1,38 @@
...
+type RegexMatchersFunctions struct {
+}
+
+func (RegexMatchersFunctions) Optimize(expr parser.Expr, _ *Opts) parser.Expr {
+	traverse(&expr, func(node *parser.Expr) {
+		switch e := (*node).(type) {
+		case *parser.VectorSelector:
+			for matcherIndex, lm := range e.LabelMatchers {
+				if lm.Type == labels.MatchRegexp {
+					if !strings.ContainsAny(lm.Value, hasRegexChars) {
```
GiedriusS, line 26: This is different to https://github.com/prometheus/prometheus/pull/12434/files#diff-c6bfdf5ae2a5edbc3df844f48f297e4d2a4f4106662f1a04c8931fbb38f1fdaeR83, though.

MichaHoffmann: the way in the prometheus pr was the only way i could come up with to make sure that its a literal without parsing though, is it problematic?

sharadgaur:
@yeya24 added the banchmark test.

yeya24:
Do we still need the optimizer with https://github.com/prometheus/prometheus/pull/12434 merged upstream?

mhoffm-aiven:
I think we need to bump prometheus dependency in thanos still though!

PR #270, opened 2023-05-21
execution: fix khashaggregate comparison logic

MichaHoffmann:
This aligns comparison logic with [upstream promql](https://github.com/prometheus/prometheus/blob/92d69803606620505ed8f0226abb640a121ec1cd/promql/engine.go#LL2703C34-L2703C82) .

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -1609,6 +1609,13 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			step:  2 * time.Second,
 		},
 		{
+			name: "topk with NaN comparison",
+			load: `load 30s
+            http_requests_total{pod="nginx-1", route="/"} NaN
+            http_requests_total{pod="nginx-2", route="/"}  NaN`,
+			query: "topk by (route) (1, http_requests_total)",
+		},
+		{
 			name: "nested topk error that should not be skipped",
 			load: `load 30s
 				X 1+1x50`,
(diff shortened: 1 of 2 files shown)


fpetkovski:
Do you think the fuzzing failure can be related
```
    enginefuzz_test.go:133: load 30s
        			http_requests_total{pod="nginx-1", route="/"} 1.00+1.00x40
        			http_requests_total{pod="nginx-2", route="/"}  1+2.00x40
    enginefuzz_test.go:134: bottomk without (pod) (
          (scalar(stddev without () ({__name__="http_requests_total"} offset 1m12s)) != bool pi()),
          sinh({__name__="http_requests_total",route="/"})
        )
    enginefuzz_test.go:136: case 75 error mismatch.
        new result: Scalar value NaN overflows int64
        old result: {pod="nginx-1", route="/"} => 1.1752011936438014 @[0]
    enginefuzz_test.go:141: failed 1 test cases
```

MichaHoffmann:
Thats most likely scalar(not_existent) issue i tried to fix here: https://github.com/thanos-io/promql-engine/pull/265/files

```
scalar(non_existent_metric) != bool pi()
```
this should result in 1 but does result in NaN on main because of issues with scalar. bottomk then rejects NaN on this PR as it should but prometheus will report bottomk(1, X) since the nested function returns different results.

fpetkovski (approved):

PR #271, opened 2023-05-21
Fix step generation in selectors

fpetkovski:
Selector operators have a bug where if no series for a given timestamp exists,
the step vector will not be generated for that step. This complicates
the implementation of the absent function, and could also cause subtle bugs
in downstream operators which expect all end-result steps to be present in a batch.

This commit fixes the issue by making sure all steps for a batch are generated
in the matrix and vector selectors before returning the batch.


MichaHoffmann (approved):
lgtm

MichaHoffmann (approved):

saswatamcode (approved):
LGTM!

PR #274, opened 2023-05-23
Fix repo references

fpetkovski:
We migrated the repository from the thanos-community to the thanos-io organization.
This updates references to reflect the new org.

cc @yeya24 @alanprot


fpetkovski:
The docs CI fails because `github.com/thanos-io/promql-engine` does not exist in https://pkg.go.dev/. I guess it would get published as soon as we have one commit on main.

saswatamcode (approved):
LGTM!

Let's merge this and ensure docs check from next commit onwards. 🙂

yeya24 (approved):
Thanks

PR #279, opened 2023-05-27
execution: add relabel operator; add label_replace function

MichaHoffmann:
* put relabel functions into their own operator to simplify the `Series` of the functionOperator
* add `label_replace` function


fpetkovski (approved):
Nice work, I hope most of it is ported over from Prometheus :)

MichaHoffmann:
> Nice work, I hope most of it is ported over from Prometheus :)

Yeah it was, just adapted to the conventions from the label_join operator but the logic is the same.

PR #280, opened 2023-05-29
execution: start implementing subqueries

MichaHoffmann:
caveats: 
* only instant queries for now

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -33,6 +33,7 @@ import (
 	"github.com/thanos-io/promql-engine/extlabels"
 	"github.com/thanos-io/promql-engine/logicalplan"
 	"github.com/thanos-io/promql-engine/parser"
+	"github.com/thanos-io/promql-engine/query"
 )
 
 type QueryType int
@@ -47,6 +48,7 @@ const (
 	subsystem    string    = "engine"
 	InstantQuery QueryType = 1
 	RangeQuery   QueryType = 2
+	stepsBatch             = 10
 )
 
 type Opts struct {
@@ -71,6 +73,10 @@ type Opts struct {
 	// EnableXFunctions enables custom xRate, xIncrease and xDelta functions.
 	// This will default to false.
 	EnableXFunctions bool
+
+	// EnableSubqueries enables the engine to handle subqueries without falling back to prometheus.
+	// This will default to false.
+	EnableSubqueries bool
 
 	// FallbackEngine
 	Engine v1.QueryEngine
@@ -219,6 +225,10 @@ func New(opts Opts) *compatibilityEngine {
 		metrics:           metrics,
 		extLookbackDelta:  opts.ExtLookbackDelta,
 		enableAnalysis:    opts.EnableAnalysis,
+		enableSubqueries:  opts.EnableSubqueries,
+		noStepSubqueryIntervalFn: func(d time.Duration) time.Duration {
+			return time.Duration(opts.NoStepSubqueryIntervalFn(d.Milliseconds()) * 1000000)
+		},
 	}
 }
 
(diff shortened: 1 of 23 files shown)


execution/subquery/subquery.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if o.currentStep > o.maxt {
+		return nil, nil
+	}
+	if err := o.initSeries(ctx); err != nil {
+		return nil, err
+	}
+
+ACC:
+	for {
```
saswatamcode, line 64: What is the reasoning behind this loop?

MichaHoffmann: the subquery might entail multiple steps of the underlying operator ( and since this PR so far only implements instant query subqueries for sum_over_time, its probably fine to just comsume all of them at the same time )

engine/engine.go
```diff
@@ -311,7 +328,18 @@ func (e *compatibilityEngine) NewRangeQuery(ctx context.Context, q storage.Query
...
+		Start:                    start,
+		End:                      end,
+		Step:                     step,
+		StepsBatch:               stepsBatch,
+		LookbackDelta:            opts.LookbackDelta,
+		ExtLookbackDelta:         e.extLookbackDelta,
+		NoStepSubqueryIntervalFn: e.noStepSubqueryIntervalFn,
+	}
+
+	exec, err := execution.New(lplan.Expr(), q, qOpts)
```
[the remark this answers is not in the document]
MichaHoffmann: I agree! Lets issue a separate PR for that i'd say though.

execution/subquery/subquery.go
MichaHoffmann: I think it works like this for instant queries; range queries will be a bit more work though; wdyt of leaving it like this for instant queries with a big disclaimer comment and getting to work for range queries in subsequent PR?

MichaHoffmann:
benchmark results from the one contrived query:
```
goos: linux
goarch: amd64
pkg: github.com/thanos-io/promql-engine/engine
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkInstantQuery/subquery_sum_over_time/old_engine-8         	      15	  68072512 ns/op	 3433186 B/op	   33436 allocs/op
BenchmarkInstantQuery/subquery_sum_over_time/new_engine-8         	      42	  26401382 ns/op	 4345721 B/op	   42269 allocs/op
PASS
ok  	github.com/thanos-io/promql-engine/engine	10.983s
```

GiedriusS (commented):
LGTM as a first step, let's see if anyone else has any comments or suggestions

execution/execution.go
```diff
@@ -120,6 +108,29 @@ func newOperator(expr parser.Expr, storage *engstore.SelectorPool, opts *query.O
 		// before it can be non-nested. https://github.com/thanos-io/promql-engine/issues/39
 		for i := range e.Args {
 			switch t := e.Args[i].(type) {
+			case *parser.SubqueryExpr:
```
fpetkovski, line 111: Shouldn't this be a top-level `case` instead of having it here in the function args?

fpetkovski (commented):
Looks good overall, just one question from my side.

As a side note, I found the refactoring to query.Options a bit distracting from the main change  introduced by the PR.

fpetkovski (approved):
lgtm

fpetkovski:
@MichaHoffmann is this good to merge?

MichaHoffmann:
> @MichaHoffmann is this good to merge?

Yeah! I rebased on latest main and it passes promql acceptance tests too !

execution/scan/subquery.go
```diff
@@ -0,0 +1,142 @@
...
+		var series []labels.Labels
+		series, err = o.next.Series(ctx)
+		if err != nil {
+			return
+		}
+
+		o.series = make([]labels.Labels, len(series))
+		o.acc = make([][]sample, len(series))
+		for i, s := range series {
+			var b labels.ScratchBuilder
```
fpetkovski, line 133: We should move this out of the loop and call `Reset` on it here.

MichaHoffmann: Moved it out of the loop; `Reset` is already called in `DropMetricName` though so i left it out.

GiedriusS (commented):
I think it's also worth adding how subqueries support works like to README :smile:

MichaHoffmann:
> I think it's also worth adding how subqueries support works like to README 😄

Yeah, valid remark; Ill add a small section in a new PR in a minute

PR #282, opened 2023-05-31
fmt: format load strings in tests

MichaHoffmann:
Adds a small tool so we can auto-format indentation for load statements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,9 @@ check-docs: $(MDOX)
 format: $(GOIMPORTS)
 	@echo ">> formatting go code"
 	@gofmt -s -w $(FILES_TO_FMT)
+	@echo ">> formatting promql tests"
+	@go run scripts/testvet/main.go -json -fix ./...
+	@echo ">> formatting imports"
 	@$(GOIMPORTS) -w $(FILES_TO_FMT)
 
 .PHONY:lint
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -137,37 +137,37 @@ func BenchmarkRangeQuery(b *testing.B) {
 	}{
 		{
 			name:    "vector selector",
-			query:   "http_requests_total",
+			query:   `http_requests_total`,
 			storage: sixHourDataset,
 		},
 		{
 			name:    "sum",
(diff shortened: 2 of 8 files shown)


ci/loadvet/main.go
saswatamcode, line 1: Maybe we can just put this in `scripts` dir?

saswatamcode, line 1: Maybe making this into its own module, so that it is separately installable might be better, in case we want to upstream it for use in prometheus/prometheus at some point? 🙂

ci/loadvet/main.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			if s.Kind != token.STRING {
+				return true
+			}
+			quote := s.Value[0]
+			if quote != '`' {
+				return true
+			}
+			cont := s.Value[1 : len(s.Value)-1]
+			if !strings.HasPrefix(cont, "load") {
```
saswatamcode, line 62: I think there are other directives as well like `clear` and `eval`: https://github.com/prometheus/prometheus/blob/main/promql/test.go#L252

saswatamcode (commented):
Nice! 

We should probably hook this into our CI as well, by adding it to `format` target.

ci/loadvet/main.go
MichaHoffmann: can do!

ci/loadvet/main.go
MichaHoffmann: i was intending it only to format our tests for now; to do it properly we would need to be able to parse those load directives.

GiedriusS (commented):
Maybe we can to run this during CI and error out if there are any changes after running this tool?

fpetkovski:
Running it in the CI seems like a good idea 👍

ci/loadvet/main.go
MichaHoffmann: done!

MichaHoffmann:
> Running it in the CI seems like a good idea 👍

hooked it up to the "format" job

saswatamcode (approved):

fpetkovski (approved):

PR #283, opened 2023-06-02
Add support for histogram and scalar binop

fpetkovski:
This commit adds support for operations between a histogram and a scalar.

### DIFF
--- a/execution/binary/scalar.go
+++ b/execution/binary/scalar.go
@@ -120,12 +120,12 @@ func (o *scalarOperator) Next(ctx context.Context) ([]model.StepVector, error) {
 	out := o.pool.GetVectorBatch()
 	for v, vector := range in {
 		step := o.pool.GetStepVector(vector.T)
-		for i := range vector.Samples {
-			scalarVal := math.NaN()
-			if len(scalarIn) > v && len(scalarIn[v].Samples) > 0 {
-				scalarVal = scalarIn[v].Samples[0]
-			}
+		scalarVal := math.NaN()
+		if len(scalarIn) > v && len(scalarIn[v].Samples) > 0 {
+			scalarVal = scalarIn[v].Samples[0]
+		}
 
+		for i := range vector.Samples {
 			operands := o.getOperands(vector, i, scalarVal)
 			val, keep := o.floatOp(operands, o.operandValIdx)
 			if o.returnBool {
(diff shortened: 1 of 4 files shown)


execution/binary/scalar.go
```diff
@@ -138,6 +140,19 @@ func (o *scalarOperator) Next(ctx context.Context) ([]model.StepVector, error) {
 			}
 			step.AppendSample(o.pool, vector.SampleIDs[i], val)
 		}
+
+		for i := range vector.HistogramIDs {
+			scalarVal := math.NaN()
```
MichaHoffmann, line 145: couldnt we move `scalarVal` out of the `for i := range` loop since its not dependent on i and used for samples and histograms?

fpetkovski: Good point, updated 👍

execution/binary/table.go
```diff
@@ -201,6 +199,42 @@ func newOperation(expr parser.ItemType, vectorBinOp bool) (operation, error) {
...
+	var operation histogramFloatOperation
+	if scalarSide == ScalarSideLeft {
+		operation = rhsHistogramOperations[t]
+	} else {
+		operation = lhsHistogramOperations[t]
+	}
+	if operation != nil {
+		return operation
+	}
+	return undefinedHistogramOp
```
MichaHoffmann, line 235: shouldnt that return something like "unimplemented" to trigger promql fallback?

fpetkovski: The fallback can only be applied at plan compilation time when we detect that a function or operator is implemented in the engine. In this specific case, we want to cover operators that are simply not "defined" between a histogram and float operand, even in Prometheus. 

For example, something like a % b is only defined when both a and b are floats. But to know their type, we have to actually execute the query and decode chunks. To further complicate matters, a series can have interleaved histograms and floats, so for some samples the operation will be defined and for others it won't. 

So bottom line, it should be fine to produce a nil value when an operation between a histogram and float is simply not defined, as in the a % b example.

MichaHoffmann: ah! ok i understand; can you add a small test to ensure that we are behaving like prometheus in that case?

MichaHoffmann: nvm, i have found 
```
If either operator has to sum up a mix of histogram samples and float samples, the corresponding vector element is removed from the output vector entirely.
```

MichaHoffmann: ok, i think this will just work!

MichaHoffmann (approved):
lgtm

saswatamcode (approved):
LGTM!

PR #288, opened 2023-06-07
Include histograms from remote instant queries

fpetkovski:
Native histograms are not loaded from remote instant queries.
    
This commit fixes that.


MichaHoffmann (approved):
LGTM

PR #289, opened 2023-06-18
 Time consumed by each operator

nishchay-veer:
This pull request enhances the codebase by introducing the `model.OperatorTelemetry` interface, allowing for better tracking and analysis of CPU time usage by operators.
• Added a new interface called `model.OperatorTelemetry`, which includes the methods `AddCPUTimeTaken `and `CPUTimeTaken` for each operator.
• The `AddCPUTimeTaken` method allows recording the CPU time taken by an operator during its execution.
• Included a new test  `TestQueryAnalyze` to verify that the query analysis produces valid results, including non-zero CPU time values, for different types of queries.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -382,7 +382,7 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
 	telemetry, obsVectors := obsv.Analyze()
-	name, _ := obsv.(model.VectorOperator).Explain()
+	name, _ := obsv.Explain()
 	var children []AnalyzeOutputNode
 	for _, vector := range obsVectors {
 		children = append(children, *analyzeVector(vector))
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -5,6 +5,7 @@ package engine
 
 import (
 	"context"
+
 	"io"
 	"math"
 	"runtime"
@@ -371,8 +372,11 @@ func (q *Query) Explain() *ExplainOutputNode {
 }
 
 func (q *Query) Analyze() *AnalyzeOutputNode {
+	if observableRoot, ok := q.exec.(model.ObservableVectorOperator); ok {
+		return analyzeVector(observableRoot)
+	}
+	return nil
 
-	return analyzeVector(q.exec.(model.ObservableVectorOperator))
 }
 
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -731,7 +731,7 @@ func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext s
 	telemetry, next := o.Analyze()
 	_, _ = w.Write([]byte(indent))
 	_, _ = w.Write([]byte("Operator Time :"))
-	_, _ = w.Write([]byte(strconv.FormatInt(int64(telemetry.CPUTimeTaken()), 10)))
+	_, _ = w.Write([]byte(strconv.FormatInt(int64(telemetry.ExecutionTimeTaken()), 10)))
 	if len(next) == 0 {
 		_, _ = w.Write([]byte("\n"))
 		return
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -376,7 +376,6 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 		return analyzeVector(observableRoot)
 	}
 	return nil
-
 }
 
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
@@ -390,7 +389,6 @@ func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
 		OperatorTelemetry: telemetry,
 		Children:          children,
 	}
-
 }
 
 func explainVector(v model.VectorOperator) *ExplainOutputNode {
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -348,7 +348,7 @@ type ExplainableQuery interface {
 	Analyze() *AnalyzeOutputNode
 }
 type AnalyzeOutputNode struct {
-	OperatorTelemetry *model.TimingInformation
+	OperatorTelemetry model.OperatorTelemetry
 	Children          []AnalyzeOutputNode
 }
 
@@ -374,6 +374,7 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 
 	return analyzeVector(q.exec.(model.ObservableVectorOperator))
 }
+
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
 	telemetry, obsVectors := obsv.Analyze()
 	var children []AnalyzeOutputNode
@@ -726,7 +727,7 @@ func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext s
 
 	// _, _ = w.Write([]byte(indent))
 	_, _ = w.Write([]byte("Operator Time :"))
-	_, _ = w.Write([]byte(strconv.FormatInt(int64(telemetry.CPUTime), 10)))
+	_, _ = w.Write([]byte(strconv.FormatInt(int64(telemetry.(*model.TimingInformation).CPUTime), 10)))
 	if len(next) == 0 {
 		_, _ = w.Write([]byte("\n"))
 		return
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -350,8 +350,9 @@ type ExplainableQuery interface {
 }
 
 type AnalyzeOutputNode struct {
-	OperatorTelemetry model.OperatorTelemetry
-	Children          []AnalyzeOutputNode
+	OperatorName      string                  `json:"name,omitempty"`
+	OperatorTelemetry model.OperatorTelemetry `json:"telemetry,omitempty"`
+	Children          []AnalyzeOutputNode     `json:"children,omitempty"`
 }
 
 type ExplainOutputNode struct {
@@ -381,12 +382,14 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
 	telemetry, obsVectors := obsv.Analyze()
+	name, _ := obsv.(model.VectorOperator).Explain()
 	var children []AnalyzeOutputNode
 	for _, vector := range obsVectors {
 		children = append(children, *analyzeVector(vector))
 	}
 
 	return &AnalyzeOutputNode{
+		OperatorName:      name,
 		OperatorTelemetry: telemetry,
 		Children:          children,
 	}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -348,8 +348,8 @@ type ExplainableQuery interface {
 	Analyze() *AnalyzeOutputNode
 }
 type AnalyzeOutputNode struct {
-	OperatorTime *model.TimingInformation
-	Children     []AnalyzeOutputNode
+	OperatorTelemetry *model.TimingInformation
+	Children          []AnalyzeOutputNode
 }
 
 type ExplainOutputNode struct {
@@ -376,15 +376,15 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 	return analyzeVector(q.obs)
 }
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
-	ti, obs_vectors := obsv.Analyze()
+	telemetry, obs_vectors := obsv.Analyze()
 	var children []AnalyzeOutputNode
 	for _, vector := range obs_vectors {
 		children = append(children, *analyzeVector(vector))
 	}
 
 	return &AnalyzeOutputNode{
-		OperatorTime: ti,
-		Children:     children,
+		OperatorTelemetry: telemetry,
+		Children:          children,
 	}
 
 }
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -737,18 +737,16 @@ func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext s
 		return
 	}
 
-	_, _ = w.Write([]byte(indent))
-	_, _ = w.Write([]byte("Children:\n"))
+	_, _ = w.Write([]byte(":\n"))
 
-	for i, obsVector := range next {
+	for i, n := range next {
 		if i == len(next)-1 {
-			analyze(w, obsVector, indentNext+"└──", indentNext+"   ")
+			analyze(w, n, indentNext+"└──", indentNext+"   ")
 		} else {
-			analyze(w, obsVector, indentNext+"├──", indentNext+"│  ")
+			analyze(w, n, indentNext+"└──", indentNext+"   ")
 		}
 	}
 }
-
 func explain(w io.Writer, o model.VectorOperator, indent, indentNext string) {
 	me, next := o.Explain()
 	_, _ = w.Write([]byte(indent))
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -737,16 +737,18 @@ func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext s
 		return
 	}
 
-	_, _ = w.Write([]byte(":\n"))
+	_, _ = w.Write([]byte(indent))
+	_, _ = w.Write([]byte("Children:\n"))
 
-	for i, n := range next {
+	for i, obsVector := range next {
 		if i == len(next)-1 {
-			analyze(w, n, indentNext+"└──", indentNext+"   ")
+			analyze(w, obsVector, indentNext+"└──", indentNext+"   ")
 		} else {
-			analyze(w, n, indentNext+"└──", indentNext+"   ")
+			analyze(w, obsVector, indentNext+"├──", indentNext+"│  ")
 		}
 	}
 }
+
 func explain(w io.Writer, o model.VectorOperator, indent, indentNext string) {
 	me, next := o.Explain()
 	_, _ = w.Write([]byte(indent))
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -375,9 +375,9 @@ func (q *Query) Analyze() *AnalyzeOutputNode {
 	return analyzeVector(q.exec.(model.ObservableVectorOperator))
 }
 func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
-	telemetry, obs_vectors := obsv.Analyze()
+	telemetry, obsVectors := obsv.Analyze()
 	var children []AnalyzeOutputNode
-	for _, vector := range obs_vectors {
+	for _, vector := range obsVectors {
 		children = append(children, *analyzeVector(vector))
 	}
 
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"runtime"
 	"sort"
+	"strconv"
 	"time"
 
 	"github.com/cespare/xxhash/v2"
@@ -514,7 +515,7 @@ func (q *compatibilityQuery) Exec(ctx context.Context) (ret *promql.Result) {
 	switch e := q.expr.(type) {
 	case *parser.StringLiteral:
 		if q.debugWriter != nil {
-			analyze(q.debugWriter, q.exec.(model.ObservableVectorOperator), "", "")
+			analyze(q.debugWriter, q.exec.(model.ObservableVectorOperator), " ", "")
 		}
 		return &promql.Result{Value: promql.String{V: e.Val, T: q.ts.UnixMilli()}}
 	}
@@ -721,29 +722,27 @@ func recoverEngine(logger log.Logger, expr parser.Expr, errp *error) {
 		*errp = errors.Wrap(err, "unexpected error")
 	}
 }
-
 func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext string) {
-	ti, obsVectors := o.Analyze()
+	telemetry, next := o.Analyze()
 
-	_, _ = w.Write([]byte(indent))
-	_, _ = w.Write([]byte(fmt.Sprintf("Operator Time: %v\n", ti.CPUTime)))
-
-	if len(obsVectors) == 0 {
+	// _, _ = w.Write([]byte(indent))
+	fmt.Println("Operator Time : ")
+	_, _ = w.Write([]byte(strconv.FormatInt(int64(telemetry.CPUTime), 10)))
+	if len(next) == 0 {
+		_, _ = w.Write([]byte("\n"))
 		return
 	}
 
-	_, _ = w.Write([]byte(indent))
-	_, _ = w.Write([]byte("Children:\n"))
+	_, _ = w.Write([]byte(":\n"))
 
-	for i, obsVector := range obsVectors {
-		if i == len(obsVectors)-1 {
-			analyze(w, obsVector, indentNext+"└──", indentNext+"   ")
+	for i, n := range next {
+		if i == len(next)-1 {
+			analyze(w, n, indentNext+"└──", indentNext+"   ")
 		} else {
-			analyze(w, obsVector, indentNext+"├──", indentNext+"│  ")
+			analyze(w, n, indentNext+"└──", indentNext+"   ")
 		}
 	}
 }
-
 func explain(w io.Writer, o model.VectorOperator, indent, indentNext string) {
 	me, next := o.Explain()
 	_, _ = w.Write([]byte(indent))
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -348,6 +348,7 @@ type ExplainableQuery interface {
 	Explain() *ExplainOutputNode
 	Analyze() *AnalyzeOutputNode
 }
+
 type AnalyzeOutputNode struct {
 	OperatorTelemetry model.OperatorTelemetry
 	Children          []AnalyzeOutputNode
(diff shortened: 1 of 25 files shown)


yeya24:
This only tracks the time consumed when initializing the operator. I think what we need is mainly time consumed when calling `series`, `Next`, etc

execution/function/scalar.go
```diff
@@ -6,15 +6,32 @@ package function
...
+	next     model.VectorOperator
+	ti       model.TimingInformation
+	duration time.Duration
+}
+
+func (o *scalarFunctionOperator) Analyze() (*model.TimingInformation, []model.ObservableVectorOperator) {
+
+	// Create a TimingInformation instance and add CPU time taken
+	timingInfo := &model.TimingInformation{}
+	timingInfo.AddCPUTimeTaken(o.duration)
```
saswatamcode, line 27: I think we should be calling this in `Next`, to populate the time? `Analyze` should only return the populated `TimingInformation` struct.

execution/scan/literal_selector.go
```diff
@@ -7,13 +7,23 @@ import (
...
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+var (
+	constantMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "operator_duration_seconds",
+		Help: "Time taken by each operator",
+	}, []string{"operator", "duration"})
+)
+
```
saswatamcode, line 26: This metric can be removed now I think, saw we are doing this in operator interface?

execution/scan/literal_selector.go
```diff
@@ -26,10 +36,12 @@ type numberLiteralSelector struct {
...
 	once        sync.Once
 
-	val float64
+	val        float64
+	recordTime bool
+	duration   time.Duration
 }
 
-func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val float64) *numberLiteralSelector {
+func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val float64, recordTime bool) *numberLiteralSelector {
```
saswatamcode, line 44: Also we don't really need the recordTime option here, I think

execution/function/scalar.go
```diff
@@ -6,15 +6,32 @@ package function
...
+	pool     *model.VectorPool
+	next     model.VectorOperator
+	ti       model.TimingInformation
+	duration time.Duration
+}
+
+func (o *scalarFunctionOperator) Analyze() (*model.TimingInformation, []model.ObservableVectorOperator) {
+
+	// Create a TimingInformation instance and add CPU time taken
+	timingInfo := &model.TimingInformation{}
```
saswatamcode, line 26: Let initialize this in the operator constructor (for this that'd be `NewFunctionOperator`).

We actually need to check if analyze was enabled, to either init this with `model.TimingInformation` or `model.NoopTimingInformation`.

execution/function/scalar.go
```diff
@@ -6,15 +6,32 @@ package function
...
+	timingInfo.AddCPUTimeTaken(o.duration)
+
+	// Create a slice to store any observable vector operators found during analysis
+	observableOperators := make([]model.ObservableVectorOperator, 0)
```
saswatamcode, line 30: Not quite. We aren't really finding  The semantics for this should be same as `Explain`, that is, it should return any next operators it can find.

execution/model/operator.go
```diff
@@ -5,10 +5,30 @@ package model
...
+
+type TimingInformation struct {
+	CPUTime time.Duration
+}
+
+func (ti *TimingInformation) AddCPUTimeTaken(t time.Duration) {
+	ti.CPUTime += t
+}
+
+type ObservableVectorOperator interface {
```
saswatamcode, line 25: We should probably use this interface while building the tree in execution.go

execution/function/scalar.go
```diff
@@ -6,15 +6,32 @@ package function
...
 type scalarFunctionOperator struct {
-	pool *model.VectorPool
-	next model.VectorOperator
+	pool     *model.VectorPool
+	next     model.VectorOperator
+	ti       model.TimingInformation
+	duration time.Duration
+}
+
+func (o *scalarFunctionOperator) Analyze() (*model.TimingInformation, []model.ObservableVectorOperator) {
```
saswatamcode, line 23: Hmm, we should probably hook this up to something like the `ExplainableQuery` interface.

saswatamcode (commented):
Great start! I think we need to make a few things clear before we proceed with this.

engine/engine.go
```diff
@@ -358,8 +371,22 @@ func (q *Query) Explain() *ExplainOutputNode {
...
 }
 
-func (q *Query) Profile() {
-	// TODO(bwplotka): Return profile.
+func (q *Query) Analyze() *AnalyzeOutputNode {
+
+	return analyzeVector(q.obs)
+}
+func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
+	ti, obs_vectors := obsv.Analyze()
```
saswatamcode, line 379: Let's not use snake case here, and name it a bit descriptively. 🙂 
```suggestion
	telemetry, obsVectors := obsv.Analyze()
```

engine/engine.go
```diff
@@ -326,18 +333,23 @@ func (e *compatibilityEngine) NewRangeQuery(ctx context.Context, q storage.Query
...
 
 type ExplainableQuery interface {
 	promql.Query
 
 	Explain() *ExplainOutputNode
-	Profile()
+	Analyze() *AnalyzeOutputNode
+}
+type AnalyzeOutputNode struct {
+	OperatorTime *model.TimingInformation
```
saswatamcode, line 351: Let's name it a bit generically, so that we can add other stuff in the future.
```suggestion
	OperatorTelemetry *model.TimingInformation
```

engine/engine.go
```diff
@@ -688,6 +723,28 @@ func recoverEngine(logger log.Logger, expr parser.Expr, errp *error) {
 	}
 }
 
+func analyze(w io.Writer, o model.ObservableVectorOperator, indent, indentNext string) {
```
saswatamcode, line 726: Nice! Got an example of an output with this?

execution/execution.go
```diff
@@ -52,7 +53,7 @@ const stepsBatch = 10
 
 // New creates new physical query execution for a given query expression which represents logical plan.
 // TODO(bwplotka): Add definition (could be parameters for each execution operator) we can optimize - it would represent physical plan.
-func New(ctx context.Context, expr parser.Expr, queryable storage.Queryable, mint, maxt time.Time, step, lookbackDelta, extLookbackDelta time.Duration) (model.VectorOperator, error) {
+func New(ctx context.Context, expr parser.Expr, queryable storage.Queryable, mint, maxt time.Time, step, lookbackDelta, extLookbackDelta time.Duration, enableAnalysis bool) (model.ObservableVectorOperator, error) {
```
saswatamcode, line 56: Why do we need to change the return type for this? We can probably just typecast it where we need to?

execution/scan/literal_selector.go
```diff
@@ -39,8 +40,21 @@ func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val f
 		currentStep: opts.Start.UnixMilli(),
 		val:         val,
 	}
+
+	// op.t = &model.NoopTimingInformation{}
```
saswatamcode, line 44: Any particular reason to comment this out? 🙂

engine/engine.go
```diff
@@ -350,6 +362,7 @@ var _ ExplainableQuery = &compatibilityQuery{}
 type Query struct {
 	exec model.VectorOperator
 	opts *promql.QueryOpts
+	obs  model.ObservableVectorOperator
```
saswatamcode, line 365: Why don't we just use exec here?

execution/scan/literal_selector.go
nishchay-veer: Because theAnalyze() methdod only returns Timing information struct , so I wasn't sure how to return 't' from `numberLiteralSelector`

engine/engine.go
nishchay-veer: Hmm, but then how can we Analyze the query, because the Analyze method is only a part of ObservableVectorOperator interface

saswatamcode: We can typecast it later on?

execution/scan/literal_selector.go
saswatamcode: Hmm, but that isn't fixed. Our suggestions are there to just give you a structure of what it could look like, you can modify the types to suit

nishchay-veer: I think we can first check if the o.t field is of type *model.TimingInformation using a type assertion and then return it

saswatamcode: Try pushing a commit with then 🙂

engine/engine.go
nishchay-veer: Screenshot from 2023-07-07 23-06-02
This is what i get on local testing the analyze function

saswatamcode: Why not add a code test? 🙂

execution/scan/literal_selector.go
```diff
@@ -39,8 +40,25 @@ func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val f
...
+	op.t = &model.NoopTimingInformation{}
+	if opts.EnableAnalysis {
+		op.t = &model.TimingInformation{CPUTime: time.Duration(0)}
+	}
+
+	return op
 }
+func (o *numberLiteralSelector) Analyze() (*model.TimingInformation, []model.ObservableVectorOperator) {
+	if telemetry, ok := o.t.(*model.TimingInformation); ok {
+		_, _ = o.Next(context.Background())
```
saswatamcode, line 53: Ok, I think we are a bit confused here. Do we need to call Next in Analyze? 

Next is only meant to be called for the query execution. Calling Next here would basically mean calling Next twice on each query if analyze param is enabled in https://github.com/thanos-io/thanos/pull/6515/files. Anaylze should only return timing for that operator. The recording of the duration is done just once in Next

execution/scan/literal_selector.go
```diff
@@ -39,8 +40,25 @@ func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val f
...
+		return telemetry, nil
+	}
+	return nil, nil
 
+}
+func (o *numberLiteralSelector) AddCPUTimeTaken(t time.Duration) {
+	o.t.AddCPUTimeTaken(t)
```
GiedriusS, line 60: I would suggest just embedding `model.OperatorTelemetry` directly so that we wouldn't have to write these functions. It would simply be possible to call `AddCPUTimeTaken` directly: https://go.dev/play/p/sYvyjlMaZh0

execution/model/operator.go
```diff
@@ -5,10 +5,34 @@ package model
...
+type TimingInformation struct {
+	CPUTime time.Duration
+}
+
+func (ti *TimingInformation) AddCPUTimeTaken(t time.Duration) {
+	ti.CPUTime += t
+}
+
+type OperatorTelemetry interface {
+	AddCPUTimeTaken(time.Duration)
```
GiedriusS, line 26:
```suggestion
        AddCPUTimeTaken(time.Duration)
	CPUTimeTaken() time.Duration
```

engine/engine.go
```diff
@@ -358,8 +370,22 @@ func (q *Query) Explain() *ExplainOutputNode {
...
 }
 
-func (q *Query) Profile() {
-	// TODO(bwplotka): Return profile.
+func (q *Query) Analyze() *AnalyzeOutputNode {
+
+	return analyzeVector(q.exec.(model.ObservableVectorOperator))
+}
+func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
+	telemetry, obs_vectors := obsv.Analyze()
```
GiedriusS, line 378:
```suggestion
	telemetry, obsVectors := obsv.Analyze()
```

execution/scan/literal_selector.go
```diff
@@ -27,10 +27,11 @@ type numberLiteralSelector struct {
 	once        sync.Once
 
 	val float64
+	t   model.OperatorTelemetry
```
GiedriusS, line 30: Don't forget to embed this directly into the operator - this would remove some code duplication:

```suggestion
	model.OperatorTelemetry
```

execution/scan/literal_selector.go
```diff
@@ -83,6 +100,8 @@ func (o *numberLiteralSelector) Next(ctx context.Context) ([]model.StepVector, e
 		o.step = 1
 	}
 	o.currentStep += o.step * int64(o.numSteps)
+	duration := time.Since(start)
+	o.AddCPUTimeTaken(duration)
```
GiedriusS, line 104:
```suggestion
	o.AddCPUTimeTaken(time.Since(start))
```

We typically avoid creating variables that are used on the next line or if they are only used in one place.

engine/engine_test.go
```diff
@@ -112,6 +113,63 @@ func TestQueryExplain(t *testing.T) {
...
+		expected *engine.AnalyzeOutputNode
+	}{
+		{
+			query:    "34",
+			expected: &engine.AnalyzeOutputNode{OperatorTelemetry: &model.TimingInformation{}, Children: concurrencyOperators},
+		},
+	} {
+		{
+			t.Run(tc.query, func(t *testing.T) {
+				ng := engine.New(engine.Opts{EngineOpts: opts})
```
GiedriusS, line 149: Don't forget to enable the analysis functionality here :smile:

engine/engine_test.go
```diff
@@ -112,6 +113,63 @@ func TestQueryExplain(t *testing.T) {
...
+		concurrencyOperators = append(concurrencyOperators, engine.AnalyzeOutputNode{
+			OperatorTelemetry: &model.TimingInformation{}, Children: []engine.AnalyzeOutputNode{
+				{OperatorTelemetry: &model.TimingInformation{}},
+			},
+		})
+	}
+
+	for _, tc := range []struct {
+		query    string
+		expected *engine.AnalyzeOutputNode
```
GiedriusS, line 140: Good stuff but the test could be improved even further if we would check whether the [CPUTime](https://github.com/thanos-io/promql-engine/pull/289/files#diff-a738ddf76b5904ba1562ec7baa7ddd82de3c3303bc8e1d006e24ba990df9a4edR18) has some value. We could add `assertFn func(t *testing.T, got *engine.AnalyzeOutputNode)` here or something like that. Inside there you could write an assertion that `got.CPUTime > 0`.

engine/engine.go
```diff
@@ -358,8 +370,23 @@ func (q *Query) Explain() *ExplainOutputNode {
 	return explainVector(q.exec)
 }
 
-func (q *Query) Profile() {
-	// TODO(bwplotka): Return profile.
+func (q *Query) Analyze() *AnalyzeOutputNode {
+
```
GiedriusS, line 374: Spurious new line.

engine/engine.go
```diff
@@ -358,8 +370,23 @@ func (q *Query) Explain() *ExplainOutputNode {
...
+	return analyzeVector(q.exec.(model.ObservableVectorOperator))
```
GiedriusS, line 375: This will crash if the type assertion fails. Perhaps we could return `nil` if this fails?

```go
observableRoot, ok := q.exec.(model.ObservableVectorOperator)
if !ok { return nil }
...
```

execution/scan/literal_selector.go
```diff
@@ -39,6 +40,21 @@ func NewNumberLiteralSelector(pool *model.VectorPool, opts *query.Options, val f
...
+
+	op.OperatorTelemetry = &model.NoopTimingInformation{}
+	if opts.EnableAnalysis {
+		op.OperatorTelemetry = &model.TimingInformation{CPUTime: time.Duration(0)}
+	}
+
+	return op
+}
+
+func (o *numberLiteralSelector) Analyze() (model.OperatorTelemetry, []model.ObservableVectorOperator) {
```
GiedriusS, line 52: I think we could just do `return o, nil`

execution/function/operator.go
```diff
@@ -91,15 +103,11 @@ func newNoArgsFunctionOperator(funcExpr *parser.Call, stepsBatch int, opts *quer
 		call:        call,
 		vectorPool:  model.NewVectorPool(stepsBatch),
 	}
+  op.OperatorTelemetry = &model.NoopTimingInformation{}
```
GiedriusS, line 106: Indentation seems off.

fpetkovski:
Skimming through the PR quickly, I am curious why we have to do so many explicit type checks. Can we embed behavior in the types themselves?

execution/aggregate/khashaggregate.go
```diff
@@ -138,6 +142,23 @@ func (a *kAggregate) GetPool() *model.VectorPool {
 	return a.vectorPool
 }
 
+func (a *kAggregate) Analyze() (model.OperatorTelemetry, []model.ObservableVectorOperator) {
+	if _, ok := a.OperatorTelemetry.(*model.TimingInformation); ok {
```
GiedriusS, line 146: Yep, we don't need these. I think Nischay just forgot to remove.

nishchay-veer: Yeah I have to remove it at some places, but I think we need to type check to check if it is of the type `TimingInformation` and not `NoopTimingInformation` to return it in Analyze.

GiedriusS: Why? `kAggregate` already implements `model.OperatorTelemetry` so we can return it directly, no?

nishchay-veer: Type checks removed

execution/function/absent.go
```diff
@@ -19,6 +20,11 @@ type absentOperator struct {
 	series   []labels.Labels
 	pool     *model.VectorPool
 	next     model.VectorOperator
+	model.OperatorTelemetry
+}
+
+func (o *absentOperator) Analyze() (model.OperatorTelemetry, []model.ObservableVectorOperator) {
+	return o, []model.ObservableVectorOperator{}
```
saswatamcode, line 27: Shouldn't we return `o.next` here?

engine/engine_test.go
```diff
@@ -112,6 +113,71 @@ func TestQueryExplain(t *testing.T) {
...
+	assertCPUTimeNonZero := func(t *testing.T, got *engine.AnalyzeOutputNode) {
+		if got != nil {
+			if got.OperatorTelemetry.CPUTimeTaken() <= 0 {
+				t.Errorf("expected non-zero CPUTime, got %v ", got.OperatorTelemetry.CPUTimeTaken())
+			}
+		}
+	}
+
+	for _, tc := range []struct {
+		query string
```
saswatamcode, line 136: Let's try with some complex queries.

engine/engine_test.go
```diff
@@ -112,6 +113,71 @@ func TestQueryExplain(t *testing.T) {
...
+		[]float64{1, 2, 3, 4, 5, 6},
+		[]string{labels.MetricName, "foo"},
+	)
+
+	start := time.Unix(0, 0)
+	end := time.Unix(1000, 0)
+
+	assertCPUTimeNonZero := func(t *testing.T, got *engine.AnalyzeOutputNode) {
+		if got != nil {
+			if got.OperatorTelemetry.CPUTimeTaken() <= 0 {
```
saswatamcode, line 129: Hmm, I think we should check for the children nodes too, and what structure those nodes have?

engine/engine_test.go
```diff
@@ -112,6 +113,71 @@ func TestQueryExplain(t *testing.T) {
...
+	series := storage.MockSeries(
+		[]int64{240, 270, 300, 600, 630, 660},
+		[]float64{1, 2, 3, 4, 5, 6},
+		[]string{labels.MetricName, "foo"},
+	)
+
+	start := time.Unix(0, 0)
+	end := time.Unix(1000, 0)
+
+	assertCPUTimeNonZero := func(t *testing.T, got *engine.AnalyzeOutputNode) {
```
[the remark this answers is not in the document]
nishchay-veer: I think we can stick with the current function and simply incorporate modifications to check for child operators. Introducing something like `testutil.Assert(t, explainableQuery.Analyze().OperatorTelemetry.CPUTimeTaken() > 0)` could result in less clean code as it might complicate the process of verifying child operators.

execution/function/operator.go
```diff
@@ -38,30 +38,53 @@ func NewFunctionOperator(funcExpr *parser.Call, nextOps []model.VectorOperator,
...
+		return s, nil
+
 	case "label_join", "label_replace":
-		return &relabelFunctionOperator{
+		r := &relabelFunctionOperator{
 			next:     nextOps[0],
 			funcExpr: funcExpr,
-		}, nil
+		}
+		r.OperatorTelemetry = &model.NoopTimingInformation{}
```
GiedriusS, line 56: You can implement this logic once and then pass it to each operator. This way you would avoid copy/paste.

execution/exchange/coalesce.go
```diff
@@ -42,15 +44,33 @@ type coalesce struct {
...
+	return c
+}
+
+func (c *coalesce) Analyze() (model.OperatorTelemetry, []model.ObservableVectorOperator) {
+	obsOperators := make([]model.ObservableVectorOperator, len(c.operators))
+	for i, operator := range c.operators {
+		if obsOperator, ok := operator.(model.ObservableVectorOperator); ok {
+			obsOperators[i] = obsOperator
+		} else {
+			obsOperators[i] = nil
```
GiedriusS, line 70: This is the only odd Analyze(). Let's do this instead:

```go
obsOperators := make([]model.ObservableVectorOperator, 0, len(c.operators))
for i, operator := range c.operators {
  if obsOperator, ok := operator.(model.ObservableVectorOperator); ok {
    obsOperators = append(obsOperators, obsOperator)
  }
}
return c, obsOperators
```

execution/model/operator.go
```diff
@@ -5,10 +5,41 @@ package model
...
+	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+type NoopTimingInformation struct{}
+
+func (ti *NoopTimingInformation) AddCPUTimeTaken(t time.Duration) {}
+
+type TimingInformation struct {
```
GiedriusS, line 17: I think in the future we will be using this struct for not just capturing CPU timing information so maybe it would be worth changing names in this PR before merging?

I suggest something like `TrackedTelemetry` and `NoopTelemetry`?

engine/engine.go
```diff
@@ -326,18 +334,24 @@ func (e *compatibilityEngine) NewRangeQuery(ctx context.Context, q storage.Query
...
 type ExplainableQuery interface {
 	promql.Query
 
 	Explain() *ExplainOutputNode
-	Profile()
+	Analyze() *AnalyzeOutputNode
+}
+
+type AnalyzeOutputNode struct {
+	OperatorTelemetry model.OperatorTelemetry
```
GiedriusS, line 353: Let's add `OperatorName string` like in `ExplainOutputNode`

engine/engine_test.go
GiedriusS: :+1: let's add this test in the future to the fuzzers

engine/engine.go
```diff
@@ -358,8 +373,26 @@ func (q *Query) Explain() *ExplainOutputNode {
...
+func (q *Query) Analyze() *AnalyzeOutputNode {
+	if observableRoot, ok := q.exec.(model.ObservableVectorOperator); ok {
+		return analyzeVector(observableRoot)
+	}
+	return nil
+}
+
+func analyzeVector(obsv model.ObservableVectorOperator) *AnalyzeOutputNode {
+	telemetry, obsVectors := obsv.Analyze()
+	name, _ := obsv.(model.VectorOperator).Explain()
```
saswatamcode, line 385: We don't need to typecast here.
```suggestion
	name, _ := obsv.Explain()
```

engine/engine.go
```diff
@@ -358,8 +373,26 @@ func (q *Query) Explain() *ExplainOutputNode {
 	return explainVector(q.exec)
 }
 
-func (q *Query) Profile() {
-	// TODO(bwplotka): Return profile.
+func (q *Query) Analyze() *AnalyzeOutputNode {
+	if observableRoot, ok := q.exec.(model.ObservableVectorOperator); ok {
+		return analyzeVector(observableRoot)
```
saswatamcode, line 378: We could also pass the explain tree here directly, but it would probably not change much.

engine/engine.go
saswatamcode, line 385: This might add some additional overhead, but we can iterate on that.

execution/function/operator.go
```diff
@@ -29,36 +30,56 @@ type functionOperator struct {
...
 	call         functionCall
 	scalarPoints [][]float64
+	model.OperatorTelemetry
+}
+
+func SetTelemetry(t model.OperatorTelemetry, opts *query.Options) {
+	t = &model.NoopTelemetry{}
+	if opts.EnableAnalysis {
+		t = &model.TrackedTelemetry{}
+	}
```
saswatamcode, line 40: This causes lint to fail. Let's return simply,
```suggestion
func SetTelemetry(opts *query.Options) model.OperatorTelemetry {
	if opts.EnableAnalysis {
		return &model.TrackedTelemetry{}
	}
	return &model.NoopTelemetry{}
```

execution/function/operator.go
```diff
@@ -29,36 +30,56 @@ type functionOperator struct {
...
+		return s, nil
+
 	case "label_join", "label_replace":
-		return &relabelFunctionOperator{
+		r := &relabelFunctionOperator{
 			next:     nextOps[0],
 			funcExpr: funcExpr,
-		}, nil
+		}
+		SetTelemetry(r.OperatorTelemetry, opts)
```
saswatamcode, line 60: With comment above we can then use it as,
```suggestion
			OperatorTelemetry: SetTelemetry(opts),
		}
```

saswatamcode (commented):
Thanks! I think now this is in a good place, some comments!

engine/engine.go
GiedriusS: Why not add the name to the original struct? This will be quite costly :sweat_smile:

nishchay-veer: We can return name as well in `Analyze() (string, OperatorTelemetry, []ObservableVectorOperator)` for each operator ?

GiedriusS: I think we can add the name to the struct and add a method to `OperatorTelemetry` like `Name() string` which would return the operator's name. :thinking: I think that would make the most sense

execution/model/operator.go
```diff
@@ -5,10 +5,63 @@ package model
 
 import (
 	"context"
+	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+type NoopTelemetry struct {
+	name string
```
saswatamcode, line 14: Do we need this for noop? `SetName` can just be blank function, and `Name` can return empty string. :)

saswatamcode (approved):
LGTM! Awesome work!
Let's merge and iterate on this! :)

GiedriusS (approved):

PR #290, opened 2023-06-27
execution: add timing information

GiedriusS:
Let's try using embedding to capture observability information for each operator instead of wrapping operator inside of operator. Main reasoning for this approach instead of the other way is to be able to catch more granular data like memory allocations. With wrapping we would have to have cross-references between operators and this seems cleaner to me.

Need to explore this further to see how it looks like.

### DIFF
--- a/execution/function/scalar.go
+++ b/execution/function/scalar.go
@@ -15,6 +15,7 @@ import (
 type scalarFunctionOperator struct {
 	pool *model.VectorPool
 	next model.VectorOperator
+	model.TimingInformation
 }
 
 func (o *scalarFunctionOperator) Explain() (me string, next []model.VectorOperator) {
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -326,18 +333,23 @@ func (e *compatibilityEngine) NewRangeQuery(ctx context.Context, q storage.Query
...
-		t:      RangeQuery,
+		Query:       &Query{exec: exec, opts: opts},
+		engine:      e,
+		expr:        expr,
+		t:           RangeQuery,
+		debugWriter: e.debugWriter,
 	}, nil
 }
 
 type ExplainableQuery interface {
```
GiedriusS, line 344: Maybe we can call it `ObservableQuery` now because it's not just Explain() anymore.

Issue #291, opened 2023-07-07
Enable injecting user-defined operators


fpetkovski:
We currently have mechanisms to reshape the logical plan using custom optimizers which can be implemented by third-party code. Such optimizers can even inject custom logical nodes which satisfy the `promql.Node` interface, but we lack a way for them to be translated into execution operators. This is because the translation logic is hardcoded in [execution.go](https://github.com/thanos-io/promql-engine/blob/main/execution/execution.go).

To enable extensibility through custom execution operators, we could maybe define a new interface for a logical node along the lines of:
```
type UserDefinedOperator interface {
  MakeExecutionOperator(<all params>) model.VectorOperator
}
```

With this mechanism, a logical optimizers we can inject custom logical nodes and the engine will know how to translate them to execution operators.

PR #293, opened 2023-07-10
Add support for injecting custom operators

fpetkovski:
This commit adds support for injecting custom operators into the engine. The machanism involves creating a custom optimizer which produces a logical node implementing the `UserDefinedExpr` interface.

This interface has a `MakeExecutionOperator` method which the user can leverage to define how the execution operator is created for the logical node.

Fixes https://github.com/thanos-io/promql-engine/issues/291.

### DIFF
--- a/logicalplan/user_defined.go
+++ b/logicalplan/user_defined.go
@@ -1,0 +1,22 @@
+package logicalplan
+
+import (
+	"github.com/prometheus/prometheus/storage"
+
+	"github.com/thanos-io/promql-engine/execution/model"
+	engstore "github.com/thanos-io/promql-engine/execution/storage"
+	"github.com/thanos-io/promql-engine/parser"
+	"github.com/thanos-io/promql-engine/query"
+)
+
+// UserDefinedExpr is an extension point which allows users to define their execution operators.
+type UserDefinedExpr interface {
+	parser.Expr
+	MakeExecutionOperator(
+		stepsBatch int,
+		vectors *model.VectorPool,
+		selectors *engstore.SelectorPool,
+		opts *query.Options,
+		hints storage.SelectHints,
+	) (model.VectorOperator, error)
+}
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The current implementation uses goroutines very liberally which means the query 
 
 ### Plan optimization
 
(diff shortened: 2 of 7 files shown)


logicalplan/user_defined.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/prometheus/prometheus/storage"
+
+	"github.com/thanos-io/promql-engine/execution/model"
+	engstore "github.com/thanos-io/promql-engine/execution/storage"
+	"github.com/thanos-io/promql-engine/parser"
+	"github.com/thanos-io/promql-engine/query"
+)
+
+// UserDefinedExpr is an extension point which allows users to define their execution operators.
+type UserDefinedExpr interface {
```
saswatamcode, line 13: While I generally agree with this, I'm not sure how this extensibility would work practically. 
With such a model, wouldn't the user essentially need to fork, add custom dep (which has their custom `UserDefinedExpr` defined) to Thanos Query and rebuild?

saswatamcode (commented):
Awesome! 💪🏻 

A small usability question:

logicalplan/user_defined.go
fpetkovski: Hm what do you mean by the user needs to fork?

Maybe the test here shows how it can be extended without forking? https://github.com/thanos-io/promql-engine/pull/293/files#diff-c7f33eb59c81b304bf23f7373af474aa330c6eada4f9d2e8b56eb77010ee7176

Using an optimizer we can inject or replace a logical node with our own. As long as that new node implements `UserDefinedExpr`, the engine will call `MakeExecutionOperator` on it, which is again a user-implemented function: https://github.com/thanos-io/promql-engine/pull/293/files#diff-ac999eb9f6193331831ad85e398328d9b516bcfa685c5cc9282ca8761fe60d51R206-R207

saswatamcode: Yes, I get that from the test. By forking, I didn't mean this promql-engine package, but rather the upstream where this is used like Thanos. 

Afaiu, you'd need to actually pass the logical node that implements `UserDefinedExpr` somehow, into Thanos, if you'd like to use this in some env. For that, wouldn't you need to rebuild Thanos with the custom logical node?

fpetkovski: The engine `opts` allow customizing the optimizers used for the logical plan. The idea is that you can write an optimizer which injects a `UserDefinedExpr` anywhere in the plan.

saswatamcode: So something like https://github.com/thanos-io/thanos/pull/6186 (although this is using strings in a flag to load optional "built-in" logical operators)?

fpetkovski: This is more of a feature for the engine and not so much for Thanos itself. For example, I am experimenting with querying different storage formats and I don't want to fork the engine, or to add operators that are not specific to PromQL.

I can't think of a use case inside Thanos where this could be used, but yes, it would be possible to extend functionality this way.

saswatamcode: Thanks for explaining!

saswatamcode (approved):
LGTM!

PR #297, opened 2023-08-01
Added new bench target 'bench-fast'

### LINKED ISSUE #292: Add bench-fast make target which excludes long-running benchmarks
During local development I often rely on the benchmarks to know whether changes I make are going in the right direction.

We currently have some very slow ones so I usually comment those out, but that creates issues when switching branches and I sometimes even commit the commented out code to my branch.

It would be nice to have a `bench-fast` make target that only runs six hour dataset benchmarks so that we can get faster feedback.

Player256:
Resolves #292 
Added a bool flag ` -fast `.

### DIFF
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -5,6 +5,7 @@ package engine_test
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"strconv"
 	"strings"
@@ -101,6 +102,10 @@ func BenchmarkSingleQuery(b *testing.B) {
 }
 
 func BenchmarkRangeQuery(b *testing.B) {
+
+	sixHourDatasetOnly := flag.Bool("fast", false, "Only runs six hour dataset")
+	flag.Parse()
+
 	samplesPerHour := 60 * 2
 	sixHourDataset := setupStorage(b, 1000, 3, 6*samplesPerHour)
 	defer sixHourDataset.Close()
@@ -288,6 +293,13 @@ func BenchmarkRangeQuery(b *testing.B) {
 
 				b.ResetTimer()
 				b.ReportAllocs()
+
+				if *sixHourDatasetOnly {
+					if tc.test != sixHourDataset {
+						return
+					}
+				}
+
 				for i := 0; i < b.N; i++ {
 					qry, err := engine.NewRangeQuery(tc.test.Context(), tc.test.Queryable(), nil, tc.query, start, end, step)
 					testutil.Ok(b, err)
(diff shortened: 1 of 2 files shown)


engine/bench_test.go
```diff
@@ -101,6 +102,10 @@ func BenchmarkSingleQuery(b *testing.B) {
 }
 
 func BenchmarkRangeQuery(b *testing.B) {
```
mhoffm-aiven, line 104: can we just use `tesitng.Short()` for this?

mhoffm-aiven (commented):
Can we also do the benchstat?

Player256:
yes

mhoffm-aiven:
Does it actually work though; the big datasets are still constructed before the benchmark, right? Ill give it a go during EU afternoon.

Player256:
> Does it actually work though; the big datasets are still constructed before the benchmark, right? Ill give it a go during EU afternoon.

hey should I put if statements using testing.short() around the larger dataset.

mhoffm-aiven:
> > Does it actually work though; the big datasets are still constructed before the benchmark, right? Ill give it a go during EU afternoon.
> 
> hey should I put if statements using testing.short() around the larger dataset.

that will hide the identifier probably; but i think it can be structured nicer somewhat; We could factor out the slow benchmarks and skip those completely if we have a "Short" flag.

fpetkovski:
Agree, we should avoid creating the large datasets altogether if `-short` is provided. We could declare them but not initialize them to make compilation pass.

mhoffm-aiven:
We could add `b.Run("fast", ...)` where we do use 6h dataset and `b.Run("slow", ...)` where we use the large datasets and that we skip when the "short" flag is set. wdyt?

Player256:
> We could add `b.Run("fast", ...)` where we do construct and use 6h dataset and `b.Run("slow", ...)` where we construct and use the large datasets and that we skip when the "short" flag is set. wdyt?

I was thinking like this : 
`var largeSixHourDataset *promql.Test
	if !testing.Short() {
		largeSixHourDataset = setupStorage(b, 10000, 10, 6*samplesPerHour)
		defer largeSixHourDataset.Close()
	}

	var sevenDaysAndTwoHoursDataset *promql.Test
	if !testing.Short() {
		sevenDaysAndTwoHoursDataset = setupStorage(b, 1000, 3, (7*24+2)*samplesPerHour)
		defer sevenDaysAndTwoHoursDataset.Close()
	}`

Player256:
@mhoffm-aiven @fpetkovski any thoughts on this?

MichaHoffmann:
I would prefer if we group them under "b.Run("fast", ...)" and "b.Run("slow", ...)" and only construct the necessary datasets for the given run

Player256:
Sure👍

PR #298, opened 2023-08-16
Propagate warnings through context

fpetkovski:
The engine does not return warnings from storage which causes problems with partial response detection in Thanos.

I initially thought we had to change the interface of each operator, but @MichaHoffmann had a neat idea to propagate warnings through the context since they have no impact on flow control.


saswatamcode (approved):
LGTM!

MichaHoffmann (approved):
LGTM

PR #300, opened 2023-08-16
Propagate warnings from remote engines

fpetkovski:
This is a follow up commit to https://github.com/thanos-io/promql-engine/pull/298 which enables propagating warnings from remote engines into the distributed query context.


GiedriusS (approved):

MichaHoffmann (approved):
lgtm

PR #301, opened 2023-08-16
logicalplan: start up passthrough optimizer

GiedriusS:
Add a passthrough optimizer to the distributed engine. Currently it only passes through a query if there is only one distributed engine. My ideas for further improvements:

* Check by matching labels if there's only one matching engine
* Check by cuckoo filter on label names whether there is only one matching engine


MichaHoffmann (approved):
lgtm

PR #303, opened 2023-08-18
logicalplan: add passthrough on labels

GiedriusS:
Add passthrough optimizations on labels. This is useful if you have queriers chained and with labels you only want to query one of those queriers. This should speed up such queries a lot because it won't be needed to pass samples around.

### DIFF
--- a/logicalplan/passthrough.go
+++ b/logicalplan/passthrough.go
@@ -59,32 +59,23 @@ func (m PassthroughOptimizer) Optimize(plan parser.Expr, opts *query.Options) pa
 		return plan
 	}
 
-	// Check matchers of each selector. If all of them match only one engine
-	// then pass the query to it.
-	var matchingEngine int = -1
-	var matchingFound bool
+	matchingLabelsEngines := make([]api.RemoteEngine, 0, len(engines))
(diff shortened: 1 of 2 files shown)


logicalplan/passthrough.go
```diff
@@ -15,15 +17,78 @@ type PassthroughOptimizer struct {
...
+	var matchingEngine int = -1
+	var matchingFound bool
+	TraverseBottomUp(nil, &plan, func(parent, current *parser.Expr) (stop bool) {
+		if vs, ok := (*current).(*parser.VectorSelector); ok {
+			for i, e := range engines {
+				if !labelSetsMatch(vs.LabelMatchers, e.LabelSets()...) {
+					continue
+				}
+
+				if matchingEngine == -1 {
```
fpetkovski, line 73: To simplify the logic, maybe we can collect matching engines in a slice, and if the slice has 1 element we forward the query to the engine?

GiedriusS: Thanks for your review, I have updated the implementation as per your suggestion. I had to force push because I forgot to add the DCO :/

fpetkovski (approved):
Great improvement 💪

PR #304, opened 2023-08-19
import prometheus acceptance tests

MichaHoffmann:
* bump prometheus and import  accepance tests
* fix fallout from bumping prometheus
* fix acceptance tests ( they pass now )

distributed engine tests started failing and a histogram test started failing and were commented out for now


bwplotka (approved):
💪🏽

util/labels.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package util
```
fpetkovski, line 4: I am not a big fan of `util` packages since they tend to grow in size and end up with many unrelated functions. Maybe we can call the label `extlabels` as in labels extensions or something similar to that.

fpetkovski:
Crazy, amazing contribution.

util/labels.go
MichaHoffmann: moved the package to "extlabels" !

fpetkovski (approved):

fpetkovski:
Not a big fan of merging on weekends, but this is just adding tests and unblocks removing the vendored parser. We can make follow up PRs if there are any comments.

Issue #140, opened 2023-01-03
Make or operator "smarter"


GiedriusS:
See this https://github.com/prometheus/prometheus/issues/9624. In case of `a or b` we only want to calculate b if a doesn't fully cover everything.

MichaHoffmann:
As it turns out the engine did not implement or at all. I attempt to add support for it in #318

yeya24:
Are we able to resolve this issue now?

Issue #66, opened 2022-10-05
"Aggressive" optimizations mode


GiedriusS:
There are some optimizations that are not done in Prometheus PromQL engine due to backward compatibility. However, those use cases that they are trying not to break only really matter to a few percent of the users. Thus, I suggest adding some kind of new parameter to the engine which would add an optimization pass to the given PromQL expression. The results/behavior wouldn't be strictly compatible with the original PromQL but I think that doesn't matter to Thanos/Cortex/... users anyway.

Some candidates:
* https://github.com/prometheus/prometheus/issues/8053
* ... more?

fpetkovski:
This makes sense. I did not understand why propagating labels across bin-ops is a breaking change though. We might want to document this well once we implement it.

GiedriusS:
Because apparently there are "storages" which return data that does not strictly match the given matchers :smile: however, I think it's probably such a rare use-case that we could enable such optimizations by default.

fpetkovski:
Makes sense to have this enabled by default in Thanos at least. It shouldn't be too hard to implement, we just need to handle the vector matching in cases like:
```metric_1 / on(label_1) group_left metric_2{a="b"}```

bwplotka:
Happy to configure different optimizers in future. But it should because different. I can't see anything "too aggressive" so far 🤔

yeya24:
I guess we can close this issue as optimizers are already configurable?

Issue #5, opened 2022-09-14
Add logical plan between AST and execution


fpetkovski:
We currently translate the AST directly to a physical plan. Having an in-between logical plan will allow us to run optimizers before the query is executed.

The logical plan would have a one to one mapping with the AST and will contain the parameters of each AST node.
Query optimizers can then transform the logical plan based on predefined heuristics. One example would be optimizing series selects in binary operations so that we do as few network calls as possible.

Finally, we would build the physical plan from the optimized logical plan instead doing it from the AST directly.

![image](https://user-images.githubusercontent.com/1286231/191909951-4f4307fd-0011-4254-a4cf-73b6a6a95517.png)

yeya24:
As discussed on Slack, https://github.com/thanos-io/thanos/issues/4407 this is something we can support when having the logical plan and optimizer.

yeya24:
I will close this as we already support logical plan

PR #309, opened 2023-08-22
[Proposal] Volcano query planner

tuannh982:
### Description:
Originally I tried to separate the AST, the logical plan and physical plan from https://github.com/thanos-io/promql-engine/issues/5. But ended up doing both separation and propose a new query planner

### Related issues (might be)
- https://github.com/thanos-io/promql-engine/issues/5
- https://github.com/thanos-io/promql-engine/issues/140
- https://github.com/thanos-io/promql-engine/issues/66
### Notes:
- This is a proposal PR. It's only mean to describe how it might look like, not a full implementation. Do not merge


MichaHoffmann:
Im ++1 for seperating physical plan and logical plan! Since thats a pretty big undertaking already can we split out the Planner maybe so it gets more feasible?

bwplotka:
Nice, quite a big refactor! Do you mind starting a quick discussion on #thanos-dev Slack channel about this? We could discuss it further. We would love to learn your use cases and plan with this advanced promql engine. Theory is one thing, but in practice it's hard to estimate true cost without knowing how much data we will have to fetch. We had some initial ideas to perform two-phase query (scan, fetch) but it's not there yet.

PR #310, opened 2023-08-23
wip: sketch for count values

MichaHoffmann:
`count_values` has the issue that we need to consume the vector before we know the labels it will return which makes it hard to fit into our `model.VectorOperator` framework.

We can get around that by only executing the inner query and then computing the aggregation in the engine before returning the result. If we want to do the same for nested `count_value` calls we need to rewrite the query to treat those count_value queries as remote executions.

This is just a sketch to get a feeling for that approach.


MichaHoffmann:
Old sketch, still somewhat interesting but way out of date and not really needed anymore; we'll implement count_values as a special aggregation that materializes results in Series() soon; not super efficient, but its "count_values" so its acceptable.

PR #311, opened 2023-08-24
engine: always fall back to prometheus

MichaHoffmann:
This change makes the engine slightly easier to use in thanos.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -78,9 +78,6 @@ type Opts struct {
 	// This will default to false.
 	EnableSubqueries bool
 
-	// FallbackEngine
-	Engine v1.QueryEngine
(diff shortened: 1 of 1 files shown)


GiedriusS (changes requested):
No, we want our users to create this Prometheus engine so that it can be shared. What problem are you trying to solve?

MichaHoffmann:
> No, we want our users to create this Prometheus engine so that it can be shared. What problem are you trying to solve?

There is a bit of a circular dependency in the querier where we need a prometheus engine to construct the opts; i wanted to pass some engine.Opts to the constructor already. The alternative is to pass engine instances to engineFactory already adn not just opts.

PR #315, opened 2023-09-04
Expose functions for matrix selector

fpetkovski:
We are using the engine to query a different type of storage and I would like to expose range functions so that they can be imported from outside the project.


yeya24 (approved):
LGTM

PR #319, opened 2023-10-03
misc: bump prometheus

MichaHoffmann:
* bump prometheus to latest main
* fix test fallout by repairing step_invariant to deal with histograms better
* used `comparer` for all comparisons between old and new results
* dropped promql annotations in `comparer` for now

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -7,18 +7,19 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"os"
 	"reflect"
 	"runtime"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/efficientgo/core/errors"
 	"github.com/efficientgo/core/testutil"
-	"github.com/prometheus/prometheus/promql/parser"
+	"go.uber.org/goleak"
+	"golang.org/x/exp/maps"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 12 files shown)


execution/step_invariant/step_invariant.go
```diff
@@ -142,7 +142,7 @@ func (u *stepInvariantOperator) cacheInputVector(ctx context.Context) error {
 		}
 		defer u.next.GetPool().PutVectors(in)
 
-		if len(in) == 0 || len(in[0].Samples) == 0 {
+		if len(in) == 0 || (len(in[0].Samples) == 0 && len(in[0].Histograms) == 0) {
```
MichaHoffmann, line 145: A bit of a bundled bugfix here; we can do that in its own PR i think too; wdyt?

engine/engine_test.go
```diff
@@ -4911,6 +4911,16 @@ func roundValues(r *promql.Result) {
 	}
 }
 
+// dropAnnotations drops anntoations from the result
+// TODO: this should be done in a comparer and is only
+// because prometheus got more liberal with its usage of
+// annotations for info purposes. It should only be used
+// if we use a mock storage that doesnt send the same info
+// annotations.
+func dropAnnotations(r *promql.Result) {
```
MichaHoffmann, line 4920: We should try to mirror the annotations that PromQL adds at some point but this is a larger undertaking so ill just drop them for comparisons sake now. 
It is not optimal and we should follow-up with a PR that does this smarter.

yeya24 (approved):
Thanks LGTM.

execution/step_invariant/step_invariant.go
yeya24: I am ok to fix it this pr

Issue #312, opened 2023-08-24
grouping and binary operator results differ between thanos vs prometheus engine


v1jayr:
Query: 

```
sum by (cluster, namespace, pod, container) (irate(container_cpu_usage_seconds_total{image!="",job="kubelet",metrics_path="/metrics/cadvisor", container="alertmanager"}[5m])) * on (cluster, namespace, pod) group_left (node) topk by (cluster, namespace, pod) (1, max by (cluster, namespace, pod, node) (kube_pod_info{node!="", pod=~"alertmanager-.*"}))
```

Results using thanos engine:
<img width="2556" alt="Screenshot 2023-08-24 at 12 29 40 PM" src="https://github.com/thanos-io/promql-engine/assets/7508769/7df97a02-f7ed-4df5-bdf4-ea3bd3c8bb27">

Results using prometheus engine:
<img width="2546" alt="Screenshot 2023-08-24 at 12 30 26 PM" src="https://github.com/thanos-io/promql-engine/assets/7508769/eb55f50b-3235-4015-9a5f-5bfce8ce9333">

As you can see the left and right side of the query is the same before and after but how the engine apply grouping results in different results.

GiedriusS:
Given that UI is missing some elements, I'm guessing that you are running an old Thanos version. Care updating to the newest one and testing again?

GiedriusS:
Adding fix here: https://github.com/thanos-io/promql-engine/pull/320. Will close this issue once the newer promql-engine is in Thanos.

PR #320, opened 2023-10-12
execution/filter: allow multiple matchers with same name

GiedriusS:
It is entirely valid to have multiple matchers with the same name e.g. `phase!="Running", phase!="Succeeded`.

See https://github.com/thanos-io/promql-engine/issues/312.


niaurys (approved):

saswatamcode (approved):
LGTM!

PR #322, opened 2023-10-25
Emit error when querying native histograms with extended functions

pedro-stanaka:
We saw panics on our setup because users tried to use the extended functions with native histograms, which led to rates calculated with only one sample in some windows.

In this PR I am introducing a safeguard against this type of situation and emiting an error when users try to query native histograms with extended functions.

### DIFF
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -172,7 +172,7 @@ func (o *matrixSelector) Next(ctx context.Context) ([]model.StepVector, error) {
 			if !o.isExtFunction {
 				rangeSamples, err = selectPoints(series.samples, mint, maxt, o.scanners[i].previousSamples)
 			} else {
-				rangeSamples, err = selectExtPoints(ctx, series.samples, mint, maxt, o.scanners[i].previousSamples, o.extLookbackDelta, &o.scanners[i].metricAppearedTs)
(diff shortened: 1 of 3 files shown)


fpetkovski (approved):

execution/scan/matrix_selector.go
```diff
@@ -351,8 +353,9 @@ loop:
 // into the [mint, maxt] range are retained; only points with later timestamps
 // are populated from the iterator.
 // TODO(fpetkovski): Add max samples limit.
-func selectExtPoints(it *storage.BufferedSeriesIterator, mint, maxt int64, out []Sample, extLookbackDelta int64, metricAppearedTs **int64) ([]Sample, error) {
+func selectExtPoints(ctx context.Context, it *storage.BufferedSeriesIterator, mint, maxt int64, out []Sample, extLookbackDelta int64, metricAppearedTs **int64) ([]Sample, error) {
```
GiedriusS, line 356: Why this context was added here? Seems like an unrelated change

pedro-stanaka: Good catch, I was using the context before to attach warns, but we saw it was better to return an error altogether.

pedro-stanaka:
Seems like we got a flaky test, `make test` on my local succeeds, but not on CI. Trying an empty commit to see if succeeds.

GiedriusS (approved):

PR #323, opened 2023-11-03
Optimize aggregate accumulators

fpetkovski:
Scalar tables used in the hash aggregate with grouping labels are currently very wasteful because of the dynamic functions they create for each output group. Each function ends up as a new object on the heap, so for queries where output groups have high cardinality we end up using a lot of memory.

This commit fixes that by converting accumulators to structs, so that functions between accumulator instances do not end up as heap-allocated objects.

Here is an extreme case where accumulators took 12GB of heap space:
image

### DIFF
--- a/execution/aggregate/accumulator.go
+++ b/execution/aggregate/accumulator.go
@@ -1,0 +1,258 @@
+package aggregate
+
+import (
+	"math"
+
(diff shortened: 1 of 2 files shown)


MichaHoffmann (approved):
lgtm

yeya24 (approved):
Thanks! LGTM

PR #324, opened 2023-11-06
Add batched streaming aggregations

fpetkovski:
With the current model we expect each Next call to return samples for unique steps. This approach works well because of its simplicity, but for high cardinality queries (100K+ series), it tends to use a lot of memory because the buffers for each step tend to be big.

This commit resolves that by allowing the aggregate to handle batches from the same step coming from subsequent Next calls. Selectors are expanded with a batchSize parameter which can be injected when a streaming aggregate is present in the plan. Using this parameter then can put an upper limit on the size of the output vectors they produce.

This is a before and after of the total heap size of all queriers from a 1M series query. The green line indicates the total heap size (`sum(go_memstats_heap_inuse_bytes)`) for all queriers in the query path when executing the query with this change. The yellow line is the total memory used by all queriers in the query path using the main branch of the engine.

There is approximately a 20% reduction in heap size because vector batches from the vector selector are capped to 32K instead of being unbounded as they are on main.

image

### DIFF
--- a/engine/bench_test.go
+++ b/engine/bench_test.go
@@ -6,6 +6,7 @@ package engine_test
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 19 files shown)


yeya24:
@fpetkovski Can you please add more description to the graph? What does the yellow and green line stand for?

fpetkovski:
Hey @yeya24 I've added a brief description, let me know if that helps. I can't really post the legend because it has some internal details.

MichaHoffmann (approved):
lgtm

yeya24 (approved):
LGTM

PR #327, opened 2023-11-19
Bench-fast 

bazooka3000:
Modified the code to build smaller dataset for "make bench-fast" target.


MichaHoffmann:
This looks pretty good to me!

GiedriusS (commented):
I'm not sure why this complexity is needed - why not just skip over the long running benchmarks if `https://pkg.go.dev/testing#Short` is set?

Issue #326, opened 2023-11-19
Fuzzy query test failed on latest main


yeya24:
Fuzzy query test failed on latest main after merging https://github.com/thanos-io/promql-engine/pull/324.

GH action log: https://github.com/thanos-io/promql-engine/actions/runs/6919173003/job/18822211713

```
    enginefuzz_test.go:128: load 30s
        			http_requests_total{pod="nginx-1", route="/"} 1.00+1.00x4
        			http_requests_total{pod="nginx-2", route="/"}  1+2.00x4
    enginefuzz_test.go:129: group by (pod, route) (atanh(-{__name__="http_requests_total"} offset -3m4s))
    enginefuzz_test.go:131: case 43 error mismatch.
        new result: 
        old result: {pod="nginx-1", route="/"} => 1 @[0]
        {pod="nginx-2", route="/"} => 1 @[0]
    enginefuzz_test.go:136: failed 1 test cases
--- FAIL: FuzzDistributedEnginePromQLSmithInstantQuery (1.89s)
    --- FAIL: FuzzDistributedEnginePromQLSmithInstantQuery/seed#0 (1.89s)
FAIL
FAIL	github.com/thanos-io/promql-engine/engine	416.641s
```

Haven't got the chance to take a closer look. If this is a false positive then we can close.

fpetkovski:
Hm this could be related to my change because, I will take a look tomorrow morning.

PR #328, opened 2023-11-20
Fix group aggregate with NaN values

fpetkovski:
The group aggregate should produce an output value even when inputs are NaN.

Fixes https://github.com/thanos-io/promql-engine/issues/326.


MichaHoffmann (approved):
lgtm

PR #330, opened 2023-11-20
plan: fix traversal for step invariant expr

MichaHoffmann:
Consider 
```
sum_over_time(sum by (series) (http_requests_total @ 10.000)[5m:1m])
```

The inner aggregation is step invariant and we never traverse into it.


fpetkovski (approved):

yeya24 (approved):
Thanks!

PR #331, opened 2023-11-23
Distribute avg aggregation

fpetkovski:
We currently do not distribute the average aggregation because would be incorrect to calculate an average over averages. However, we could rewrite it as sum / count and distribute those two aggregations independently.

### DIFF
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -129,6 +129,7 @@ var distributiveAggregations = map[parser.ItemType]struct{}{
 	parser.SUM:     {},
 	parser.MIN:     {},
 	parser.MAX:     {},
+	parser.AVG:     {},
 	parser.GROUP:   {},
 	parser.COUNT:   {},
 	parser.BOTTOMK: {},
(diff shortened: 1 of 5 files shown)


GiedriusS (approved):
Taking a page out of promxy's book? :smile:

GiedriusS (approved):

logicalplan/distribute.go
```diff
@@ -129,6 +129,7 @@ var distributiveAggregations = map[parser.ItemType]struct{}{
 	parser.SUM:     {},
 	parser.MIN:     {},
 	parser.MAX:     {},
+	parser.AVG:     {},
```
MichaHoffmann, line 132: I think this might be harmless, but technically its not true that AVG is distributive right? We work around it by rewriting it to sum/count in a different optimizer so that we never have to see avg in this optimizer if i understand correctly. Adding this here seems like it could be a subtle bug waiting to happen

fpetkovski: I think you are right! 

I will try to fix this next week, since we have a separate optimizer that rewrites `avg`, we should remove it from this list.

PR #332, opened 2023-11-25
plan, engine: get rid of trim sorts optimizer

MichaHoffmann:
This is not really an optimizer in a way since it changes the meaning of the query. Its more an engine feature that we can do this since the engine itself does the sorting when presenting results. Every optimizer can expect that; thats why it makes sense just to get rid of sorts before even doing optimizations.


fpetkovski (approved):
Nice 👍

logicalplan/plan.go
```diff
@@ -46,6 +46,9 @@ func New(expr parser.Expr, opts *query.Options) Plan {
 	setOffsetForAtModifier(opts.Start.UnixMilli(), expr)
 	setOffsetForInnerSubqueries(expr, opts)
 
+	// the engine handles sorting at the presentation layer
+	expr = trimSorts(expr)
```
yeya24, line 50: How do we support sort_desc right now?
The presentation layer needs to know it needs to sort in reverse order?

yeya24: And I wonder the behavior of topk and bottomk together with this change

MichaHoffmann: Sorting is done here https://github.com/thanos-io/promql-engine/blob/7a183df0925bee2179aa351b235ed090b82dca9e/engine/engine.go#L475. Care needs to be taken to compute the sorting order before this step in the plan but if that is done then sorting functions are irrelevant for the remaining computation.

yeya24: Thanks. Then this makes sense

yeya24 (approved):

PR #333, opened 2023-11-25
misc cleanups

MichaHoffmann:
* spread related functionality in seperate files
* get rid of debug writer, rather print output when testing with "-v -explain" by using a debugEngine
* use comparer in tests that were not using it before
* move pool construction into operator constructors to make execute.go easier and not pass opts twice
* use opts.IsInstantQuery to decide if we have an instant query
* minor other cleanups


fpetkovski (commented):
Thanks for the cleanup. Unfortunately this also hides the debug option and makes it hard to use is outside of the project. The refactoring makes sense, but I would prefer it if we kept some flag in the engine options.

MichaHoffmann:
Its not hidden, you can test with "-explain" and opt into it!

PR #334, opened 2023-11-25
engine: add annotation if we trigger promql fallback

MichaHoffmann:
I just thought that this might give users a nice feedback if their query triggers fallback in the thanos engine, which might be undesired but hard to detect.


fpetkovski:
This behavior might be a bit strange for users who are not familiar with the internals. Perhaps it could be an optional behavior?

PR #336, opened 2023-11-26
execution: fix label handling in histogram_quantile

MichaHoffmann:


MichaHoffmann:
the vector aggregator String() change is a bit sneaky; if needed i can take it out.

engine/engine_test.go
```diff
@@ -2980,6 +2980,13 @@ func TestInstantQuery(t *testing.T) {
 			queryTime: time.Unix(0, 0),
 			query:     "sort_desc(http_requests_total)",
 		},
+		{
+			name: "histogram_quantile with mock duplicate labels",
```
fpetkovski, line 2984: What does mock duplicate labels means in this context?

fpetkovski (approved):

engine/engine_test.go
MichaHoffmann: "-X or X" will have name labels dropped from the left and not from the right but its kinda mock because it never can return both. It just illustrates the limitation of calculating all possible series beforehand; thats why we need to calculate duplicate labels in the engine and not in the operators.

fpetkovski (approved):

PR #337, opened 2023-11-27
Fix distributing parenthesis

fpetkovski:
Queries with parenthesis are not being distributed right now because we never run `transform` on the inner node.

This commit fixes that and adds tests to verify behavior.


MichaHoffmann (approved):
lgtm

PR #340, opened 2023-11-28
Prune remote engines by time

fpetkovski:
The distributed optimizer generates one query per remote engine without taking into account the range of the engine. This reduces resiliency since an engine with an older timestamp could be unavailable, but it still gets queried for recent data.

This commit adds time-based pruning to the optimizer which excludes endpoints which do not match the query range.


MichaHoffmann (approved):
lgtm

PR #341, opened 2023-11-29
execution: fix min/max aggregate

MichaHoffmann:
NaN handling in min/max is sufficiently special to warrant having their own operators. To summarize; NaN will be set if the aggregator has no value yet but ignored otherwise. If the value of the aggregator is NaN we will also always override it despite for example "x < NaN" evaluating to false.


fpetkovski (approved):
Thanks for the fix, hopefully this is the last of those issues.

PR #345, opened 2023-12-05
plan: fix distributed avg

MichaHoffmann:
The we should rewrite to sum by (xxx) / on (xxx) count by (xxx).

### DIFF
--- a/logicalplan/distribute_avg.go
+++ b/logicalplan/distribute_avg.go
@@ -27,10 +27,14 @@ func (r DistributeAvgOptimizer) Optimize(plan parser.Expr, _ *query.Options) par
 			count := *(*current).(*parser.AggregateExpr)
 			count.Op = parser.COUNT
 			*current = &parser.BinaryExpr{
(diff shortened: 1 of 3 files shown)


logicalplan/distribute_avg.go
```diff
@@ -27,10 +27,14 @@ func (r DistributeAvgOptimizer) Optimize(plan parser.Expr, _ *query.Options) par
...
 			*current = &parser.BinaryExpr{
-				Op:             parser.DIV,
-				LHS:            &sum,
-				RHS:            &count,
-				VectorMatching: &parser.VectorMatching{},
+				Op:  parser.DIV,
+				LHS: &sum,
+				RHS: &count,
+				VectorMatching: &parser.VectorMatching{
+					Include:        aggr.Grouping,
```
MichaHoffmann, line 34: We need to include that to make sure that __name__ is not dropped ( its usually dropped by vector binary "/" )

PR #346, opened 2023-12-06
engine, plan: distribute binary vector expressions if appropriate

MichaHoffmann:
If the signature for the binary matching contains the external labels then all matches must be inside the remote engine which means we can push them down.

Credit for the idea has to go to @fpetkovski

### DIFF
--- a/engine/distributed_test.go
+++ b/engine/distributed_test.go
@@ -220,6 +220,8 @@ func TestDistributedAggregations(t *testing.T) {
 		{name: "double aggregation", query: `max by (pod) (sum by (pod) (bar))`},
 		{name: "aggregation with function operand", query: `sum by (pod) (rate(bar[1m]))`},
 		{name: "binary expression with constant operand", query: `sum by (region) (bar * 60)`},
+		{name: "binary expression with distributable pairing", query: `sum by (pod) (bar * bar)`},
+		{name: "binary expression with non-distributable pairing", query: `sum by (pod) (foo * ignoring (region, zone) xyz)`},
 		{name: "binary aggregation", query: `sum by (region) (bar) / sum by (pod) (bar)`},
 		{name: "binary nested with constants", query: `(1 + 2) + (1 atan2 (-1 % -1))`},
 		{name: "binary nested with functions", query: `(1 + exp(vector(1))) + (1 atan2 (-1 % -1))`},
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -399,13 +399,7 @@ func isDistributive(expr *parser.Expr) bool {
 
 	switch aggr := (*expr).(type) {
 	case *parser.BinaryExpr:
-		// Binary expressions are joins and need to be done across the entire
-		// data set. This is why we cannot push down aggregations where
-		// the operand is a binary expression.
-		// The only exception currently is pushing down binary expressions with a constant operand.
-		lhsConstant := isConstantExpr(aggr.LHS)
-		rhsConstant := isConstantExpr(aggr.RHS)
-		return (lhsConstant || rhsConstant)
+		return isBinaryExpressionWithOneConstantSide(aggr) || isBinaryExpressionWithDistributableMatching(aggr)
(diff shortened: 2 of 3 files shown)


SuperQ (approved):
This would be very nice as we enforce at least one external label on all of our queries.

engine/distributed_test.go
```diff
@@ -218,6 +218,8 @@ func TestDistributedAggregations(t *testing.T) {
 		{name: "double aggregation", query: `max by (pod) (sum by (pod) (bar))`},
 		{name: "aggregation with function operand", query: `sum by (pod) (rate(bar[1m]))`},
 		{name: "binary expression with constant operand", query: `sum by (region) (bar * 60)`},
+		{name: "binary expression with distributable pairing", query: `sum by (pod) (bar * bar)`},
+		{name: "binary expression with non-distributable pairing", query: `sum by (pod) (bar * ignoring (region) bar)`},
```
fpetkovski, line 222: As a sanity check, could you make [this](https://github.com/thanos-io/promql-engine/pull/346/files#diff-6a4b9ef5cc84db6638cc2cf5bce5e1345aedbf74145971341be9df1ca75949c4R421) function return `true` locally and rerun the test? I want to make sure the test case fails if we distribute the binary operation.

MichaHoffmann: Added a test that fails if i mistakenly allow distribution and succeeds if i correct it again: 

```
$ go test . -run TestDistributedAggregations/.*/binary_expression_with_non-distributable_pairing/.* -failfast -test.count=1
--- FAIL: TestDistributedAggregations (0.28s)
    --- FAIL: TestDistributedAggregations/count_by___name___label#02 (0.00s)
        --- FAIL: TestDistributedAggregations/count_by___name___label#02/binary_expression_with_non-distributable_pairing (0.00s)
            --- FAIL: TestDistributedAggregations/count_by___name___label#02/binary_expression_with_non-distributable_pairing/instant/ts=121 (0.00s)
                distributed_test.go:288: Query: sum by (pod) (foo * ignoring (region, zone) xyz)
                    Explanation:
                    [*concurrencyOperator(buff=2)]:
                    └──[*aggregate] sum by ([pod]):
                       └──[*concurrencyOperator(buff=2)]:
                          └──[*dedup]:
                             └──[*coalesce]:
                                ├──[*concurrencyOperator(buff=2)]:
                                │  └──[*remoteExec] sum by (pod, zone) (foo * ignoring (region, zone) xyz) (121, 121)
                                └──[*concurrencyOperator(buff=2)]:
                                   └──[*remoteExec] sum by (pod, zone) (foo * ignoring (region, zone) xyz) (121, 121)


                distributed_test.go:289:
                distributed_test.go:290: {pod="nginx-1"} => 36 @[121000]
                testutil.go:91: distributed_test.go:291: ""

                    	exp: &promql.Result{Err:error(nil), Value:promql.Vector{promql.Sample{T:121000, F:36, H:(*histogram.FloatHistogram)(nil), Metric:labels.Labels{labels.Label{Name:"pod", Value:"nginx-1"}}}}, Warnings:annotations.Annotations{}}

                    	got: &promql.Result{Err:error(nil), Value:promql.Vector{}, Warnings:annotations.Annotations(nil)}

                    Diff:
                    --- Expected
                    +++ Actual
                    @@ -1,2 +1,2 @@
                    -(*promql.Result)({pod="nginx-1"} => 36 @[121000])
                    +(*promql.Result)()



FAIL
FAIL	github.com/thanos-io/promql-engine/engine	0.308s
FAIL
 fedora  ~  git  promql-engine  engine   mhoffm-distribute-binary-expressions-where-signature-contains-all-labels   
$ go test . -run TestDistributedAggregations/.*/binary_expression_with_non-distributable_pairing/.* -failfast -test.count=1
ok  	github.com/thanos-io/promql-engine/engine	0.430s
```

Since in the "count by __name__" scenario xyz and bar inhabit different remote engines; if i try to distribute it i get an empty result but if i dont it passes again.

PR #349, opened 2023-12-07
Prune engines for absent and absent_over_time

fpetkovski:
Absent and absent_over_time are distributed differently and the pruning logic was not applied for these functions.

This commit implements time-based pruning and adds test cases to illustrate when it happens.


MichaHoffmann (approved):
lgtm

PR #350, opened 2023-12-07
Add support for absent_over_time

fpetkovski:
This commit adds support for the absent_over_time function. It is implemented through existing operators by wrapping last_over_time with absent.


MichaHoffmann (approved):
lgtm

PR #351, opened 2023-12-08
engine: make fuzzing repeatable

MichaHoffmann:
This change makes fuzzing ( local engine ) repeatable meaning we can check in fuzzing crashes to make sure they stay fixed without needing to convert every one into a unittest.


yeya24 (approved):
Thanks! Great work

engine/testdata/fuzz/FuzzEnginePromQLSmithInstantQuery/010a0568eb7e1f6d
```diff
@@ -0,0 +1,7 @@
+go test fuzz v1
```
fpetkovski, line 1: Is this supposed to be commited?

MichaHoffmann: Yeah ( i mean arguably no but its a known fuzzing failure [ due to a defect i introduced to test fuzzing ] ); those files should eventually function as a regression suite that is built up through fuzzing, i think they are supposed to be committed.

fpetkovski (approved):

PR #353, opened 2023-12-09
engine: comment out long benchmarks

MichaHoffmann:
This is only temporary so we can run regular benchmarks again. If we cannot run the long running ones locally or on CI then they do more harm then good.


fpetkovski (approved):
We can get these out of the way until we figure out what to do with them.

PR #357, opened 2023-12-12
Distribute label_replace

fpetkovski:
The label_replace function is not distributed because we don't handle string literals in the traversal function.

This commit fixes that.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -18,6 +18,7 @@ import (
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/util/annotations"
 	"github.com/prometheus/prometheus/util/stats"
 	v1 "github.com/prometheus/prometheus/web/api/v1"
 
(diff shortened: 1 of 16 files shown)


MichaHoffmann:
So I wonder what happens if you label replace an engine label away and execute a sum aggregation; is that problematic; can we add a test for it?

fpetkovski:
This is a good catch. For now let's not distribute that case at all and emit a warning when a user runs this type of a query.

engine/engine.go
```diff
@@ -330,7 +334,8 @@ func (q *compatibilityQuery) Exec(ctx context.Context) (ret *promql.Result) {
 		return &promql.Result{Value: promql.String{V: e.Val, T: q.ts.UnixMilli()}}
 	}
 	ret = &promql.Result{
-		Value: promql.Vector{},
+		Value:    promql.Vector{},
+		Warnings: q.warns,
```
MichaHoffmann, line 338: I think we need to merge here with the warnings from the context maybe?

fpetkovski: Done

MichaHoffmann (approved):
lgtm

PR #358, opened 2023-12-12
Make binary pushdown optional

fpetkovski:
Pushing down binary operations with no vector matching makes sense for Thanos and we should do have it enabled by default. However, we use the engine to query other metric sources and here we see issues where matching series between two metrics are not present in the same engine.

To provide an escape hatch while we figure out a more generic solution, this commit makes the binary pushdown optional and enabled by default, but can be disabled by setting a flag.


MichaHoffmann (approved):
lgtm

PR #359, opened 2023-12-12
Do not distribute aggregations over absent

fpetkovski:
Absent should be treated as an aggregation, and we should not distribute another aggregation on top of it. It is not explicitly defined as one in PromQL and it falls under functions, but its behavior is more of an aggregation.


MichaHoffmann (approved):
lgtm

PR #361, opened 2023-12-13
execution: add timestamp function

MichaHoffmann:
This PR implements "timestamp" by pushing it down into the vector scanner.
It is communicated via selector hints.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2672,6 +2672,36 @@ func TestInstantQuery(t *testing.T) {
 		queryTime time.Time
 	}{
 		{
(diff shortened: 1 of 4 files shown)


fpetkovski (commented):
Really nice work 👍 Could we add some tests? :)

MichaHoffmann:
> Really nice work 👍 Could we add some tests? :)

Ill add some! I did omit them since during implementation the acceptance tests from promql upstream failed and i assumed its already covered but having them explicitly is much better of course!

fpetkovski (approved):

PR #363, opened 2023-12-14
Update prometheus dependency

fpetkovski:
This commit updates prometheus to d0c2d9c0b9cc0f79582ebfe0f1439e3a8d01fc78 which contains changes for using a loser tree when merging postings.

This should allow us to update Thanos to the same Prometheus version so that we can pick up updates from both Prometheus and the engine.


MichaHoffmann (approved):
lgtm, well fix linter errors offline

PR #367, opened 2023-12-19
plan: unwrap paren expressions

MichaHoffmann:


logicalplan/plan.go
```diff
@@ -172,6 +173,20 @@ func trimSorts(expr parser.Expr) parser.Expr {
 	return expr
 }
 
+func trimParens(expr parser.Expr) parser.Expr {
+	TraverseBottomUp(nil, &expr, func(parent, current *parser.Expr) bool {
```
fpetkovski, line 177: I think that we can use the prometheus function here because at this point we don't have any custom nodes.

fpetkovski (approved):

PR #368, opened 2023-12-20
Do not distribute subqueries

fpetkovski:
Distributing subqueries currently yields incorrect results due multiple range windows. For now we can disable them to avoid wrong results.


MichaHoffmann (approved):
lgtm

PR #369, opened 2023-12-20
execution: add forgotten subquery guards

MichaHoffmann:


fpetkovski (approved):

yeya24:
Is subquery in a range query not supported so far?

MichaHoffmann:
> Is subquery in a range query not supported so far?

No we only did instant queries so far

PR #370, opened 2023-12-21
Distribute subqueries

fpetkovski:
Distributing subqueries can be done in the same manner like we can distribute regular matrix selectors. The only difference is that we need to make sure a remote engine has sufficient scope to calculate both the inner and outer query range. To do this, we simply add the two ranges together when calculating the offset for the remote query and based on this we can decide when it is safe to distribute.


MichaHoffmann (approved):
lgtm

PR #373, opened 2023-12-23
engine,execution: use logical matrix and vector

MichaHoffmann:
As the last phase of creating a new plan we replace all Matrix and Vector selectors by our logical ones. This makes it possible to not handle them twice everywhere in the codebase. It will also facilitate more general sharding of them later on since we can annotate them directly with their sharding information.
We had to do matrix and vector selector in the same PR since the distributed engine cannot handle logical nodes well (because they generally change the String() representation of the node). We also had to change the String() representation of the logical nodes to be equivalent to the original PromQL expression they came from.


MichaHoffmann:
Changing "String()" for VectorSelector breaks some "plan" tests that use it for asserting.

fpetkovski (approved):
Really great change 👍

PR #375, opened 2024-01-01
Add suport for range subqueries

fpetkovski:
This commit extends the implementation of subqueries to also cover
range queries. It does that by introducing a generic, timestamp-aware ringbuffer
which can be used to add and drop samples within a certain window.
    
For now the ringbuffer is only used for subqueries, but the idea is to
also use it in the matrix selector to simplify the windowing logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,10 @@ The following table shows operations which are currently supported by the engine
 |------------------------|------------------------------------------------------------------------------------------------|----------|
 | Binary expressions     | Full support                                                                                   |          |
 | Histograms             | Full support                                                                                   |          |
+| Subqueries             | Full support                                                                                   |          |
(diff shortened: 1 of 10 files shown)


mhoffm-aiven (approved):
Amazing Work, LGTM

mhoffm-aiven (approved):

MichaHoffmann (approved):

fpetkovski:
Posting benchmark results for reference
```
CPU
RangeQuery/subquery-8       81.9ms ± 0%    37.7ms ± 1%   -53.91%  (p=0.008 n=5+5)

Mem
RangeQuery/subquery-8       24.2MB ± 0%    31.5MB ± 0%   +30.49%  (p=0.008 n=5+5)`
```

PR #378, opened 2024-01-02
engine,execution: generally enable subqueries

MichaHoffmann:


execution/scan/subquery.go
```diff
@@ -116,10 +116,10 @@ func (o *subqueryOperator) Next(ctx context.Context) ([]model.StepVector, error)
 			for j, vector := range vectors {
 				if vector.T > maxt {
 					o.lastVectors = vectors
+					o.lastCollected = j - 1
```
MichaHoffmann, line 119: if first vector has larger timestamp and we had no overflow before then lastCollected would be length of previous vectors and new vectors might be shorter resulting in a panic; if we set it like this its always set properly.

execution/execution.go
```diff
@@ -267,7 +261,10 @@ func newSubqueryFunction(e *parser.Call, t *parser.SubqueryExpr, storage *engsto
 	if parse.IsExtFunction(e.Func.Name) {
 		return nil, parse.ErrNotImplemented
 	}
-	// TODO: only instant queries for now.
+	// TODO: We dont pass arguments yet
+	if e.Func.Name == "quantile_over_time" {
```
MichaHoffmann, line 265: we dont yet pass arguments in subqueries so if we invoke quantile_over_time we get a panic since the scalar arguments of the function were not populated

execution/scan/subquery.go
MichaHoffmann: first step vector of the current batch i mean

fpetkovski (approved):

PR #379, opened 2024-01-03
Fix fuzz queryables

yeya24:
Originally found by @MichaHoffmann, the same storage got appended twice to queryables.

Also try to re-enable distributed engine tests as well as upgrading Prometheus to main as the commit in @fpetkovski has been removed.


MichaHoffmann (approved):
Lgtm

PR #381, opened 2024-01-03
Fuzz Test: change load to a string slice

yeya24:
change load to a string slice in order to store multiple initial load string. Useful for distributed query case when there are multiple store to load initial data.


MichaHoffmann (approved):
lgtm

PR #386, opened 2024-01-05
Set correct query opts for distributed subqueries

fpetkovski:
When distributing aggregations which are part of a subquery, we need to use subquery opts rather than the root query opts.

The approach is not completely optimal since we need to reassemble the parent path and jump through a bunch of pointers. As a follow up we can modify the `TraverseBottomUp` function to pass in the entire parent path instead of only the immediate parent.

### DIFF
--- a/execution/exchange/coalesce.go
+++ b/execution/exchange/coalesce.go
@@ -50,9 +50,6 @@ type coalesce struct {
 }
 
 func NewCoalesce(pool *model.VectorPool, opts *query.Options, batchSize int64, operators ...model.VectorOperator) model.VectorOperator {
-	if len(operators) == 1 {
-		return operators[0]
-	}
 	c := &coalesce{
 		pool:          pool,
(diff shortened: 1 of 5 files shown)


execution/exchange/coalesce.go
```diff
@@ -50,6 +50,9 @@ type coalesce struct {
 }
 
 func NewCoalesce(pool *model.VectorPool, opts *query.Options, batchSize int64, operators ...model.VectorOperator) model.VectorOperator {
+	if len(operators) == 1 {
```
mhoffm-aiven, line 53: residue from debugging

fpetkovski: Yeah good catch. But now that I think about it, would it be a good optimization to have?

mhoffm-aiven: oh yeah generally yes, but we can do that with the late coalesce PR i think

MichaHoffmann (approved):
lgtm

PR #388, opened 2024-01-06
Tidy up analysis

fpetkovski:
Telemetry is not consistently enabled or disabled in all operators and there is a bit of duplication between the Explain and Analyze methods on each operator.

This commit tidies up creation of telemetry trackers and removes the Analyze method on individual operators. Instead, the Explain method is used to traverse the operator tree and each child is checked for telemetry information. The advantage of this method is that Explain is enforced by the operator interface and has to be implemented, whereas Analyze is an optional method. If one operator does not implement Analyze, the entire traversal would stop, which is not the case with Explain.


MichaHoffmann (approved):
lgtm

MichaHoffmann:
CC @GiedriusS @saswatamcode; I was a bit trigger happy on the merge; do you guys have any concerns or comments here?

PR #389, opened 2024-01-06
execution: misc fuzzing fixes

MichaHoffmann:
* use make to orchestrate fuzzing
* more robust label computation in absent
* minor cleanups


engine/testdata/fuzz/FuzzEnginePromQLSmithInstantQuery/61a4f8d0ac706041
```diff
@@ -0,0 +1,7 @@
+go test fuzz v1
+int64(166)
+uint32(7)
+float64(60)
+float64(8)
+float64(11)
+float64(2)
```
yeya24, line 7: What's the expectation and points for us to check in those fuzz files?
Do we need to check them regularly and see if those cases are fixed?

MichaHoffmann: The serve as regression checks for previous failures

fpetkovski: Are they used as seed into the fuzzer?

MichaHoffmann: Yes they serve as seeds ( which also guarantees that we regularly run the same things that failed as part of regression checks ). I think those seeds are also run as part of unittest suite

yeya24: @MichaHoffmann How those test cases got included in the unit tests? It is part of `make test`?

MichaHoffmann: `go test .` detects and runs fuzzing seed corpus tests afaik

yeya24 (approved):

fpetkovski (approved):

PR #391, opened 2024-01-07
Adding filter to vector scanner

### LINKED ISSUE #338: Push boolean selectors into scanners
Certain PromQL expression can be considered as (sample) selectors and can be evaluated directly in scanners. 

One example is:
```
sum(kube_pod_info == 1)
```

For this query, we could apply `==1` directly into the vector_selector (maybe we should rename this to vector_scanner) and avoid emitting samples which do not match that binary operation.

harsh-ps-2003:
Fixes #338 

**Changes**
Adding a filtering function to the vector scanner which would take in a step vector and produce a new step vector with filtered samples to optimize the query execution process. 

**Verification**
Added a test case `sum(kube_pod_info == 1)`

Tests are passing locally using `make test`

### DIFF
--- a/execution/scan/vector_selector.go
+++ b/execution/scan/vector_selector.go
@@ -23,9 +23,10 @@ import (
 )
 
 type vectorScanner struct {
-	labels    labels.Labels
-	signature uint64
-	samples   *storage.MemoizedSeriesIterator
+	labels     labels.Labels
+	signature  uint64
+	samples    *storage.MemoizedSeriesIterator
+	filterFunc func(model.StepVector) model.StepVector
 }
 
 type vectorSelector struct {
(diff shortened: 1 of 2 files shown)


harsh-ps-2003:
@fpetkovski the tests are failing :
```
panic: test timed out after 10m0s
running tests:
        TestNativeHistograms (3m59s)
        TestNativeHistograms/integer_histograms (3m59s)
        TestNativeHistograms/integer_histograms/absent (4s)
        TestNativeHistograms/integer_histograms/absent/mixedTypes=false (4s)
        TestNativeHistograms/integer_histograms/absent/mixedTypes=false/instant (0s)
```

Still in the process of figuring out what I am messing up!

execution/scan/vector_selector.go
```diff
@@ -23,9 +23,10 @@ import (
 )
 
 type vectorScanner struct {
-	labels    labels.Labels
-	signature uint64
-	samples   *storage.MemoizedSeriesIterator
+	labels     labels.Labels
+	signature  uint64
+	samples    *storage.MemoizedSeriesIterator
+	filterFunc func(model.StepVector) model.StepVector
```
MichaHoffmann, line 29: I dont think this is ever populated anywhere right?

MichaHoffmann:
I think we need to add a "FilterFunc" to `logicalplan.VectorSelector` and then pass it down from `execute.go` into the constructor of the `scan.VectorScan` operator. The logical filter should be populated by an optimizer that also removes the binary expression (which is the ultimate benefit ~ no binary operator on top of the vector scan anymore).  I think that would be the way to implement the feature.

execution/scan/vector_selector.go
harsh-ps-2003: There might be some gaps in my understanding (sorry) but I thought, the filterFunc field within the vectorScanner struct is being populated within the Next method of the vectorSelector struct. The Next method creates a new vectorScanner and assigns the filterFunc to it before returning the result. After the filterFunc is assigned within the Next method of the vectorSelector struct, the actual filtering process occurs when iterating through the vectors and applying the filterFunc to each StepVector. When the Next method is called, it returns a slice of model.StepVector instances. Each model.StepVector in the returned slice is processed by the filterFunc that was assigned to the corresponding vectorScanner during its creation. The filterFunc is applied to each model.StepVector to perform the desired filtration based on the defined logic within the filterFunc.

MichaHoffmann: I mean it needs to be instantiated somewhere, right now it is always nil as it is never assigned any filter function. You would need to do that. Thats also what the linter is complaining about!

harsh-ps-2003: Oh! I will have to Modify the constructor for logicalplan.VectorSelector to accept the FilterFunc as a parameter and assign it to the FilterFunc field and then pass the instantiated FilterFunc to the vector selector constructor. I will try!

PR #393, opened 2024-01-09
Fix resizing rate buffer tail

fpetkovski:
We use a single slice to rotate buffers for matrix selectors in order to reuse memory. Sizing the slice is done using slices.Grow, but this method does not ensure exact capacity, it only guarantees that the slice will have at least the requested capacity.

To fix the issue, we need to use a subslice explicitly. We saw this issue in production with native histogram queries and it's not easy to reproduce with a test.

### DIFF
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -301,7 +301,7 @@ func selectPoints(
 		for drop = 0; out[drop].T < mint; drop++ {
 		}
 		// Rotate the slice around drop and reduce the length to remove samples.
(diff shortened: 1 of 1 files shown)


MichaHoffmann (commented):
Do we need the same for selectExtPoints?

fpetkovski:
We do not seem to be using this technique for the ext function because it's still not implemented for native histograms. The idea of buffer rotation is to reuse histogram objects, which we ignore for x functions.

MichaHoffmann (approved):
Lgtm

PR #394, opened 2024-01-09
Use ringbuffer for windowing functions

fpetkovski:
The current implementation for matrix selector is based on the one from Prometheus and uses two buffers to keep track of samples, a buffered iterator and a slice of decoded samples for each series. In addition to being harder to reason about, it is also inefficient because we traverse both buffers twice for each step.

This commit rewrites the way points are selected for each step by using the ringbuffer implemented for the subquery operator. This way we only keep a single buffer and we don't need to manage the complexity of rotating samples in multiple places.

Benchmark for relevant queries

```
name                                            old time/op    new time/op    delta
RangeQuery/rate-8                                 20.4ms ± 1%    18.0ms ± 3%  -11.55%  (p=0.008 n=5+5)
RangeQuery/subquery-8                             38.6ms ± 1%    37.7ms ± 1%   -2.35%  (p=0.008 n=5+5)
RangeQuery/sum_rate-8                             18.1ms ± 0%    15.8ms ±11%  -12.67%  (p=0.008 n=5+5)
RangeQuery/sum_by_rate-8                          19.8ms ± 0%    16.9ms ± 6%  -14.58%  (p=0.016 n=4+5)

name                                            old alloc/op   new alloc/op   delta
RangeQuery/rate-8                                 28.6MB ± 0%    27.6MB ± 0%   -3.58%  (p=0.008 n=5+5)
RangeQuery/subquery-8                             31.7MB ± 0%    30.7MB ± 0%   -3.44%  (p=0.008 n=5+5)
RangeQuery/sum_rate-8                             9.33MB ± 0%    8.35MB ± 0%  -10.51%  (p=0.016 n=5+4)
RangeQuery/sum_by_rate-8                          16.8MB ± 0%    15.9MB ± 1%   -5.33%  (p=0.008 n=5+5)

name                                            old allocs/op  new allocs/op  delta
RangeQuery/rate-8                                  72.1k ± 0%     63.2k ± 0%  -12.37%  (p=0.008 n=5+5)
RangeQuery/subquery-8                              92.9k ± 0%     83.9k ± 0%   -9.69%  (p=0.008 n=5+5)
RangeQuery/sum_rate-8                              66.4k ± 0%     57.5k ± 0%  -13.39%  (p=0.000 n=5+4)
RangeQuery/sum_by_rate-8                           85.5k ± 0%     76.6k ± 0%  -10.38%  (p=0.008 n=5+5)

```

### DIFF
--- a/execution/scan/matrix_selector.go
+++ b/execution/scan/matrix_selector.go
@@ -6,15 +6,14 @@ package scan
 import (
 	"context"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
 	"github.com/efficientgo/core/errors"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/value"
-	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
-	"golang.org/x/exp/slices"
 
 	"github.com/thanos-io/promql-engine/execution/function"
 	"github.com/thanos-io/promql-engine/execution/model"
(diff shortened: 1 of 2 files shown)


MichaHoffmann (approved):
Great improvement!

PR #395, opened 2024-01-09
Do not strip parenthesis

fpetkovski:
Trimming parenthesis leads to bugs with distributed query evaluation because Prometheus does not properly stringify subquery expressions when they are not wrapped.

For example, the query `max_over_time((bar/bar)[30s:15s])` gets transformed into `max_over_time(bar/bar[30s:15s])`.


fpetkovski:
Hm, this seems to break explicit assertions now, like `timestamp((bar))`.

MichaHoffmann (approved):
Lgtm

PR #397, opened 2024-01-10
execution,plan: misc cleanups

MichaHoffmann:
* move timestamp handling to plan

### DIFF
--- a/logicalplan/plan.go
+++ b/logicalplan/plan.go
@@ -48,8 +48,7 @@ func New(expr parser.Expr, opts *query.Options) Plan {
 	setOffsetForAtModifier(opts.Start.UnixMilli(), expr)
 	setOffsetForInnerSubqueries(expr, opts)
 
-	// * the engine handles sorting at the presentation layer
-	// * parens are just annoying and getting rid of them doesnt change the query
+	// the engine handles sorting at the presentation layer
 	expr = trimSorts(expr)
 
 	// replace scanners by our logical nodes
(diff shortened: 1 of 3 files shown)


logicalplan/plan.go
```diff
@@ -185,6 +185,23 @@ func replaceSelectors(plan parser.Expr) parser.Expr {
 			*current = &MatrixSelector{MatrixSelector: t, OriginalString: t.String()}
 		case *parser.VectorSelector:
 			*current = &VectorSelector{VectorSelector: t}
+		case *parser.Call:
```
fpetkovski, line 188: Will this code be executed if the MergeSelectors optimizer is disabled?

MichaHoffmann: It runs before optimizers I think

fpetkovski: Got it, I got it confused with something else.

fpetkovski (approved):
Nice cleanup, keeping this sort of logic outside of execution makes complete sense.

fpetkovski:
Tested this a bit, and I think that it introduces a bug with the remote mode since the `timestamp` function will now not be propagated.

MichaHoffmann:
> Tested this a bit, and I think that it introduces a bug with the remote mode since the `timestamp` function will now not be propagated.

Good catch! I fixed it by rendering timestamp function in the vector selector if the flag is set!

fpetkovski (approved):

PR #398, opened 2024-01-11
Transform step invariant nodes during traversal

fpetkovski:
For completeness we should pass step invariant nodes into the transform function during traversal.


MichaHoffmann (approved):
lgtm

PR #399, opened 2024-01-11
execution, engine: duplicate label detection fixes

MichaHoffmann:
* from fuzzing it seems that prometheus checks for duplicate labels in the whole range not just per step

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -127,6 +127,13 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 )`,
 		},
 		{
+			name: "duplicate label fuzz 2",
+			load: `load 30s
+            			http_requests_total{pod="nginx-1"} 1.00+64.00x15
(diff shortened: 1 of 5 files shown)


engine/engine_test.go
```diff
@@ -126,6 +126,13 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
     (-group({__name__="http_requests_total"} @ 54.013) or {__name__="http_requests_total"} offset 1m32s)
 )`,
 		},
+		{
+			name: "duplicate label fuzz 2",
+			load: `load 30s
+            			http_requests_total{pod="nginx-1"} 1.00+64.00x15
+            			http_requests_total{pod="nginx-2"}  6+12.00x21`,
+			query: `-(sum_over_time(http_requests_total[1m]) or http_requests_total) + http_requests_total`,
```
MichaHoffmann, line 134: No step actually has duplicated labels but there are duplicated ones for different timestamps

PR #401, opened 2024-01-12
plan, execution: make duplicate label check a logical node

MichaHoffmann:


fpetkovski (commented):
Do we need to update existing assertions for node types in optimizers for example?

MichaHoffmann:
No because technically this runs after optimizers! Non-recursive calls in execution.go might get confused though

fpetkovski (approved):
Nice cleanup, thanks!

PR #402, opened 2024-01-14
engine: fix range query duplicate label handling

MichaHoffmann:
Prometheus seems to care about duplicate samples for label only during the same timestamp. Since our operators might return samples for one result in two different series we need to merge them to one result again.


engine/engine_test.go
```diff
@@ -144,6 +144,13 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
                 (http_requests_total < 2*http_requests_total)
               )`,
 		},
+		{
+			name: "duplicate label - top level presentation",
```
fpetkovski, line 148: Can we extend the description with what the expected result here is?

MichaHoffmann: Added a slightly better test name! The underlying issue is that we would return something like
```
{} =>
a @ [0]
b @ [3000]
{} =>
c @ [6000]
d @ [9000]
```
on main since the results appear in different series slots at different times

PR #403, opened 2024-01-17
Make duplicate label checks optional

fpetkovski:
The duplicate label checks seem to be breaking use cases where labels are not fully materialized, such as projecting only labels which are included in aggregations or filters.

We are experimenting with a different storage format and this check is breaking some of our queries. I would like to make it possible to disable it until we find out how to deal with it.


MichaHoffmann (approved):
lgtm

PR #406, opened 2024-02-05
Fix histogram queries with start offset

fpetkovski:
The matrix selector needs to ignore histogram samples whose timestamp is before the window mint. We already seem to be doing this correcntly for floats.


MichaHoffmann (approved):
lgtm

PR #407, opened 2024-02-12
Allow injecting custom scanners

fpetkovski:
The current implementation of the engine is coupled to the Prometheus storage model. The input is a SeriesSet which gets iterated on one series at a time. In cases where data is pulled from a different storage, it first has to be transformed into the Prometheus model before it can be queried. This transformation can be resource intensive if the data is not sorted by series and time.

This commit allows extending the engine with custom scanners which take in the logical node and return an execution operator. With this extension point we should get decoupling from Prometheus and can use the engine more efficiently with different storages.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -17,7 +17,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
-	"github.com/prometheus/prometheus/storage"
+	promstorage "github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/util/annotations"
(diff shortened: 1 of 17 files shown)


yeya24:
It is an interesting idea. I wonder what storage you are using rather than Prometheus TSDB.

MichaHoffmann:
> It is an interesting idea. I wonder what storage you are using rather than Prometheus TSDB.

It can enable us down the road to get away from `Select` limitations in thanos.

MichaHoffmann (approved):
lgtm

fpetkovski:
> It is an interesting idea. I wonder what storage you are using rather than Prometheus TSDB.

We are starting to explore storing data in parquet files, possibly by even using https://github.com/polarsignals/frostdb.

yeya24 (approved):

PR #408, opened 2024-02-15
Add projection field to vector selector

fpetkovski:
The Prometheus storage interface selects all labels from the TSDB, even when some labels are not needed for the query. We have tried to add this detection to Thanos itself, but it has turned out to be cumbersome, especially with binary operations that select multiple metrics.

In order to enable this optimization, this commit adds a projection field to the vector selector node which can be used to populate information about labels needed for a query by an optimizer.


MichaHoffmann (approved):
lgtm

PR #409, opened 2024-03-13
Distribute binary expressions with top level functions

fpetkovski:
It's handy to distribute binary expressions where one side of the operation is a no-argument function, since the results will be identical in all remote engines.

### DIFF
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -533,9 +533,6 @@ func isConstantExpr(expr parser.Expr) bool {
 	case *parser.ParenExpr:
 		return isConstantExpr(texpr.Expr)
 	case *parser.Call:
-		if len(texpr.Args) == 0 {
-			return true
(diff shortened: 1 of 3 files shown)


logicalplan/distribute.go
```diff
@@ -535,6 +533,9 @@ func isConstantExpr(expr parser.Expr) bool {
 	case *parser.ParenExpr:
 		return isConstantExpr(texpr.Expr)
 	case *parser.Call:
+		if len(texpr.Args) == 0 {
```
MichaHoffmann, line 536: technically not needed i think; we will already return true if tepr.Args is empty list

MichaHoffmann (approved):
lgtm

logicalplan/distribute.go
fpetkovski: Yes good catch.

PR #410, opened 2024-03-15
engine,plan,execution: string and number literals as logical nodes

MichaHoffmann:
Convert more PromQL AST Expressions into Thanos logical nodes. This is an ongoing effort, this PR only does NumberLiteral and StringLiteral.


fpetkovski (approved):
Will this also work for label_replace and label_join?

MichaHoffmann:
> Will this also work for label_replace and label_join?

Yes! After this; all expressions returned from "Plan()" will use the logical nodes from this commit

logicalplan/plan.go
```diff
@@ -136,10 +135,10 @@ func Traverse(expr *parser.Expr, transform func(*parser.Expr)) {
 
 func TraverseBottomUp(parent *parser.Expr, current *parser.Expr, transform func(parent *parser.Expr, node *parser.Expr) bool) bool {
 	switch node := (*current).(type) {
-	case *parser.StringLiteral:
-		return false
-	case *parser.NumberLiteral:
-		return false
+	case *parser.StringLiteral, *StringLiteral:
```
fpetkovski, line 138: Maybe not something for now, but I think long term all logical nodes should be value objects instead of pointers.

MichaHoffmann: I agree! Lets do that once we have all logical plan nodes in place!

PR #411, opened 2024-03-15
Adding support for quantile_over_time in subqueries

pedro-stanaka:
## Summary

Adding support for quantile_over_time in sub-queries.


execution/scan/subquery.go
```diff
@@ -89,6 +89,11 @@ func (o *subqueryOperator) Next(ctx context.Context) ([]model.StepVector, error)
 		return nil, err
 	}
 
+	scalarArg := 0.0
+	if o.funcExpr.Func.Name == "quantile_over_time" {
+		scalarArg = o.funcExpr.Args[0].(*parser.StepInvariantExpr).Expr.(*parser.NumberLiteral).Val
```
MichaHoffmann, line 94: can we recursively unwrap here? I think we also do that in the matrix selector

MichaHoffmann: can we maybe put https://github.com/thanos-io/promql-engine/blob/8c1e6aea415471112118df6fb3d3bb61a1566883/storage/prometheus/scanners.go#L103

into "extscanners" module and import it here?

MichaHoffmann (approved):
Thanks! LGTM generally; Small nit mentioned above but we can also leave that for a cleanup followup PR i think!

execution/execution.go
```diff
@@ -196,10 +196,14 @@ func newSubqueryFunction(e *parser.Call, t *parser.SubqueryExpr, storage storage
 	if parse.IsExtFunction(e.Func.Name) {
 		return nil, parse.ErrNotImplemented
 	}
-	// TODO: We dont pass arguments yet
 	if e.Func.Name == "quantile_over_time" {
-		return nil, parse.ErrNotImplemented
+		err := errors.Wrapf(parse.ErrNotSupportedExpr, "quantile_over_time with expression as first argument is not supported")
+		_, unwrapErr := extscanners.UnwrapConstVal(e.Args[0])
+		if unwrapErr != nil {
+			return nil, err
```
MichaHoffmann, line 203: can we just return err here without defining it up there?

execution/scan/subquery.go
```diff
@@ -89,6 +93,15 @@ func (o *subqueryOperator) Next(ctx context.Context) ([]model.StepVector, error)
 		return nil, err
 	}
 
+	scalarArg := 0.0
+	var err error
+	if o.funcExpr.Func.Name == "quantile_over_time" {
+		scalarArg, err = extscanners.UnwrapConstVal(o.funcExpr.Args[0])
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to unwrap scalar argument for quantile_over_time")
```
MichaHoffmann, line 101: here we also should return something that implies "not implemented" so we trigger fallback into prometheus engine

extscanners/unwrap.go
```diff
@@ -0,0 +1,19 @@
...
+
+func UnwrapConstVal(e parser.Expr) (float64, error) {
+	switch c := e.(type) {
+	case *parser.NumberLiteral:
+		return c.Val, nil
+	case *parser.StepInvariantExpr:
+		return UnwrapConstVal(c.Expr)
+	}
+
+	return 0, errors.Wrap(parse.ErrNotSupportedExpr, "argument must be a constant value")
```
MichaHoffmann, line 18: I would return regular error here and let the caller decide what to do with it i think

MichaHoffmann:
A few nits again, we can do all of those in followup too if you prefer! The change itself is great

execution/scan/subquery.go
MichaHoffmann: It comes from the unwrap now, but i would prefer marking it here explicitely as not supported instead of reusing the unwrap const val error for that

MichaHoffmann:
I have opened https://github.com/thanos-io/promql-engine/pull/415 today too, we could merge this, rebase and call in the subquery operator; wdyt?

pedro-stanaka:
> I have opened #415 today too, we could merge this, rebase and call in the subquery operator; wdyt?

Please take a last look, if you feel there is something left I think we can take the approach you mentioned here. I was in a hurry and did not pay attention to what I was returning during the refactor.

execution/execution.go
pedro-stanaka: Yep, totally. Inlined it.

execution/scan/subquery.go
pedro-stanaka: But this an error during execution, right? This won't trigger a fallback anymore. Theoretically this should never happen, because it would be caught before when creating the execution plan.

extscanners/unwrap.go
pedro-stanaka: Yeah, I created a new error so we can control on the callee level what we want to do with it. Let me know if you think that is better.

execution/scan/subquery.go
MichaHoffmann: Ah! I missed that and thought it was at creation! True but then again, cannot we unwrap once at construction and remember the value like in matrix selector?

MichaHoffmann (approved):
Lgtm

fpetkovski (approved):
🎉

PR #412, opened 2024-03-15
Improve test suite speed

pedro-stanaka:
## Summary

Using `t.Parallel` to speed up tests and also creating a faster make target to test locally (running all tests, but using build caches).
Also, pulling some of data creation on tests outside so we can avoid generating a lot of data without reason.

Having faster test suit improves feedback loop and CI times.

The native histogram tests are still taking an awful amount of time, but I did not had time to take a look into extracting data creation.

On main I get:

```
      329.71 real       411.05 user        37.16 sys
```

On my branch I get:

```
       80.68 real       434.55 user        90.90 sys
```

without racing it is a much better:

```
real    0m13.486s    user    0m48.597s   sys     0m4.594s
```

#### Changes

- **test speedup**
- **Improving test speed by using parallel testing**
- **Adding target to test without race detector**
- **The export in the makefile had no side-effects**

### DIFF
--- a/engine/distributed_test.go
+++ b/engine/distributed_test.go
@@ -51,14 +51,7 @@ func (p partition) mint() int64 {
 }
 
 func TestDistributedAggregations(t *testing.T) {
-	localOpts := engine.Opts{
-		EngineOpts: promql.EngineOpts{
-			Timeout:              1 * time.Hour,
-			MaxSamples:           1e10,
-			EnableNegativeOffset: true,
-			EnableAtModifier:     true,
-		},
-	}
+	t.Parallel()
 
 	instantTSs := []time.Time{
 		time.Unix(75, 0),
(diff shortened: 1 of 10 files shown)


engine/distributed_test.go
```diff
@@ -255,34 +248,54 @@ func TestDistributedAggregations(t *testing.T) {
...
+					EnableAtModifier:     true,
+				},
+			}
 			for _, test := range tests {
+				test := test
+				var allSeries []*mockSeries
+				remoteEngines := make([]api.RemoteEngine, 0, len(test.seriesSets)+1)
+				for _, s := range test.seriesSets {
+					remoteEngines = append(remoteEngines, engine.NewRemoteEngine(
+						localOpts,
```
MichaHoffmann, line 268: before we would have set lookback here; but now we declare the engines with default lookback right?

pedro-stanaka: You are right, good catch. Pulled data creation for this test again and put engine creation a bit futher down so we can use the new `localOpt` with the proper lookbackDelta.
Cyclomatic complexity on this test is really high.

engine/engine_test.go
```diff
@@ -4740,24 +4759,29 @@ func TestNativeHistograms(t *testing.T) {
 		},
 	}
 
+	defer pprof.StopCPUProfile()
```
MichaHoffmann, line 4762: this still needed? we could just run with "--cpuprofile test.pprof" instead i think

MichaHoffmann:
Awesome, this is great! Thanks

```
$ time go test .
ok  	github.com/thanos-io/promql-engine/engine	30.370s
```

MichaHoffmann (approved):
LGTM

engine/distributed_test.go
MichaHoffmann: Yeah, I think at some point we need to clean up our testing story somehow!

PR #414, opened 2024-03-16
Return series even when duplicated

fpetkovski:
The engine returns an error when duplicate series are present in the final result. It would still be useful to return the result in addition to the error for debugging purposes.


MichaHoffmann (approved):
This makes sense!

MichaHoffmann (approved):

PR #416, opened 2024-03-17
execution: minor subquery cleanup

MichaHoffmann:


extexpr/unwrap.go
```diff
@@ -43,6 +43,13 @@ func UnwrapFloat(expr parser.Expr) (float64, error) {
 	}
 }
 
+// UnsafeUnwrapFloat is like UnwrapFloat but should only be used in cases where the parser
+// guarantees success by already only allowing floats wrapped in parentheses.
+func UnsafeUnwrapFloat(expr parser.Expr) float64 {
```
fpetkovski, line 48: Should we inject this from `execution.go?`

fpetkovski (approved):

extexpr/unwrap.go
MichaHoffmann: :heavy_check_mark:

PR #420, opened 2024-03-21
Clean up preprocess copy

fpetkovski:
Since we removed the vendored parser, we can now also remove the preprocessor copy and use the one from upstream Prometheus.


MichaHoffmann (approved):
Nice

PR #421, opened 2024-03-21
*: rework Explain()/Analyze() name generation

GiedriusS:
Currently the operator descriptions in Explain/Analyze are disconnected. We want for operators to look the same no matter if Explain or Analyze are used. Hence, add support for fmt.Stringer everywhere and use it.


fpetkovski:
Looks good to me, we just need to fix CI.

GiedriusS:
@fpetkovski should be good to go now

MichaHoffmann (approved):
Lgtm

PR #422, opened 2024-03-21
Add NewQueryFromPlan method

fpetkovski:
We would like to be able to propagate the entire plan remotely in order to take advantage of label projections - the ability to read only of subset of labels at query execution time.

In order to do this, we need to also be able to create a query from an existing plan instead of just creating it from a promql string.


MichaHoffmann (approved):
lgtm

PR #423, opened 2024-03-21
Replace function call node

fpetkovski:
This commit replaces the *parser.Call node coming from the AST with a custom logical node type. Long term, we should use our own nodes for everything so that we can enforce certain invariants using our own interface. One method would be having a Children method used for traversal, or having Marshal/Unmarshal methods so we can pass the plan over the wire.

### DIFF
--- a/logicalplan/plan.go
+++ b/logicalplan/plan.go
@@ -112,7 +112,7 @@ func Traverse(expr *parser.Expr, transform func(*parser.Expr)) {
 		transform(expr)
 		Traverse(&node.Param, transform)
 		Traverse(&node.Expr, transform)
-	case *parser.Call:
+	case *FunctionCall:
 		transform(expr)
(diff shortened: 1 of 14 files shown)


logicalplan/plan.go
```diff
@@ -221,7 +228,7 @@ func replacePrometheusNodes(plan parser.Expr) parser.Expr {
 	case *parser.VectorSelector:
 		return &VectorSelector{VectorSelector: t}
 
-	//TODO: we dont yet have logical nodes for these, keep traversing here but set fields in-place
+	// TODO: we dont yet have logical nodes for these, keep traversing here but set fields in-place
```
MichaHoffmann, line 231: Technically we now have one for call so we could move that comment a bit deeper

MichaHoffmann (approved):
lgtm

PR #424, opened 2024-03-21
Replace AST aggregation with custom node

fpetkovski:
This is a follow up to #423 and applies the same replacement to the aggregation node.


MichaHoffmann (approved):
Lgtm

PR #425, opened 2024-03-22
Replace unary and parenthesis with logical nodes

fpetkovski:
This commit replaces parenthesis and the unary expression coming from the AST with logical nodes.


MichaHoffmann:
Lgtm

PR #427, opened 2024-03-22
Replace SubqueryExpr with logical node

fpetkovski:
This commit replaces the SubqueryExpr with a custom logical node.

### DIFF
--- a/logicalplan/options.go
+++ b/logicalplan/options.go
@@ -1,0 +1,4 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package logicalplan
--- a/logicalplan/options.go
+++ b/logicalplan/options.go
@@ -1,4 +1,0 @@
-// Copyright (c) The Thanos Community Authors.
-// Licensed under the Apache License 2.0.
-
-package logicalplan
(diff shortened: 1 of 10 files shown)


logicalplan/options.go
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package logicalplan
```
MichaHoffmann, line 4: do we need this file still?

logicalplan/plan.go
```diff
@@ -247,7 +247,13 @@ func replacePrometheusNodes(plan parser.Expr) parser.Expr {
 	case *parser.StepInvariantExpr:
 		return &StepInvariantExpr{Expr: replacePrometheusNodes(t.Expr)}
 	case *parser.MatrixSelector:
-		return &MatrixSelector{VectorSelector: replacePrometheusNodes(t.VectorSelector), Range: t.Range, OriginalString: t.String()}
+		return &MatrixSelector{
+			VectorSelector: &VectorSelector{
+				VectorSelector: t.VectorSelector.(*parser.VectorSelector),
```
MichaHoffmann, line 252: did recursion break?

MichaHoffmann (approved):
lgtm, small nit

logicalplan/plan.go
fpetkovski: I made `VectorSelector` a concrete type in MarixSelector so we would avoid casts in the rest of the codebase.

MichaHoffmann: Ah! Yeah makes sense!

MichaHoffmann (approved):
Lgtm

PR #428, opened 2024-03-23
Replace parser.Expr with custom Node object

fpetkovski:
This commit introduces a Node interface which is going to contain methods specific to logical nodes. Some of those methods will be Clone, Marshal, Children etc.

In order to keep the changeset manageable, the interface just embeds parser.Expr and allows us to maintain backwards compatibility with the rest of the codebase.

Subsequent changes will remove coupling from parser.Expr and start using Node everywhere.


MichaHoffmann (approved):
🎉

PR #429, opened 2024-03-23
Implement Cloneable interface

fpetkovski:
In #413 we started passing the plan to remote engines instead of a PromQL string. The idea is to allow serializing the plan on the wire with more contextual information which might not be available once the query has been distributed.

That PR also introduced a potential footgun because distributing a query is now done by using multiple pointer copies to the same logical node, instead of making copies of the nodes themselves.

To resolve that, this PR implements a Clone method on all logical nodes and calls it when we need to "fork" a query into multiple subqueries.


MichaHoffmann (approved):
lgtm

PR #430, opened 2024-03-24
Decouple plan from parser.Expr

fpetkovski:
The `logicalplan.Node` interface currently embeds `parser.Expr` for backwards compatibility with the rest of the codebase. This makes it possible to mix AST types with logical plan types which could lead to bugs in type switches and other similiar unexpected behavior.

This commit makes sure that these two interfaces are not interchangeable anymore, which should help us fully decouple the logical plan from the AST coming from the parser.

### DIFF
--- a/execution/step_invariant/step_invariant.go
+++ b/execution/step_invariant/step_invariant.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/efficientgo/core/errors"
 	"github.com/prometheus/prometheus/model/labels"
-	"github.com/prometheus/prometheus/promql/parser"
 
 	"github.com/thanos-io/promql-engine/execution/model"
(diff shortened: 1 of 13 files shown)


execution/step_invariant/step_invariant.go
```diff
@@ -67,7 +66,7 @@ func NewStepInvariantOperator(
 	// We do not duplicate results for range selectors since result is a matrix
 	// with their unique timestamps which does not depend on the step.
 	switch expr.(type) {
-	case *logicalplan.MatrixSelector, *parser.SubqueryExpr:
```
fpetkovski, line 70: Example of a footgun.

MichaHoffmann (approved):
lgtm

PR #431, opened 2024-03-25
Use interface to traverse nodes

fpetkovski:
Now that all logical nodes have the same interface, we can refactor traversal functions to rely on it instead if having a type switch statement.


MichaHoffmann (approved):
lgtm

PR #432, opened 2024-03-25
Implement Marshal and Unmarshal for logical plan.

fpetkovski:
We would like to take advantage of features like projections and read only a subset of labels from store APIs when a query includes an aggregation.

In order to enable this, we need to be able to send the entire plan over the wire since PromQL has no concept of projections. This commit implements the json marshaler methods so that we can serialize the plan and send it in a query request.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -492,7 +492,7 @@ loop:
 	}
 
 	var result parser.Value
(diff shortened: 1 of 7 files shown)


MichaHoffmann (approved):
lgtm

PR #433, opened 2024-03-25
Add constructor for new plan from bytes

fpetkovski:
This commit adds new logical plan constructors which be used to reconstruct an existing plan, without needing a promql query.


MichaHoffmann (approved):
lgtm

PR #437, opened 2024-03-27
Fix unmarshal for strings

fpetkovski:
Fixes unmarshaling of string literals from the plan and adds a test to cover regressions.


pedro-stanaka (approved):
LG

MichaHoffmann (approved):
lgtm, good catch

PR #438, opened 2024-03-28
Refine distributing label_replace

fpetkovski:
We currently fall back with distributing label_replace if the target label is an external label. This is generally correct behavior since external labels are used for deduplication. However, if label-replace is done in the root engine, it is safe to distribute everything bellow it.


logicalplan/distribute_avg.go
```diff
@@ -57,5 +57,5 @@ func isDistributiveOrAverage(expr *Node, skipBinaryPushdown bool) bool {
 	if aggr, ok := (*expr).(*Aggregation); ok {
 		isAvg = aggr.Op == parser.AVG
 	}
-	return isDistributive(expr, skipBinaryPushdown) || isAvg
+	return isDistributive(expr, skipBinaryPushdown, map[string]struct{}{}) || isAvg
```
MichaHoffmann, line 60: dont we also need to check external labels here?

logicalplan/distribute.go
```diff
@@ -20,10 +19,6 @@ import (
 	"github.com/thanos-io/promql-engine/query"
 )
 
-var (
-	RewrittenExternalLabelWarning = errors.Newf("%s: rewriting an external label with label_replace could lead to unpredictable results", annotations.PromQLWarning.Error())
```
MichaHoffmann, line 24: should we keep this as info annotation instead?

fpetkovski: I wouldn't say that results should be unpredictable with this change, but there could be performance issues since a smaller part of the query can be distributed. Should we say that instead?

MichaHoffmann: Yeah, exactely! We could say that that might prevent valuable optimizations.

logicalplan/distribute_avg.go
fpetkovski: I have added a test for the logical plan to illustrate what happens with `avg`. I don't think it is necessary to check labels here because this optimizer just rewrites `avg` as `sum / count`. But please let me know if I am missing something.

MichaHoffmann: Mh worst case we rewrite it to that in something that doesn't distribute because of label replace. It won't be wrong, we just won't use avg locally then right?

fpetkovski: Yes, that's right. We can see if it leads to issues and refine it. At this point it should be related to performance and not correctnes.

logicalplan/distribute.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	RewrittenExternalLabelWarning = errors.Newf("%s: rewriting an external label with label_replace could lead to unpredictable results", annotations.PromQLWarning.Error())
+	RewrittenExternalLabelWarning = errors.Newf("%s: rewriting an external label with label_replace optimizing the query for distributed execution", annotations.PromQLWarning.Error())
```
pedro-stanaka, line 24: The warning is now hard to read. Maybe: 

```suggestion
	RewrittenExternalLabelWarning = errors.Newf("%s: rewriting an external label with label_replace disable optimizations for query distributed execution", annotations.PromQLWarning.Error())
```

pedro-stanaka (approved):
LG

MichaHoffmann (approved):
Lgtm

PR #439, opened 2024-03-28
Minor fixes to InstantQueryFromPlan

fpetkovski:
The query type for instant queries from an existing plan were set to RangeQuery because we were using the same function as the one used for range queries. This leads to empty results for those queries.

We still need to infer the sort order for the instant query, but this might not be needed since the sort would be done at the root anyway.

We will need better integration tests for this function in the future to prevent regressions.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -264,6 +264,44 @@ func (e *Engine) NewInstantQuery(ctx context.Context, q storage.Queryable, opts 
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


yeya24:
We have the promqlsmith based Fuzz tests for Distributed execution as well. It is probably just skipped during CI.
Does it help capture the problem earlier?

MichaHoffmann:
> We have the promqlsmith based Fuzz tests for Distributed execution as well. It is probably just skipped during CI. Does it help capture the problem earlier?

We have to revisit them at some point; right now its not easy to get them to run well without failures

MichaHoffmann (approved):
Lgtm

fpetkovski:
Yeah fuzzing is disabled, but we should try enabling it again. We should also use this method in distributed tests, even though it will require adding another for loop :)

PR #440, opened 2024-03-30
engine,execution: range-vector sub queries with non constant params

MichaHoffmann:
TODO:

* add predict_linear
* allow non-constant parameters in subqueries
* equate - and + Inf in tests ( its a known failure if  we divide by a sum that sums to -0 we return +0 and +inf then )

### DIFF
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -211,12 +211,17 @@ func newSubqueryFunction(e *logicalplan.FunctionCall, t *logicalplan.Subquery, s
 		outerOpts.End = time.UnixMilli(*t.Timestamp)
 	}
 
-	// TODO: predict_linear
 	var scalarArg model.VectorOperator
 	switch e.Func.Name {
 	case "quantile_over_time":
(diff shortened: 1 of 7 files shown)


MichaHoffmann:
Making the parameter non-constant in matrix_selector is slightly trickier and is deferred to follow-up PR

execution/execution.go
```diff
@@ -210,7 +210,26 @@ func newSubqueryFunction(e *logicalplan.FunctionCall, t *logicalplan.Subquery, s
...
+			return nil, err
+		}
+	case "predict_linear":
+		// predict_linear(range-vector, scalar)
+		scalarArg, err = newOperator(e.Args[1], storage, opts, hints)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		scalarArg = noop.NewOperator()
```
fpetkovski, line 229: Beautiful.

fpetkovski (approved):
This only solves subqueries correct? We still need to implement these functions for regular matrix selector.

PR #443, opened 2024-04-08
Fixing peak sample calculation

pedro-stanaka:
## Summary

We introduced a new feature to count peak samples and total samples loaded by query executions, but there was a logic problem in the peak samples function in the AnalyzeOutputNode.
This PR fixes the problems and also introduces a new way to visualize the query analysis to make sure we dont miss changes on sample counting.

### DIFF
--- a/engine/explain_test.go
+++ b/engine/explain_test.go
@@ -183,7 +183,7 @@ func TestQueryAnalyze(t *testing.T) {
 
 func TestAnalyzeOutputNode_Samples(t *testing.T) {
 	t.Parallel()
-	ng := engine.New(engine.Opts{EngineOpts: promql.EngineOpts{Timeout: 1 * time.Hour}, EnableAnalysis: true, MaxShards: 2})
+	ng := engine.New(engine.Opts{EngineOpts: promql.EngineOpts{Timeout: 1 * time.Hour}, EnableAnalysis: true, DecodingConcurrency: 2})
 	ctx := context.Background()
 
 	load := `load 30s
--- a/engine/explain_test.go
+++ b/engine/explain_test.go
@@ -199,6 +199,10 @@ func TestAnalyzeOutputNode_Samples(t *testing.T) {
 	testutil.Ok(t, err)
 	queryResults := query.Exec(context.Background())
 	testutil.Ok(t, queryResults.Err)
+	explainableQuery := query.(engine.ExplainableQuery)
+	analyzeOutput := explainableQuery.Analyze()
+	require.Greater(t, analyzeOutput.PeakSamples(), int64(0))
+	require.Greater(t, analyzeOutput.TotalSamples(), int64(0))
 
 	rangeQry, err := ng.NewRangeQuery(
 		ctx,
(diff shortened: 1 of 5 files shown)


engine/explain_test.go
```diff
@@ -186,39 +187,83 @@ func TestAnalyzeOutputNode_Samples(t *testing.T) {
...
-				http_requests_total{pod="nginx-2"} 1+2x14`
+				http_requests_total{pod="nginx-1"} 1+1x100
+				http_requests_total{pod="nginx-2"} 1+1x100`
 
 	tstorage := promql.LoadedStorage(t, load)
 	defer tstorage.Close()
+	minT := tstorage.Head().Meta().MinTime
+	maxT := tstorage.Head().Meta().MaxTime
+
+	query, err := ng.NewInstantQuery(ctx, tstorage, nil, "http_requests_total", time.Unix(0, 0))
```
fpetkovski, line 198: What is the goal of this instant query? We don't seem to assert anything.

engine/explain_test.go
```diff
@@ -186,39 +187,83 @@ func TestAnalyzeOutputNode_Samples(t *testing.T) {
...
+	explainableQuery = rangeQry.(engine.ExplainableQuery)
 	analyzeOutput = explainableQuery.Analyze()
 	require.Greater(t, analyzeOutput.PeakSamples(), int64(0))
 	require.Greater(t, analyzeOutput.TotalSamples(), int64(0))
+	result := renderAnalysisTree(analyzeOutput, 0)
+	expected := `[duplicateLabelCheck]: 0 peak: 0
+|---[concurrent(buff=2)]: 0 peak: 0
+|   |---[aggregate] sum by ([pod]): 0 peak: 0
+|   |   |---[duplicateLabelCheck]: 0 peak: 0
+|   |   |   |---[coalesce]: 0 peak: 0
```
fpetkovski, line 230: Why do we get 5 shards here? Would this test depend on how many cores a machine has?

MichaHoffmann: I think amount shards should come from query options with default taken from the current "number of cores". That way we could just set it for the test to something lower.

fpetkovski: As long as we do it in a backwards compatible way, and we don't blow up the scope of the PR, I would be fine with that change.

engine/explain_test.go
pedro-stanaka: THe assertions got lost during refactoring. Fixed it

storage/prometheus/scanners.go
```diff
@@ -106,3 +100,13 @@ func (p prometheusScanners) NewMatrixSelector(
 
 	return exchange.NewCoalesce(model.NewVectorPool(opts.StepsBatch), opts, vs.BatchSize*int64(numShards), operators...), nil
 }
+
+func getMaxShards(maxShards int) int {
+	if maxShards < 1 {
+		maxShards = runtime.GOMAXPROCS(0) / 2
+	}
+	if maxShards < 1 {
```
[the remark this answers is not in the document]
pedro-stanaka: This was the previous logic, I guess if GOMAXPROCS is set to 1 this could be 0 (1/2 = 0.5).

fpetkovski (approved):

storage/prometheus/scanners.go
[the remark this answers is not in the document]
MichaHoffmann: I htink we should just consume whatever the opts said here and not have logic here. We should initialize opts with whatever engine says and configure engine by default with what we had before if it was not provided in engine opts.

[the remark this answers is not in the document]
pedro-stanaka: Nice idea. Done

PR #444, opened 2024-04-08
Propagate warnings from execution.New

fpetkovski:
We currently have no way of propagating warnings coming from the construction of the execution plan. This PR should fix that by injecting a context that is capable of gathering warnings using the same mechanism that we use for query execution.

The logical plan returns warnings as a return argument, so we probably want to unify these two mechanisms at one point.

This PR also removes the context from the query.Options struct per Go's guidelines.


MichaHoffmann (approved):
lgtm

PR #445, opened 2024-04-16
Recreate selector matchers during unmarshal

fpetkovski:
When unmarshaling a plan, we need to recreate regex matchers since they contain an internal regex object which is used for matching label values.

### DIFF
--- a/logicalplan/codec.go
+++ b/logicalplan/codec.go
@@ -5,6 +5,8 @@ package logicalplan
 
 import (
 	"encoding/json"
+
+	"github.com/prometheus/prometheus/model/labels"
 )
 
 type jsonNode struct {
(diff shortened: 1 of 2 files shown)


logicalplan/codec.go
```diff
@@ -54,6 +56,13 @@ func unmarshalNode(data []byte) (Node, error) {
 		if err := json.Unmarshal(t.Data, v); err != nil {
 			return nil, err
 		}
+		var err error
+		for i, m := range v.LabelMatchers {
+			v.LabelMatchers[i], err = labels.NewMatcher(m.Type, m.Name, m.Value)
```
fpetkovski, line 61: We could explicitly check for regex types, but recreating everything might be more future proof.

pedro-stanaka (approved):
LG

MichaHoffmann (approved):
lgtm

PR #446, opened 2024-04-16
Chore: Updating prometheus to 0.51.2

pedro-stanaka:
## Summary

Just a small update to Prometheus 0.51.2.


engine/engine.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/util/annotations"
 	"github.com/prometheus/prometheus/util/stats"
+	v1 "github.com/prometheus/prometheus/web/api/v1"
```
fpetkovski, line 23: Hmm was this interface moved back?

pedro-stanaka: Related PR: https://github.com/prometheus/prometheus/pull/13713

MichaHoffmann: Should we perspectively move fallback into the consumers of this library? i.e. make thanos use prometheus engine if creation of a query fails with thanos engine? That way we can skip fallback and prometheus engine here.

MichaHoffmann (approved):
lgtm

engine/engine.go
pedro-stanaka: nvm, you were right. I was not taking `main`.

pedro-stanaka:
Seems like upstream prometheus changed behavior of some aggregation functions, I will try to work on this to unblock the update moving forward. I will have to touch in the implementation here as well. https://github.com/prometheus/prometheus/pull/13744

MichaHoffmann:
The tests are only failing if the first aggregation parameter is a series right?

pedro-stanaka:
> The tests are only failing if the first aggregation parameter is a series right?

Yes

MichaHoffmann:
I think it's a big in Prometheus

PR #447, opened 2024-05-02
Implement active query tracking

fpetkovski:
This commit implements active query tracking in the engine. It uses the existing interface passed in from engine opts and maintains compatibility with the Prometheus engine.


MichaHoffmann (approved):
Lgtm

PR #452, opened 2024-05-14
Tracking peak samples in engine

pedro-stanaka:
## Summary

Right now the engine will report peak samples (in a single step) and total samples always as the same, because we are using the method `IncrementSamplesAtStep` which will update peak and total with the same value. In this PR I am using both values and calling methods on the `stats.QuerySamples` struct directly.


fpetkovski (approved):

fpetkovski:
I guess the question is whether we want to eventually bubble up processed samples from remote engines

PR #453, opened 2024-05-16
Optimize queries on histogram statistics

fpetkovski:
This commit optimizes queries on native histograms when we only need to get the histogram count or sum value.

The approach uses an optimizer which can set a flag in the vector selector to hint to scanners that only histogram stats are needed for the query. When scanners detect this flag, they would use histogram decoders which return histogram objects with populated values for sum, count and the counter reset hint, completely dropping buckets and spans.

Benchmarks
```
benchstat benchmarks/main.out benchmarks/new.out                                                                                                      Filips-MBP-2.fritz.box: Thu May 16 14:03:09 2024

name                                                   old time/op    new time/op    delta
NativeHistograms/selector-11                             82.1ms ± 1%    83.0ms ± 1%   +1.02%  (p=0.032 n=5+5)
NativeHistograms/sum-11                                   127ms ± 1%     128ms ± 3%     ~     (p=0.222 n=5+5)
NativeHistograms/rate-11                                  112ms ± 2%     116ms ± 8%     ~     (p=0.095 n=5+5)
NativeHistograms/sum_rate-11                              148ms ± 1%     149ms ± 2%     ~     (p=0.548 n=5+5)
NativeHistograms/histogram_sum-11                         304ms ± 2%     293ms ± 1%   -3.73%  (p=0.016 n=5+4)
NativeHistograms/histogram_count-11                       306ms ± 2%     295ms ± 2%   -3.57%  (p=0.008 n=5+5)
NativeHistograms/histogram_count_with_sum_and_rate-11     149ms ± 1%      71ms ± 2%  -52.53%  (p=0.016 n=4+5)
NativeHistograms/histogram_quantile-11                    135ms ± 2%     136ms ± 4%     ~     (p=1.000 n=5+5)
NativeHistograms/histogram_scalar_binop-11                208ms ± 1%     216ms ± 6%   +4.07%  (p=0.008 n=5+5)

name                                                   old alloc/op   new alloc/op   delta
NativeHistograms/selector-11                              435MB ± 0%     435MB ± 0%     ~     (p=0.310 n=5+5)
NativeHistograms/sum-11                                   417MB ± 0%     417MB ± 0%     ~     (p=0.310 n=5+5)
NativeHistograms/rate-11                                  389MB ± 0%     389MB ± 0%     ~     (p=0.548 n=5+5)
NativeHistograms/sum_rate-11                              370MB ± 0%     370MB ± 0%     ~     (p=0.222 n=5+5)
NativeHistograms/histogram_sum-11                         437MB ± 0%     346MB ± 0%  -20.74%  (p=0.016 n=4+5)
NativeHistograms/histogram_count-11                       437MB ± 0%     346MB ± 0%  -20.74%  (p=0.008 n=5+5)
NativeHistograms/histogram_count_with_sum_and_rate-11     370MB ± 0%     226MB ± 0%  -39.03%  (p=0.016 n=4+5)
NativeHistograms/histogram_quantile-11                    432MB ± 0%     432MB ± 0%     ~     (p=0.190 n=5+4)
NativeHistograms/histogram_scalar_binop-11                610MB ± 0%     610MB ± 0%     ~     (p=0.841 n=5+5)

name                                                   old allocs/op  new allocs/op  delta
NativeHistograms/selector-11                              5.20M ± 0%     5.20M ± 0%     ~     (p=0.151 n=5+5)
NativeHistograms/sum-11                                   5.19M ± 0%     5.19M ± 0%     ~     (p=0.278 n=5+5)
NativeHistograms/rate-11                                  7.56M ± 0%     7.56M ± 0%     ~     (p=0.730 n=5+5)
NativeHistograms/sum_rate-11                              7.55M ± 0%     7.55M ± 0%     ~     (p=0.595 n=5+5)
NativeHistograms/histogram_sum-11                         5.21M ± 0%     2.36M ± 0%  -54.68%  (p=0.008 n=5+5)
NativeHistograms/histogram_count-11                       5.21M ± 0%     2.36M ± 0%  -54.68%  (p=0.008 n=5+5)
NativeHistograms/histogram_count_with_sum_and_rate-11     7.55M ± 0%     1.66M ± 0%  -78.09%  (p=0.016 n=4+5)
NativeHistograms/histogram_quantile-11                    5.23M ± 0%     5.23M ± 0%     ~     (p=0.286 n=5+4)
NativeHistograms/histogram_scalar_binop-11                8.85M ± 0%     8.85M ± 0%     ~     (p=1.000 n=5+5)
```

### DIFF
--- a/logicalplan/histogram_stats.go
+++ b/logicalplan/histogram_stats.go
@@ -1,3 +1,6 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
 package logicalplan
 
 import (
--- a/logicalplan/histogram_stats.go
+++ b/logicalplan/histogram_stats.go
@@ -12,14 +12,11 @@ import (
 type DetectHistogramStatsOptimizer struct{}
 
 func (d DetectHistogramStatsOptimizer) Optimize(plan Node, _ *query.Options) (Node, annotations.Annotations) {
-	return d.optimize(plan)
+	return d.optimize(plan, false)
 }
 
-func (d DetectHistogramStatsOptimizer) optimize(plan Node) (Node, annotations.Annotations) {
-	var (
-		stop        bool
-		decodeStats bool
-	)
+func (d DetectHistogramStatsOptimizer) optimize(plan Node, decodeStats bool) (Node, annotations.Annotations) {
+	var stop bool
 	Traverse(&plan, func(node *Node) {
 		if stop {
 			return
@@ -27,18 +24,10 @@ func (d DetectHistogramStatsOptimizer) optimize(plan Node) (Node, annotations.An
 		switch n := (*node).(type) {
 		case *VectorSelector:
 			n.DecodeNativeHistogramStats = decodeStats
-		case *Binary:
-			n.LHS, _ = d.optimize(n.LHS)
-			n.RHS, _ = d.optimize(n.RHS)
-			stop = true
-			return
 		case *FunctionCall:
 			if n.Func.Name == "histogram_count" || n.Func.Name == "histogram_sum" {
-				decodeStats = true
-				return
-			}
-			if n.Func.Name == "histogram_quantile" || n.Func.Name == "histogram_fraction" {
-				decodeStats = false
+				n.Args[0], _ = d.optimize(n.Args[0], true)
+				stop = true
 				return
 			}
 		}
--- a/logicalplan/histogram_stats.go
+++ b/logicalplan/histogram_stats.go
@@ -25,8 +25,17 @@ func (d DetectHistogramStatsOptimizer) optimize(plan Node, decodeStats bool) (No
 		case *VectorSelector:
 			n.DecodeNativeHistogramStats = decodeStats
 		case *FunctionCall:
-			if n.Func.Name == "histogram_count" || n.Func.Name == "histogram_sum" {
+			switch n.Func.Name {
+			case "histogram_count", "histogram_sum":
 				n.Args[0], _ = d.optimize(n.Args[0], true)
+				stop = true
+				return
+			case "histogram_quantile":
+				n.Args[1], _ = d.optimize(n.Args[1], false)
+				stop = true
+				return
+			case "histogram_fraction":
+				n.Args[2], _ = d.optimize(n.Args[2], false)
 				stop = true
 				return
 			}
--- a/logicalplan/histogram_stats.go
+++ b/logicalplan/histogram_stats.go
@@ -12,11 +12,26 @@ import (
 type DetectHistogramStatsOptimizer struct{}
 
 func (d DetectHistogramStatsOptimizer) Optimize(plan Node, _ *query.Options) (Node, annotations.Annotations) {
-	var decodeStats bool
+	return d.optimize(plan)
+}
+
+func (d DetectHistogramStatsOptimizer) optimize(plan Node) (Node, annotations.Annotations) {
+	var (
+		stop        bool
+		decodeStats bool
+	)
 	Traverse(&plan, func(node *Node) {
+		if stop {
+			return
+		}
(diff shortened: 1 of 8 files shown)


logicalplan/histogram_stats.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type DetectHistogramStatsOptimizer struct{}
+
+func (d DetectHistogramStatsOptimizer) Optimize(plan Node, _ *query.Options) (Node, annotations.Annotations) {
+	var decodeStats bool
+	Traverse(&plan, func(node *Node) {
+		switch n := (*node).(type) {
+		case *VectorSelector:
+			n.DecodeNativeHistogramStats = decodeStats
```
pedro-stanaka, line 16: Isnt it better to be explicit here and set as false?

storage/prometheus/histograms_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"testing"
+
+	"github.com/prometheus/prometheus/model/histogram"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+	"github.com/prometheus/prometheus/tsdb/tsdbutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHistogramStatsDecoding(t *testing.T) {
```
pedro-stanaka, line 12: nit: 
```suggestion
func TestHistogramStatsIterator_Decode(t *testing.T) {
```

This makes "jump to test/implementation" work better.

pedro-stanaka (approved):
lg, linting failing

logicalplan/histogram_stats.go
fpetkovski: What do you mean by this? Set what to false?

pedro-stanaka: nvm, just realized we are traversing 🙂

logicalplan/histogram_stats.go
```diff
@@ -0,0 +1,36 @@
...
+	var stop bool
+	Traverse(&plan, func(node *Node) {
+		if stop {
+			return
+		}
+		switch n := (*node).(type) {
+		case *VectorSelector:
+			n.DecodeNativeHistogramStats = decodeStats
+		case *FunctionCall:
+			if n.Func.Name == "histogram_count" || n.Func.Name == "histogram_sum" {
```
MichaHoffmann, line 28: i think we never set the flag to false if we hit `histogram_fraction` under a `histogram_count` node

fpetkovski: Yes, but I think this query will return an empty result anyway because `histogram_fraction` expects a vector of histograms as input, and `histogram_count` produces a vector of floats.

MichaHoffmann: Imagine `histogram_count(histogram_fraction(H1) * H2)` kinda query! that would break right?

fpetkovski: In this case `histogram_fraction` also returns float as result, and histogram_count works only on histograms. But maybe it's better to add this reset to stay on the safe side. Would be nice if we had some type checking built in though.

MichaHoffmann: but if H2 is a histogram, wont "float*H2" also be a histogram?

fpetkovski: Ah right, but in the end we just want the count and right? So the multiplication will scale the count from H2 and in the end we will ignore the buckets.

MichaHoffmann: But probably sum could change!

fpetkovski: Let me add a test case for this.

fpetkovski: You are right, this would lead to wrong results. I added test cases and fixed the bug.

MichaHoffmann (approved):
lgtm

PR #455, opened 2024-05-20
misc: bump prometheus to 0.52.0

MichaHoffmann:
For compatibility with prometheus we need to introduce a preprocessor to make scalar arguments in aggregations step invariant. Functions like `quantile_over_time` are exempt from that so far so we dont need to preprocess them as of now.


MichaHoffmann:
cc @pedro-stanaka

yeya24 (approved):

yeya24:
Which upstream commit is it?

> For compatibility with prometheus we need to introduce a preprocessor to make scalar arguments in aggregations step invariant

Does it mean this query is not valid anymore? `topk(scalar(sum(some_metric)), another_metric)`

MichaHoffmann:
> Which upstream commit is it?
> 
> > For compatibility with prometheus we need to introduce a preprocessor to make scalar arguments in aggregations step invariant
> 
> Does it mean this query is not valid anymore? `topk(scalar(sum(some_metric)), another_metric)`

It's valid but its treated as step invariant I think.

fpetkovski (approved):

PR #457, opened 2024-05-31
Cache calls to engine labelsets in distributed execution optimizer

pedro-stanaka:
Calls to `LabelSets()` are quite expensive because Thanos Queriers have to compute label set and remove external ones,
some of our profiles are dominated by those calls, specially here in the distribute optimizer


MichaHoffmann (changes requested):
Since LabelSets() is a method of the RemoteEngine interface, this should probably be a concern of the implementation of the interface and not the distributed optimizer!

MichaHoffmann:
I think this is done in thanos itself now right? Can we close the PR?

pedro-stanaka:
Yep, there is still a loop we could avoid but with the cached call on thanos side this is obsolete now.

PR #458, opened 2024-06-02
logicalplan: best effort reduce constant expressions

MichaHoffmann:
This PR adds a preprocessor pass that reduces constant expressions in a best effort way. This prevents fallback for the remaining functions that cannot take operator arguments in cases where those operators would just reduce to a constant number.

### DIFF
--- a/logicalplan/plan.go
+++ b/logicalplan/plan.go
@@ -65,6 +65,9 @@ func NewFromAST(ast parser.Expr, queryOpts *query.Options, planOpts PlanOptions)
 	// the engine handles sorting at the presentation layer
 	expr = trimSorts(expr)
 
+	// best effort evaluate constant expressions
+	expr = reduceConstantExpressions(expr)
+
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):

logicalplan/plan.go
```diff
@@ -275,6 +278,60 @@ func trimSorts(expr Node) Node {
...
+			if err != nil {
+				return false
+			}
+			rnum, err := UnwrapFloat(tparent.RHS)
+			if err != nil {
+				return false
+			}
+			switch tparent.Op {
+			case parser.ADD:
+				*parent = &NumberLiteral{Val: lnum + rnum}
```
GiedriusS, line 315: Could this impact accuracy if there are multiple additions and we don't use something like Kahan summation?

MichaHoffmann: This is what binary expression would evaluate too at query time too. Since we only likely can have a few summands here it's probably fine!

GiedriusS (approved):

PR #459, opened 2024-06-05
Fix marshaling of +Inf/-Inf values

fpetkovski:
Marshaling of these values currently fails since the default json marshaler does not have support for them.


pedro-stanaka (approved):

MichaHoffmann (approved):
lgtm

PR #465, opened 2024-07-05
Fix distributing constant-side binary operations

fpetkovski:
Binary operations are distributed if one side is a constant expression. This only applies for scalars and not for constant series because series need to be matched on labels.

This commit makes that distinction by introducing a `IsConstantScalarExpr` function and using it in the distributed optimizer.

### DIFF
--- a/execution/aggregate/hashaggregate.go
+++ b/execution/aggregate/hashaggregate.go
@@ -130,7 +130,10 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 		return nil, err
 	}
 	for i := range args {
-		a.params[i] = args[i].Samples[0]
+		a.params[i] = math.NaN()
+		if len(args[0].Samples) == 1 {
+			a.params[i] = args[i].Samples[0]
+		}
 		a.paramOp.GetPool().PutStepVector(args[i])
 	}
 	a.paramOp.GetPool().PutVectors(args)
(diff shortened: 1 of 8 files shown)


fpetkovski:
It looks like the compactor interface changed so I need to add an adapter from the Prometheus to the Thanos interface.

logicalplan/distribute.go
```diff
@@ -463,7 +463,7 @@ func isDistributive(expr *Node, skipBinaryPushdown bool, engineLabels map[string
 	case Deduplicate, RemoteExecution:
 		return false
 	case *Binary:
-		return isBinaryExpressionWithOneConstantSide(e) || (!skipBinaryPushdown && isBinaryExpressionWithDistributableMatching(e))
```
saswatamcode, line 466: Seems like lint fails now as this is unused

fpetkovski:
It turns out the problem was distributing `vector(1)` and similar constant series. I've updated the PR description, PTAL again.

execution/aggregate/hashaggregate.go
```diff
@@ -130,7 +130,10 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 		return nil, err
 	}
 	for i := range args {
-		a.params[i] = args[i].Samples[0]
```
MichaHoffmann, line 133: it looks like we are fixing a defect with "scalar" here; we should fix the root-cause why args has no samples rather

fpetkovski: Hm so will this always come from `scalar` now? Tests started to fail once we made `noop` return empty vectors.

fpetkovski: I updated the PR to not pass `noop` into the aggregate and no we check for `nil` explicitly.

MichaHoffmann (approved):

saswatamcode (approved):
LGTM

PR #470, opened 2024-07-23
Serialize aggregation name

fpetkovski:
When we currently marshal the logical plan, we serialize the integer code of the aggregation instead of the string. These codes can change as new tokens are added to the PromQL grammar. As a result, engines running different versions of Prometheus code might not have identical codes and could see the plan differently.

This commit addresses the issue by making sure the aggregation gets serialized as a PromQL string, which is a much more stable token.


pedro-stanaka (approved):
👍

MichaHoffmann (approved):

saswatamcode (approved):

PR #471, opened 2024-07-25
Remove generics from ringbuffer

fpetkovski:
I would like to start working on streaming windowing functions (rate, x_over_time) in order to reduce memory usage for queries over longer durations.

The idea would be to pass an accumulator function into the ringbuffer which can reduce samples as they come in. In order to enable this, we need to remove the generic type constraint and use a concrete type so that we can access those fields inside the ringbuffer.


MichaHoffmann (approved):
Lgtm

yeya24 (approved):

PR #472, opened 2024-07-29
plan: distribute more binary joins

MichaHoffmann:
We can distribute if we are either matching `on (...)` and "..." contains at least one engine label or if we are matching `without (...)` and "..." does not contain an engine label. If we are matching without either, i.e. `X * Y` we are implicitly matching `without ()`.


fpetkovski:
I think that we can solve the problem in the engine client by passing partition labels as remote engine labels. This way we don't need to change the engine, we can just add a flag in Thanos to determine which labels should be used to identify a remote engine.

MichaHoffmann:
> I think that we can solve the problem in the engine client by passing partition labels as remote engine labels. This way we don't need to change the engine, we can just add a flag in Thanos to determine which labels should be used to identify a remote engine.

Removed the partition labels part and just distribute if the matching contains all engine labels for now!

saswatamcode (approved):
LGTM!

PR #473, opened 2024-07-30
Simplify the ringbuffer

fpetkovski:
This commit simplifies the ringbuffer data structure so that we can remove an extra method.

The commit also simplifies the matrix selector to use a single selectPoints function instead of copying the same logic twice.

### DIFF
--- a/ringbuffer/functions.go
+++ b/ringbuffer/functions.go
@@ -1,7 +1,7 @@
 // Copyright (c) The Thanos Community Authors.
 // Licensed under the Apache License 2.0.
 
-package scan
+package ringbuffer
 
 import (
 	"math"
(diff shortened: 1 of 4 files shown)


ringbuffer/functions.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) The Thanos Community Authors.
 // Licensed under the Apache License 2.0.
 
-package scan
+package ringbuffer
```
fpetkovski, line 4: Moving this to the `ringbuffer` package since the ringbuffer will execute the windowing function in the future. Specialized implementations will do stream based windowing.

MichaHoffmann (approved):

PR #477, opened 2024-08-01
plan: remove dedicated average distribution optimizer

MichaHoffmann:
We move the avg => sum/count optimizer into a preprocessor step of the distributed optimizer. This simplifies the logic a bit and fixes a (admittedly very minor) bug where the avg distribution optimizer considered a label_replace operation that relabels an engine label as distributable.


fpetkovski (commented):
I think it's nice to keep the optimizer separate if possible because we can disable it selectively.

PR #480, opened 2024-08-12
execution: partial responses in distributed engine

MichaHoffmann:
For very distributed setups we need to be able to deal with partial failures. This commit adds an option to continue evaulation if we encounter an error in a remote engine but dont want to fail the whole query.

### DIFF
--- a/engine/distributed_test.go
+++ b/engine/distributed_test.go
@@ -375,3 +375,46 @@ func TestDistributedEngineWarnings(t *testing.T) {
 	res := q.Exec(context.Background())
 	testutil.Equals(t, 1, len(res.Warnings))
 }
+
+func TestDistributedEnginePartialResponses(t *testing.T) {
+	t.Parallel()
+
+	querierErr := &storage.MockQueryable{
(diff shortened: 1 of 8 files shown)


harry671003 (approved):
Thank you. Overall the changes looks good.  
The lint is failing for a minor issue.

fpetkovski:
I am not sure if this belongs in the engine to be fair. We already have this feature in Thanos so we could convert error responses to warnings when executing remote queries: https://github.com/thanos-io/thanos/blob/main/pkg/query/remote_engine.go#L231

MichaHoffmann:
> I am not sure if this belongs in the engine to be fair. We already have this feature in Thanos so we could convert error responses to warnings when executing remote queries: https://github.com/thanos-io/thanos/blob/main/pkg/query/remote_engine.go#L231

Im kinda going back and forth about this in my head, but currently I feel like because the query engine orchestrates all the requests, it is the right place to make decisions about partial responses too. But this is mostly vibe based, just feels better to me right now.

fpetkovski (approved):

PR #482, opened 2024-08-15
Fix counter reset for rate function

fpetkovski:
There is a subtle bug with the rate calculation where we record a counter reset even when the previous sample falls outside of the rate window. This leads to spurious spikes in results, especially when the step is larger than the rate interval.


MichaHoffmann (approved):
lgtm

PR #484, opened 2024-09-17
storage/prometheus: close querier at the end

GiedriusS:
We are trying to reuse memory and hitting a bug where the querier is closed too soon. Prometheus closes the querier at the end of Exec(). We should do that too.

### DIFF
--- a/storage/prometheus/scanners.go
+++ b/storage/prometheus/scanners.go
@@ -6,8 +6,10 @@ package prometheus
 import (
 	"context"
 	"math"
+	"time"
 
 	"github.com/efficientgo/core/errors"
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/promql/parser/posrange"
 	"github.com/prometheus/prometheus/storage"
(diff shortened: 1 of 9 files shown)


MichaHoffmann (approved):

storage/prometheus/scanners.go
```diff
@@ -23,10 +25,149 @@ import (
...
+
+	return minTimestamp, maxTimestamp
 }
 
-func NewPrometheusScanners(queryable storage.Queryable) *Scanners {
-	return &Scanners{selectors: NewSelectorPool(queryable)}
+func NewPrometheusScanners(queryable storage.Queryable, qOpts *query.Options, lplan logicalplan.Plan) (*Scanners, error) {
+	var min, max int64
+	if lplan != nil {
+		min, max = findMinMaxTime(lplan, qOpts)
```
MichaHoffmann, line 161: shouldnt this be set in logicalplan?

GiedriusS: Moved the logic to the logical plan.

fpetkovski (approved):

PR #490, opened 2024-10-03
*: add possibility to add fanout metadata

GiedriusS:
Add a mechanism for adding fanout metadata. This will allow surfacing data that so far is only available in traces and/or logs.

### DIFF
--- a/execution/function/relabel.go
+++ b/execution/function/relabel.go
@@ -37,7 +37,7 @@ func newRelabelOperator(
 		next:     next,
 		funcExpr: funcExpr,
 	}
-	oper.OperatorTelemetry = model.NewTelemetry(oper, opts)
+	oper.OperatorTelemetry = model.NewTelemetry(context.Background(), oper, opts)
 
 	return oper
 }
--- a/execution/model/operator.go
+++ b/execution/model/operator.go
@@ -6,6 +6,7 @@ package model
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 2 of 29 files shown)


pedro-stanaka (approved):

execution/model/operator.go
```diff
@@ -14,6 +15,44 @@ import (
 	"github.com/thanos-io/promql-engine/query"
 )
 
+type metadataKey string
```
MichaHoffmann, line 18: we have similar pattern in /execution/warning  i think, should we move this to /execution/metadata so its similarly structured?

execution/function/relabel.go
```diff
@@ -37,7 +37,7 @@ func newRelabelOperator(
 		next:     next,
 		funcExpr: funcExpr,
 	}
-	oper.OperatorTelemetry = model.NewTelemetry(oper, opts)
+	oper.OperatorTelemetry = model.NewTelemetry(context.Background(), oper, opts)
```
MichaHoffmann, line 40: Can we use context from constructor and pass through from engine? This is a bit more work but feels cleaner, we do same with warnings fwiw - if we want to use different kind of metadata someday

MichaHoffmann (approved):
love the idea! lgtm, just 2 comments

execution/model/operator.go
```diff
@@ -14,6 +15,44 @@ import (
...
+
+func GetMetadataStorage(ctx context.Context) *FanoutMetadata {
+	v := ctx.Value(key)
+	if v == nil {
+		return nil
+	}
+	return v.(*FanoutMetadata)
+}
+
+type FanoutMetadata struct {
```
MichaHoffmann, line 36: I think this could just be "Metadata" as its just a map; applications outside could build on top of this map since "fanout" is somewhat a thanos concept that this engine doesnt know about; its more an implementation detail of Select. Thanos could use this to just add fan out metadata in its Select implementation but no need that this library knows what this key contains really.

MichaHoffmann (changes requested):
On second thought, couldnt this be implemented completely in thanos? As in "create context, add stuff during select, read out in api and add to analysis" kinda way?  If we made sure to use same context for "Select" that was used to create the query?
Kinda like we already pass warnings for partial response.

MichaHoffmann:
> > On second thought, couldnt this be implemented completely in thanos? As in "create context, add stuff during select, read out in api and add to analysis" kinda way? If we made sure to use same context for "Select" that was used to create the query? Kinda like we already pass warnings for partial response.
> 
> How will you attach select metadata to individual operators?

Mh yeah that would not be possible; but then again if its only for fanout there is only vector/matrix selector that do fanout anyway. But for more general metadata thats not possible yeah.

PR #495, opened 2024-11-06
Optimize analysis calculation

fpetkovski:
Aggregating analyzed results is slow for complex expressions because we aggregate samples over and over again.

This commit speeds that up by memoizing aggregates.

### DIFF
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -198,9 +198,9 @@ func (m DistributedExecutionOptimizer) Optimize(plan Node, opts *query.Options) 
 				return true
 			}
 
-			sum := *(*current).(*Aggregation)
+			sum := *(*current).Clone().(*Aggregation)
 			sum.Op = parser.SUM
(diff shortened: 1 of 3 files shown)


logicalplan/distribute.go
```diff
@@ -198,9 +198,9 @@ func (m DistributedExecutionOptimizer) Optimize(plan Node, opts *query.Options)
 				return true
 			}
 
-			sum := *(*current).(*Aggregation)
+			sum := *(*current).Clone().(*Aggregation)
 			sum.Op = parser.SUM
-			count := *(*current).(*Aggregation)
+			count := *(*current).Clone().(*Aggregation)
```
pedro-stanaka, line 203: Is this related?

fpetkovski: Good catch, reverted.

MichaHoffmann (approved):

saswatamcode (approved):
Thanks!

PR #499, opened 2024-12-03
Fix subquery over avg

fpetkovski:
The refactor in https://github.com/thanos-io/promql-engine/pull/477 introduces a bug where subquery on top of avg produces wrong results.

This is because the parents are calculated before the query is rewritten, so the new nodes end up with no parents.


engine/distributed_test.go
```diff
@@ -232,6 +232,8 @@ func TestDistributedAggregations(t *testing.T) {
 		{name: "absent for existing metric with aggregation", query: `sum(absent(foo))`},
 		{name: "absent for existing metric", query: `absent(bar{pod="nginx-1"})`},
 		{name: "absent for existing metric with aggregation", query: `sum(absent(bar{pod="nginx-1"}))`},
+		{name: "subquery with sum/count", query: `max_over_time((sum(bar)/count(bar))[30s:15s])`},
+		{name: "subquery with avg", query: `max_over_time(avg(bar)[30s:15s])`},
```
fpetkovski, line 236: This test fails on main.

MichaHoffmann (approved):

PR #501, opened 2024-12-15
engine: allow to override opts at query time

MichaHoffmann:
This PR allows to override engine settings at query time by providing a custom QueryOption struct that implements the promql.QueryOpts interface and can be provided at runtime.
This prepares eventually overriding more options through this to allow controlling more behavior at query time ( i.e. optimizers, fallback; etc).

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -98,10 +98,25 @@ func (o Opts) getLogicalOptimizers() []logicalplan.Optimizer {
 
 // QueryOpts implements promql.QueryOpts but allows to override more engine default options.
 type QueryOpts struct {
-	promql.QueryOpts
+	lookbackDelta time.Duration
+
+	enablePerStepStats bool
 
 	// DecodingConcurrency can be used to override the DecodingConcurrency engine setting.
 	DecodingConcurrency int
+}
+
+func (opts QueryOpts) LookbackDelta() time.Duration { return opts.lookbackDelta }
+func (opts QueryOpts) EnablePerStepStats() bool     { return opts.enablePerStepStats }
+
+func fromPromQLOpts(opts promql.QueryOpts) *QueryOpts {
+	if opts == nil {
+		return &QueryOpts{}
+	}
+	return &QueryOpts{
+		lookbackDelta:      opts.LookbackDelta(),
+		enablePerStepStats: opts.EnablePerStepStats(),
+	}
 }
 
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -462,19 +432,38 @@ func (e *Engine) NewRangeQueryFromPlan(ctx context.Context, q storage.Queryable,
...
-	return qOpts
+	if opts == nil {
+		return res
+	}
+
+	if opts.LookbackDelta() > 0 {
+		res.LookbackDelta = opts.LookbackDelta()
+	}
+	if opts.EnablePerStepStats() {
+		res.EnablePerStepStats = opts.EnablePerStepStats()
```
MichaHoffmann, line 455: This maybe should be &= but im not sure if original logic was a bug or not!

engine/engine.go
```diff
@@ -96,6 +96,14 @@ func (o Opts) getLogicalOptimizers() []logicalplan.Optimizer {
 	return optimizers
 }
 
+// QueryOpts implements promql.QueryOpts but allows to override more engine default options.
+type QueryOpts struct {
+	promql.QueryOpts
+
+	// DecodingConcurrency can be used to override the DecodingConcurrency engine setting.
+	DecodingConcurrency int
```
yeya24, line 104: Would this become a new query parameter in Thanos to make it overridable?

yeya24 (approved):
LGTM

engine/engine.go
MichaHoffmann: Yeah we could do that if we wanted! It just seemed as good as a start as any. I have planned some followups for this and wanted to keep the PR somewhat scoped

MichaHoffmann: I.e. We could use this to communicate partial response strategy to remote engine too.

yeya24: Sounds good!

fpetkovski:
Maybe we should first create our own interface, and an adapter for the Prometheus engine. It looks like we are struggling to add features because we are coupled to the Prometheus interface.

pedro-stanaka (approved):

MichaHoffmann:
> Maybe we should first create our own interface, and an adapter for the Prometheus engine. It looks like we are struggling to add features because we are coupled to the Prometheus interface.

Will do

fpetkovski (approved):

PR #503, opened 2025-01-03
Implement Native histogram Binary operations and Change Functions for Compatibility with v3.1.0 

SungJin1212:
Implement missing part to address these PRs.
- https://github.com/prometheus/prometheus/pull/14477 - Delayed name dropping
- https://github.com/prometheus/prometheus/pull/14688 - Moved the native histogram binary operation tests into acceptance tests.
- https://github.com/prometheus/prometheus/pull/14611 and more - Added annotations to native histogram

Update
- Update the Prometheus version to v3.1.0 to this PR: https://github.com/prometheus/prometheus/pull/15250)
 
Implement v3.1.0 Prometheus PRs

- https://github.com/prometheus/prometheus/pull/14941
- https://github.com/prometheus/prometheus/pull/15469
- https://github.com/prometheus/prometheus/pull/14910

The `Delayed name dropping` implementation exists in another PR.
- https://github.com/thanos-io/promql-engine/pull/507


pedro-stanaka:
I would suggest updating the title to better reflect what you are actually doing. Are those PRs part of the Prometheus 3.0 update?

SungJin1212:
Yeah, this PR is for the Prometheus 3.0 update.

execution/aggregate/accumulator.go
```diff
@@ -389,6 +406,14 @@ func (a *avgAcc) AddVector(vs []float64, hs []*histogram.FloatHistogram) error {
 	}
 	for _, h := range hs {
 		if err := a.Add(0, h); err != nil {
+			if errors.Is(err, histogram.ErrHistogramsIncompatibleSchema) {
+				// skip warning
```
MichaHoffmann, line 410: Why dont we also write a warning here ( and in the other cases )?

ringbuffer/functions.go
```diff
@@ -170,6 +206,8 @@ var rangeVectorFuncs = map[string]FunctionCall{
 		if err != nil {
 			return 0, nil, false, err
 		}
+		fmt.Println("v, h", v, h)
```
MichaHoffmann, line 209: We should remove this before merging

MichaHoffmann (commented):
I think delayed name removal has couple bugs; as some functions are not implemented in the function operator and either have special operators or are pushed into the vector operator  (like timestamp) or matrix operator (like any range vector function). I would vote that we ignore delayed name removal for now since its behind feature flag in prometheus anyway. If its decided that its going to be removed we have spent effort here that we need to roll back later. Otherwise looks great, thank you for the effort!

execution/aggregate/accumulator.go
SungJin1212: At the Prometheus it is a `warning` but we get the `fail` if we return the error. So, I temporality return `nil`.

SungJin1212:
@MichaHoffmann 
Yes, It's a draft PR, and I gradually update it under some decisions. :D

execution/aggregate/accumulator.go
MichaHoffmann: You can add warnings too here - we have a package for it "execution/warning" I think

SungJin1212:
@MichaHoffmann 
I think we need to upgrade the Prometheus to 3.1.0 to apply https://github.com/prometheus/prometheus/pull/15250, and some package structures have been changed.

harry671003:
@SungJin1212 - Could you also change the branch name in the workflow configuration to `update-prometheus-3.0`, so that it'll trigger the tests?

https://github.com/thanos-io/promql-engine/blob/update-prometheus-3.0/.github/workflows/test.yaml#L7

MichaHoffmann:
> @MichaHoffmann 
> I think we need to upgrade the Prometheus to 3.1.0 to apply https://github.com/prometheus/prometheus/pull/15250, and some package structures have been changed.

I think should not implement the delayed name removal flag here yet! Or at least it should be a separate PR but I think we should not do it unless it's the default in Prometheus

yeya24:
@SungJin1212 Maybe let's do what @MichaHoffmann to temporarily remove the delayed name removal flag.

If the only failed test is warning and info comparison, I think we are ok to merge to unblock 3.0 upgrade even if it failed the CI.

execution/aggregate/accumulator.go
```diff
@@ -287,6 +309,7 @@ func (c *countAcc) Reset(_ float64) {
 }
 
 type avgAcc struct {
+	ctx         context.Context
```
MichaHoffmann, line 312: can we just add it to the "Add" and "AddVector" method as first argument instead as a struct field of the accumulator?

execution/scan/subquery.go
```diff
@@ -170,15 +169,15 @@ func (o *subqueryOperator) Next(ctx context.Context) ([]model.StepVector, error)
...
+			f, h, ok, err := rangeSamples.Eval(ctx, o.params[i], nil)
 			if err != nil {
 				return nil, err
 			}
 			if ok {
 				if h != nil {
 					sv.AppendHistogram(o.pool, uint64(sampleId), h)
-				} else {
-					sv.AppendSample(o.pool, uint64(sampleId), f)
+				} else if f != nil {
```
SungJin1212, line 179: @MichaHoffmann
I made the float64 to pointer to skip the value append when the engine should append nothing.

execution/binary/vector.go
```diff
@@ -12,6 +12,8 @@ import (
 
 	"github.com/cespare/xxhash/v2"
 	"github.com/efficientgo/core/errors"
+	"github.com/prometheus/prometheus/promql/parser/posrange"
```
MichaHoffmann, line 15: can we group with the other prometheus imports?

execution/binary/vector.go
```diff
@@ -323,21 +349,88 @@ func (o *vectorOperator) execBinaryArithmetic(lhs, rhs model.StepVector) (model.
...
+		var err error
+
+		if jp.histogramVal != nil {
+			_, h, keep, err = o.computeBinaryPairing(ctx, 0, 0, hcs.Histograms[i], jp.histogramVal)
+		} else {
+			_, h, keep, err = o.computeBinaryPairing(ctx, 0, jp.val, hcs.Histograms[i], nil)
+		}
 		if err != nil {
-			return model.StepVector{}, err
+			if errors.Is(err, annotations.PromQLInfo) || errors.Is(err, annotations.PromQLWarning) {
```
MichaHoffmann, line 362: why does computeBinaryPairing produce an "info/warning" error? that should not happen i think

execution/binary/vector.go
```diff
@@ -507,78 +600,94 @@ func signatureFunc(on bool, names ...string) func(labels.Labels) uint64 {
...
+		}
+	case hlhs != nil && hrhs == nil:
+		{
+			switch op {
+			case parser.MUL:
+				return 0, hlhs.Copy().Mul(rhs).Compact(0), true, nil
+			case parser.DIV:
+				return 0, hlhs.Copy().Div(rhs).Compact(0), true, nil
+			case parser.ADD, parser.SUB, parser.POW, parser.MOD, parser.EQLC, parser.NEQ, parser.GTR, parser.LSS, parser.GTE, parser.LTE, parser.ATAN2:
+				err := annotations.NewIncompatibleTypesInBinOpInfo("histogram", opName, "float", posrange.PositionRange{})
```
MichaHoffmann, line 659: I dont think we should return an error and an annotation at this point; that feels redundant.

MichaHoffmann: Prometheus also does it, but can we at least add the annotations somewhere else then?

execution/binary/vector.go
SungJin1212: The reason why use `continue` is to ignore append, the current implementation return `model.StepVector{}, nil` , then it appends `0`.
We get the fail like this test caes:
```
eval_info range from 0 to 24m step 6m left_histograms == bool right_floats_for_histograms
  # No results.
```

engine/sort.go
```diff
@@ -7,6 +7,7 @@ import (
 	"math"
 
 	"github.com/facette/natsort"
+
```
harry671003, line 10: nit: Some of these files contain changes that are not relevant like adding whitespace or re-ordering the imports. Could you remove them?

execution/binary/vector.go
yeya24: Is this resolved @SungJin1212?

SungJin1212: I refactored at https://github.com/thanos-io/promql-engine/pull/503/commits/5d75bd37d3b78970af2af8eb07986783bd28970e, but I'm not sure it is the right way.

yeya24:
@fpetkovski Can you also help take a look? If this looks ok we plan to get it merged into main in next PR

execution/aggregate/accumulator.go
```diff
@@ -19,14 +26,14 @@ const (
...
-	Add(v float64, h *histogram.FloatHistogram) error
+	Add(ctx context.Context, v float64, h *histogram.FloatHistogram) error
 	Value() (float64, *histogram.FloatHistogram)
 	ValueType() ValueType
 	Reset(float64)
 }
 
 type vectorAccumulator interface {
-	AddVector(vs []float64, hs []*histogram.FloatHistogram) error
+	AddVector(ctx context.Context, vs []float64, hs []*histogram.FloatHistogram) error
```
fpetkovski, line 36: It seems that the context is used is only used to store warnings. My instinct would be to return those warnings as values instead, and operators can add them to the context.

fpetkovski (approved):
Really great work. I just have one suggestion about the way we process warnings, but it is not a blocker. We can make these changes as a follow up if needed.

yeya24 (approved):

yeya24:
Discussed with @MichaHoffmann offline, let's merge this PR for now and we can iterate on it

ringerc:
For other readers who might've missed it: this PR does _not_ implement the `--promql-delayed-name-removal` feature (https://github.com/prometheus/prometheus/pull/14477) in Thanos; it was split out into https://github.com/thanos-io/promql-engine/pull/507 .

PR #504, opened 2025-01-03
execution: initialize next in aggregation even if we aggregate the labels away

MichaHoffmann:
In aggregations where we aggregate all labels away ( like sum(X) ) we are lazily initializing the "next" vector operator. This leads to us not parallelizing initialization if we are executing queries like `sum by (foo) (X) / on () group_left () sum(X)`. If initialization is expensive, for example in distributed engine; this double query latency.

### DIFF
--- a/execution/aggregate/hashaggregate.go
+++ b/execution/aggregate/hashaggregate.go
@@ -220,6 +220,11 @@ func (a *aggregate) initializeTables(ctx context.Context) error {
 }
 
 func (a *aggregate) initializeVectorizedTables(ctx context.Context) ([]aggregateTable, []labels.Labels, error) {
+	// perform initialization of the underlying operator even if we are aggregating the labels away
+	_, err := a.next.Series(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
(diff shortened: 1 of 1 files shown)


MichaHoffmann:
Traces show that some binary expressions execute each of their arms sequentially instead of parallel, traced it to this missing initialization: 
image

execution/aggregate/hashaggregate.go
```diff
@@ -220,6 +220,11 @@ func (a *aggregate) initializeTables(ctx context.Context) error {
 }
 
 func (a *aggregate) initializeVectorizedTables(ctx context.Context) ([]aggregateTable, []labels.Labels, error) {
+	// perform initialization of the underlying operator even if we are aggregating the labels away
+	_, err := a.next.Series(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
```
fpetkovski, line 227:
```suggestion
	if _, err := a.next.Series(ctx); err != nil {
		return nil, nil, err
	}
```

fpetkovski (approved):
Interesting, not an easy thing to catch!

yeya24 (approved):

PR #506, opened 2025-01-04
Fix sample stats for step invariant and subquery

harry671003:
### Changes
- The samples were not counted correctly for step invariant operator
- Also added the total samples and total samples per step comparison in fuzz tests

### Note
Fuzzing did uncover a case which cannot be addressed without an upstream change


engine/enginefuzz_test.go
```diff
@@ -444,16 +494,27 @@ func getSeries(ctx context.Context, q storage.Queryable) ([]labels.Labels, error
...
-			for _, load := range c.loads {
-				t.Logf(load)
-			}
-			t.Logf(c.query)
+			logQuery(c)
 
 			t.Logf("case %d error mismatch.\nnew result: %s\nold result: %s\n", i, c.newRes.String(), c.oldRes.String())
 			failures++
 		}
+		if c.oldStats.Samples.TotalSamples != c.newStats.Samples.TotalSamples {
```
MichaHoffmann, line 510: I think we use cmp.Equal or something like this normally here

engine/explain.go
```diff
@@ -61,13 +61,25 @@ func (a *AnalyzeOutputNode) aggregateSamples() {
...
 			childPeak := child.PeakSamples()
 			a.peakSamples = max(a.peakSamples, childPeak)
+
+			if a.OperatorTelemetry.SubQuery() {
+				// Skip aggregating from subquery to avoid double counting samples from children.
+				continue
+			}
+
+			if a.OperatorTelemetry.Invariant() {
+				// Invariant operators should have only one samples but they should be counted across all the steps
```
fpetkovski, line 71: What does "should have only one samples" mean here?
```suggestion
				// Invariant operators should have only one samples but they should be counted across all the steps
```

engine/enginefuzz_test.go
```diff
@@ -8,9 +8,12 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"reflect"
 	"testing"
 	"time"
 
+	"github.com/prometheus/prometheus/util/stats"
```
yeya24, line 15: Nit better to group this one with below

execution/model/operator.go
```diff
@@ -22,18 +22,26 @@ type OperatorTelemetry interface {
 	IncrementSamplesAtTimestamp(samples int, t int64)
 	Samples() *stats.QuerySamples
 	SubQuery() bool
+	Invariant() bool
```
yeya24, line 25: Nit, should we name it `StepInvariant` here?

engine/explain.go
```diff
@@ -61,13 +61,25 @@ func (a *AnalyzeOutputNode) aggregateSamples() {
...
 			a.peakSamples = max(a.peakSamples, childPeak)
+
+			if a.OperatorTelemetry.SubQuery() {
+				// Skip aggregating from subquery to avoid double counting samples from children.
+				continue
+			}
+			if a.OperatorTelemetry.StepInvariant() {
+				// Children of step invariant operator outputs one step, but they should be counted towards all the steps.
+				for i := 0; i < len(a.totalSamplesPerStep); i++ {
+					childSamples := child.TotalSamples()
```
fpetkovski, line 72: Do we need to recalculate this for each loop iteration?

I also wonder if we can modify the implementation of the step invariant analysis to fit into the existing algorithm instead of tweaking the algorithm to specially handle the step invariant case.

engine/enginefuzz_test.go
```diff
@@ -32,9 +33,222 @@ import (
...
+				valid = false
+				return errors.New("error")
+			}
+		}
+		return nil
+	})
+	return valid
+}
+
+func FuzzEnginePromQLSmithQueryStatsRangeQuery(f *testing.F) {
```
MichaHoffmann, line 61: Do we need two new fuzz tests? couldnt we incorporate this into existing fuzz tests?

engine/enginefuzz_test.go
MichaHoffmann: Dont see how this was resolved :see_no_evil:

execution/model/operator.go
```diff
@@ -22,18 +22,26 @@ type OperatorTelemetry interface {
 	IncrementSamplesAtTimestamp(samples int, t int64)
 	Samples() *stats.QuerySamples
 	SubQuery() bool
+	StepInvariant() bool
```
MichaHoffmann, line 25: It feels odd to use booleans here, is there a situation where subquery and setpinvariant are true at same time? maybe we could just inject the logicalplan.Node here - it probably has all the information ( and more ) we could ever want

engine/enginefuzz_test.go
harry671003: Oops. Sorry. I thought you meant the comparison of TotalSamplesPerStep below. Latest revision has addressed this.

execution/model/operator.go
harry671003: Does this mean, NewTelemetry should accept an additional parameter for logical node?

```
func NewStepInvariantOperator(...) (model.VectorOperator, error) {
	u := &stepInvariantOperator{
		...
	}
	u.OperatorTelemetry = model.NewTelemetry(u, opts, StepInvariantExpr)
	
}

func NewTelemetry(operator fmt.Stringer, logicalNode Node, opts *query.Options) {
    ...
}

```

engine/explain.go
harry671003: > I also wonder if we can modify the implementation of the step invariant analysis

The other option is to IncrementSamples() in StepInvariant operator and skip counting samples from its children. I think I prefer this option.

execution/model/operator.go
harry671003: I followed your suggestion of injecting the logicalplan.Node there. I had to move the refactor telemetry interfaces into its own package to avoid circular dependency.
It does look better.

fpetkovski (approved):

yeya24 (approved):

PR #507, opened 2025-01-09
Implement delayed name removal

SungJin1212:
Implement delayed name removal
The Prometheus PR:
-  https://github.com/prometheus/prometheus/pull/14477 - Delayed name dropping

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -184,14 +184,15 @@ func NewWithScanners(opts Opts, scanners engstorage.Scanners) *Engine {
 		disableDuplicateLabelChecks: opts.DisableDuplicateLabelChecks,
 		disableFallback:             opts.DisableFallback,
 
-		logger:             opts.Logger,
-		lookbackDelta:      opts.LookbackDelta,
-		enablePerStepStats: opts.EnablePerStepStats,
-		logicalOptimizers:  opts.getLogicalOptimizers(),
-		timeout:            opts.Timeout,
-		metrics:            metrics,
-		extLookbackDelta:   opts.ExtLookbackDelta,
-		enableAnalysis:     opts.EnableAnalysis,
+		logger:                   opts.Logger,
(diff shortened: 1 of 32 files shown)


harry671003:
@SungJin1212 - Is this change still relevant? Could you please rebase?

engine/engine.go
```diff
@@ -327,11 +330,12 @@ func (e *Engine) MakeInstantQueryFromPlan(ctx context.Context, q storage.Queryab
...
 		t:      InstantQuery,
 		// TODO(fpetkovski): Infer the sort order from the plan, ideally without copying the newResultSort function.
-		resultSort: noSortResultSort{},
-		scanners:   scnrs,
-		start:      ts,
-		end:        ts,
-		step:       0,
+		resultSort:               noSortResultSort{},
+		scanners:                 scnrs,
+		enableDelayedNameRemoval: e.enableDelayedNameRemoval,
```
harry671003, line 335: nit: Does this look better in QueryOpts?

execution/model/operator.go
```diff
@@ -18,7 +18,7 @@ type VectorOperator interface {
 	// Series returns all series that the operator will process during Next results.
 	// The result can be used by upstream operators to allocate output tables and buffers
 	// before starting to process samples.
-	Series(ctx context.Context) ([]labels.Labels, error)
+	Series(ctx context.Context) ([]promql.Series, error)
```
harry671003, line 21: This seems like a big change that could break consumers that extend this package. 

Since delayed name removal is [experimental](https://prometheus.io/docs/prometheus/latest/feature_flags/#delay-__name__-label-removal-for-promql-engine), should we wait for it to be stable?

SungJin1212: > should we wait for it to be stable
Yeah, we can wait until it is to be stable.

PR #509, opened 2025-01-15
engine,execution: make remote endpoints injectable

MichaHoffmann:
This PR makes remote endpoints injectable into the distributed engine - this makes it possible to provide endpoints that are configured at runtime - the usecase for this is for example "partial_response" query parameter handling in Thanos.
This mirrors how queryable is passed in at runtime to construct non-distributed queries.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -21,6 +21,7 @@ import (
 	"github.com/prometheus/prometheus/util/annotations"
 	"github.com/prometheus/prometheus/util/stats"
 
+	"github.com/thanos-io/promql-engine/api"
 	"github.com/thanos-io/promql-engine/execution"
 	"github.com/thanos-io/promql-engine/execution/function"
 	"github.com/thanos-io/promql-engine/execution/model"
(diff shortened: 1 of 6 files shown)


engine/engine.go
```diff
@@ -437,12 +436,139 @@ func (e *Engine) MakeRangeQueryFromPlan(ctx context.Context, q storage.Queryable
...
 // NewRangeQuery implements the promql.Engine interface.
 func (e *Engine) NewRangeQuery(ctx context.Context, q storage.Queryable, opts promql.QueryOpts, qs string, start, end time.Time, step time.Duration) (promql.Query, error) {
-	return e.MakeRangeQuery(ctx, q, fromPromQLOpts(opts), qs, start, end, step)
+	return e.MakeRangeQuery(ctx, q, FromPromQLOpts(opts), qs, start, end, step)
+}
+
+// Distributed constructors
+
+// MakeDistributedInstantQuery creates an instant query that is distributed among the passed endpoints.
+func (e *Engine) MakeDistributedInstantQuery(ctx context.Context, q storage.Queryable, endpoints api.RemoteEndpoints, opts *QueryOpts, qs string, ts time.Time) (promql.Query, error) {
```
MichaHoffmann, line 450: q is actually used by distriubted engine sometimes - it feels like a bug, but fixing is out of scope of this PR.

PR #510, opened 2025-01-20
Fix distributed queries with explicit timestamp

fpetkovski:
The distributed optimizer does not consider queries with a specific timestamp which causes them to be incorrect for some cases.

This commit adds a solution for this problem where such queries get distributed only to engines which have sufficient scope to cover its range.

### DIFF
--- a/logicalplan/plan.go
+++ b/logicalplan/plan.go
@@ -245,9 +245,6 @@ func replacePrometheusNodes(plan parser.Expr) Node {
 	case *parser.NumberLiteral:
 		return &NumberLiteral{Val: t.Val}
 	case *parser.StepInvariantExpr:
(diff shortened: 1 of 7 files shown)


pedro-stanaka (approved):

logicalplan/plan.go
```diff
@@ -244,6 +245,9 @@ func replacePrometheusNodes(plan parser.Expr) Node {
 	case *parser.NumberLiteral:
 		return &NumberLiteral{Val: t.Val}
 	case *parser.StepInvariantExpr:
+		//if opts.StripStepInvariant {
```
MichaHoffmann, line 248: some residue left

MichaHoffmann (approved):

PR #513, opened 2025-01-31
engine: fix distributing avg with without grouping

MichaHoffmann:
without (no pun intended) this we rewrite to 
```
sum without (L) (X) / on (L) count without(L) (X) 
```

of course we dont have L label anymore and only one group - the empty one. if now we hhave other labels we get unwanted many-to-many matches and the test errors.

```
                    +(*promql.Result)(found duplicate series for the match group {} on the right hand-side of the operation: [{zone="us-east2"}, {zone="us-east1"}];many-to-many matching not allowed: matching labels must be unique on one side)

```


saswatamcode (approved):

fpetkovski (approved):
Ouch, such a silly bug!

PR #518, opened 2025-02-19
engine: remove implicit fallback

MichaHoffmann:
What to do in cases where this engine does not understand the query should be decided in the calling code. Taking this responsibility of the engine also forces us to deal with queries that we do not understand explicitly in acceptance tests ( by skipping them ) without accidentally falling back.
It also makes the code and test surface a bit smaller for the engine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The project is currently under active development.
 
 ## Roadmap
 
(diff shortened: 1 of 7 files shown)


saswatamcode (approved):
Thanks! Looks good, just small doc comment

fpetkovski (approved):
Thanks, lgtm. We should not forget to update Thanos and move the fallback there.

PR #519, opened 2025-02-20
Do not return pointers from windowing functions

fpetkovski:
Commit c2f322e2f767ce54daac49572e4306c4edf6dd65 introduced large regressions since due to two changes:
* Changed return values of windowing functions from float64 to *float64, causing excessive allocations.
* Added an expensive check on the hot path for a condition that could be calculated once for the entire query.

This commit reverts that changes and adds a better way to handle cases where both float and histogram values are missing, but there is no error in the evaluation. We can easily do this by returning a boolean value.

`BenchmarkSingleQuery` result

```
before c2f322e:
BenchmarkSingleQuery
BenchmarkSingleQuery-11    	       5	 204554700 ns/op	66048614 B/op	  540159 allocs/op

for c2f322e:
BenchmarkSingleQuery
BenchmarkSingleQuery-11    	       4	 279080729 ns/op	153497414 B/op	11378958 allocs/op

for this branch:
BenchmarkSingleQuery
BenchmarkSingleQuery-11    	       5	 204368625 ns/op	64872708 B/op	  540106 allocs/op
```

@SungJin1212 @yeya24 could you take a look?

### DIFF
--- a/execution/scan/subquery.go
+++ b/execution/scan/subquery.go
@@ -201,8 +201,8 @@ func (o *subqueryOperator) Next(ctx context.Context) ([]model.StepVector, error)
 			if ok {
 				if h != nil {
 					sv.AppendHistogram(o.pool, uint64(sampleId), h)
-				} else if f != nil {
-					sv.AppendSample(o.pool, uint64(sampleId), *f)
+				} else {
+					sv.AppendSample(o.pool, uint64(sampleId), f)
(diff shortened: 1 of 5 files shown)


MichaHoffmann (approved):

yeya24 (approved):
Fix makes sense to me. Thanks!

PR #521, opened 2025-03-02
execution: backport histogram stdvar and stddev

MichaHoffmann:
Backport histogram_stdvar and histogram_stddev. Eventually we should import those from prometheus like we do with histogram_quantile and histogram_fraction.


fpetkovski (approved):
LGTM.

Let's maybe open a Prometheus issue to see if we can export these functions?

MichaHoffmann:
> LGTM.
> 
> Let's maybe open a Prometheus issue to see if we can export these functions?

I was planning on issuing PR for it anyway

PR #524, opened 2025-03-06
execution: dont reset aggregation hints in call

MichaHoffmann:
We would like to have grouping hints in select calls for expressions
    like "sum by (foo) (rate(metric[5m]))" too.


MichaHoffmann:
Note - this is tricky because "topk(10, X)" ~ has all labels; "sum(X)" ~ has no labels; both have same groupings; meaning that we cannot use this for projection hints.

MichaHoffmann:
I think this is not well defined sadly

PR #525, opened 2025-03-12
tests: Add fuzz test for native histograms

### LINKED ISSUE #479: Add fuzzing for native histogram queries
The test notation now supports creating native histogram objects, so we can add fuzzing for this datatype too.

Some examples can be found in Prometheus testdata: https://github.com/prometheus/prometheus/blob/main/promql/promqltest/testdata/native_histograms.test.

Saumya40-codes:
Resolves #479 

- This PR adds fuzzing for native histograms
- Also few aggregates functions (max,min) are now fixed on prometheus side, thus related part of code is uncommented

### DIFF
--- a/engine/enginefuzz_test.go
+++ b/engine/enginefuzz_test.go
@@ -298,11 +298,21 @@ func validateTestCases(t *testing.T, cases []*testCase) {
 	}
 }
 
+// Considering first bucket is at index 1
+func getMaxNativeHistogramSumLimit(schema int8, noOfBuckets int) float64 {
+	var maxSum float64
+	for idx := 1; idx <= noOfBuckets; idx++ {
+		maxSum += math.Pow(math.Pow(2, math.Pow(2, float64(-1*schema))), float64(idx))
+	}
+
+	return maxSum
+}
+
 func FuzzNativeHistogramQuery(f *testing.F) {
 	f.Skip()
-	f.Add(int64(0), uint32(0), uint32(60), uint32(120), int16(0), int16(0), float64(1), float64(2), uint64(1), uint64(2), uint64(1))
+	f.Add(int64(0), uint32(0), uint32(60), uint32(120), int8(0), int8(0), uint64(1), uint64(2), uint64(1))
 
-	f.Fuzz(func(t *testing.T, seed int64, startTS, endTS, intervalSeconds uint32, schema1 int16, schema2 int16, sum1, sum2 float64, bucketValue1, bucketValue2, bucketValue3 uint64) {
+	f.Fuzz(func(t *testing.T, seed int64, startTS, endTS, intervalSeconds uint32, schema1 int8, schema2 int8, bucketValue1, bucketValue2, bucketValue3 uint64) {
 		t.Parallel()
 
 		if endTS < startTS || intervalSeconds <= 0 {
@@ -312,6 +322,9 @@ func FuzzNativeHistogramQuery(f *testing.F) {
 		if schema1 < -4 || schema1 > 8 || schema2 < -4 || schema2 > 8 {
 			return
 		}
+
+		sum1 := getMaxNativeHistogramSumLimit(schema1, 3)
+		sum2 := getMaxNativeHistogramSumLimit(schema2, 3)
 
 		if sum1 <= 0 || sum2 <= 0 {
 			return
@@ -336,13 +349,13 @@ func FuzzNativeHistogramQuery(f *testing.F) {
 		}
 
 		load := fmt.Sprintf(`load 2m
-			http_request_duration_seconds{pod="nginx-1"} {{schema:%d count:%d sum:%.2f buckets:[%d %d]}}+{{schema:%d count:%d sum:%.2f buckets:[%d %d %d]}}x20 
-			http_request_duration_seconds{pod="nginx-2"} {{schema:%d count:%d sum:%.2f buckets:[%d]}}+{{schema:%d count:%d sum:%.2f buckets:[%d %d %d]}}x20`,
+			http_request_duration_seconds{pod="nginx-1"} {{schema:%d count:%d sum:%.2f buckets:[%d %d]}}+{{schema:%d count:%d buckets:[%d %d %d]}}x20 
+			http_request_duration_seconds{pod="nginx-2"} {{schema:%d count:%d sum:%.2f buckets:[%d]}}+{{schema:%d count:%d buckets:[%d %d %d]}}x20`,
 			schema1, count1-bucket1[2], sum1, bucket1[0], bucket1[1],
-			schema1, count1, sum1, bucket1[0], bucket1[1], bucket1[2],
+			schema1, count1, bucket1[0], bucket1[1], bucket1[2],
 
 			schema2, count2-bucket2[1]-bucket2[2], sum2, bucket2[0],
-			schema2, count2, sum2, bucket2[0], bucket2[1], bucket2[2],
+			schema2, count2, bucket2[0], bucket2[1], bucket2[2],
 		)
 
 		opts := promql.EngineOpts{
--- a/engine/enginefuzz_test.go
+++ b/engine/enginefuzz_test.go
@@ -299,9 +299,9 @@ func validateTestCases(t *testing.T, cases []*testCase) {
 }
 
 func FuzzNativeHistogramQuery(f *testing.F) {
-	f.Add(int64(0), uint32(0), uint32(30), int16(0), int16(0), float64(1), float64(2), uint64(1), uint64(2), uint64(1))
+	f.Add(int64(0), uint32(30), int16(0), int16(0), float64(1), float64(2), uint64(1), uint64(2), uint64(1))
 
-	f.Fuzz(func(t *testing.T, seed int64, ts, intervalSeconds uint32, schema1 int16, schema2 int16, sum1, sum2 float64, bucketValue1, bucketValue2, bucketValue3 uint64) {
+	f.Fuzz(func(t *testing.T, seed int64, ts uint32, schema1 int16, schema2 int16, sum1, sum2 float64, bucketValue1, bucketValue2, bucketValue3 uint64) {
 		t.Parallel()
 		if schema1 < -4 || schema1 > 8 || schema2 < -4 || schema2 > 8 {
 			return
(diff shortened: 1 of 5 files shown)


Saumya40-codes:
No results gets produced when doing `OR` when a histogram is involved

image

though on prometheus side it works as expected.

---

```
(request_duration_seconds_native) or request_duration_seconds_bucket #normal histogram
```

on prometheus, this contains the combination of the o/p of both of this (LHS and RHS). Though in Thanos engine only the normal histogram is recognized

Saumya40-codes:
Err I was on different go version locally so didnt saw the same error. Fixed now

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,248 @@ func validateTestCases(t *testing.T, cases []*testCase) {
...
+		bucket2 := []uint64{2 * bucketValue1, bucketValue2, 2 * bucketValue3}
+		count1 := bucket1[0] + bucket1[1] + bucket1[2]
+		count2 := bucket2[0] + bucket2[1] + bucket2[2]
+
+		if count1 == 0 || count2 == 0 {
+			return
+		}
+
+		load := fmt.Sprintf(`load 2m
+			http_request_duration_seconds{pod="nginx-1"} {{schema:%d count:%d sum:%.2f buckets:[%d %d]}}+{{schema:%d count:%d sum:%.2f buckets:[%d %d %d]}}x20
```
GiedriusS, line 334: The testing framework sets whatever `sum` you provide, so it could accidentally be invalid. You are calculating the count correctly, but `sum` is limited by the schema and buckets.

```
The upper inclusive limit of a positive bucket: (2**2**-n)**i 
```

So, let's say with schema -1 and buckets 1, 2 the sum cannot be higher than:

(2**2**-1)**1 + (2**2**-1)**2 ~ 3.41. So, you can just take the highest value and you can set it as `sum`. This calculation can be converted into a function.

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,248 @@ func validateTestCases(t *testing.T, cases []*testCase) {
...
+			})
+		}
+
+		if len(cases) > 0 {
+			validateTestCases(t, cases)
+		}
+	})
+}
+
+func FuzzNativeHistogramOnRangeQuery(f *testing.F) {
```
GiedriusS, line 422: I suggest just joining these two together with some extra parameters because a lot of stuff is copied/pasted.

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,248 @@ func validateTestCases(t *testing.T, cases []*testCase) {
...
+			cases = append(cases, &testCase{
+				query:           query,
+				newRes:          newResult,
+				newStats:        newStats,
+				oldRes:          oldResult,
+				oldStats:        oldStats,
+				loads:           []string{load},
+				start:           startTime,
+				end:             endTime,
+				validateSamples: false,
```
GiedriusS, line 537: Maybe we can set this to `true`?

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,248 @@ func validateTestCases(t *testing.T, cases []*testCase) {
 		t.Fatalf("failed %d test cases", failures)
 	}
 }
+
+func FuzzNativeHistogramOnInstantQuery(f *testing.F) {
```
GiedriusS, line 302: Maybe we can add `f.Skip()` for now because these tests unearthed compatibility that will take some time to fix?

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,248 @@ func validateTestCases(t *testing.T, cases []*testCase) {
...
+			Timeout:              1 * time.Hour,
+			MaxSamples:           1e10,
+			EnableNegativeOffset: true,
+			EnableAtModifier:     true,
+			EnablePerStepStats:   true,
+		}
+
+		qOpts := promql.NewPrometheusQueryOpts(true, 0)
+		storage := promqltest.LoadedStorage(t, load)
+		defer storage.Close()
```
GiedriusS, line 353: Let's prefer t.Cleanup

engine/enginefuzz_test.go
Saumya40-codes: Ah I didnt looked through this part and took it like that sum will act same for both native and normal histograms 🫠

Will change based upon the comments!

engine/enginefuzz_test.go
```diff
@@ -298,3 +298,188 @@ func validateTestCases(t *testing.T, cases []*testCase) {
...
+		if math.IsNaN(sum1) || math.IsNaN(sum2) {
+			return
+		}
+		if math.IsInf(sum1, 0) || math.IsInf(sum2, 0) {
+			return
+		}
+
+		rnd := rand.New(rand.NewSource(seed))
+
+		bucket1 := []uint64{bucketValue1, bucketValue2, bucketValue3}
```
GiedriusS, line 342: Small nit: maybe we can move `bucket1`, `bucket2` above `sum1`, `sum2` and use `len(bucket1)`, `len(bucket2)` instead of `3`? It should make it harder to make a mistake in the future.

Saumya40-codes: updated!

engine/enginefuzz_test.go
harry671003: I think Prometheus engine counts samples differently. It divides the size of a histogram point / 16. We could address this in a separate PR.


See: https://github.com/prometheus/prometheus/blob/2bf6f4c9dcbb1ad2e8fef70c6a48d8fc44a7f57c/promql/engine.go#L1559

And: https://github.com/prometheus/prometheus/blob/2bf6f4c9dcbb1ad2e8fef70c6a48d8fc44a7f57c/promql/value.go#L178-L180

harry671003 (approved):
LGTM! Please rebase.

engine/enginefuzz_test.go
Saumya40-codes: Thanks! I did saw total no. of histogram samples being scaled down in all cases, wasn't able to track the reason back then

Saumya40-codes:
Test seems to be failing for few cases locally

```
enginefuzz_test.go:274: load 2m
                                http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20 
                                http_request_duration_seconds{pod="nginx-2"} {{schema:0 count:2 sum:14.00 buckets:[2]}}+{{schema:0 count:6 buckets:[2 2 2]}}x20
    enginefuzz_test.go:276: query: --sum by (pod) ({__name__="http_request_duration_seconds"}), start: 0, end: 0, interval: 0s
    enginefuzz_test.go:282: case 5 error mismatch.
        new result: {pod="nginx-1"} => {count:3, sum:14, (0.5,1]:1, (1,2]:2} @[0]
        {pod="nginx-2"} => {count:2, sum:14, (0.5,1]:2} @[0]
        old result: {pod="nginx-1"} => {count:3, sum:14, (0.5,1]:1, (1,2]:2} @[0]
        {pod="nginx-2"} => {count:2, sum:14, (0.5,1]:2} @[0]
    enginefuzz_test.go:274: load 2m
                                http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20 
                                http_request_duration_seconds{pod="nginx-2"} {{schema:0 count:2 sum:14.00 buckets:[2]}}+{{schema:0 count:6 buckets:[2 2 2]}}x20
    enginefuzz_test.go:276: query: stddev_over_time({__name__="http_request_duration_seconds"} @ end()[1h:1m] offset 28s), start: 0, end: 0, interval: 0s
    enginefuzz_test.go:282: case 12 error mismatch.
        new result: {pod="nginx-1"} => 0 @[0]
        {pod="nginx-2"} => 0 @[0]
        old result: 
    enginefuzz_test.go:297: failed 2 test cases
```
Didn't understand why in the first case, 2nd one has to be with no support for histograms in `stddev_over_time` . Ill try look into 👀

harry671003:
```
enginefuzz_test.go:274: load 2m
                                http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20 
                                http_request_duration_seconds{pod="nginx-2"} {{schema:0 count:2 sum:14.00 buckets:[2]}}+{{schema:0 count:6 buckets:[2 2 2]}}x20
    enginefuzz_test.go:276: query: --sum by (pod) ({__name__="http_request_duration_seconds"}), start: 0, end: 0, interval: 0s
    enginefuzz_test.go:282: case 5 error mismatch.
        new result: {pod="nginx-1"} => {count:3, sum:14, (0.5,1]:1, (1,2]:2} @[0]
        {pod="nginx-2"} => {count:2, sum:14, (0.5,1]:2} @[0]
        old result: {pod="nginx-1"} => {count:3, sum:14, (0.5,1]:1, (1,2]:2} @[0]
        {pod="nginx-2"} => {count:2, sum:14, (0.5,1]:2} @[0]
```


Looks like the problem in the first case is that the compare logic doesn't know about histograms. I can try to push a fix for that today.

Saumya40-codes:
> Looks like the problem in the first case is that the compare logic doesn't know about histograms. I can try to push a fix for that today.

comparer does have method for the same imo

```
compareHistograms := func(l, r *histogram.FloatHistogram) bool {
          if l == nil && r == nil {
          return true
          }
          
          if l == nil && r != nil {
          return false
          }
          
          return l.Equals(r)           <- this is method on prometheus side
}
```

For that particular testcase of what i saw, it looks to be a mismatch in the positive spans, 

result:
[{0 2}] [{0 3}]

expected:
[{0 3}]  [{0 2}] 

Though they both have the share the same meaning overall ig

harry671003:
#576 and #577 should fix the failing fuzz tests. Please rebase after they are merged.

execution/aggregate/accumulator.go
```diff
@@ -454,6 +454,9 @@ func (a *avgAcc) AddVector(ctx context.Context, vs []float64, hs []*histogram.Fl
 }
 
 func (a *avgAcc) Value() (float64, *histogram.FloatHistogram) {
+	if a.histSum != nil {
+		a.histSum.Compact(0)
+	}
```
Saumya40-codes, line 459: same behavior as https://github.com/thanos-io/promql-engine/pull/576 is expected in avg

PR #526, opened 2025-03-17
add support for binaryOps for native histograms

Saumya40-codes:
Found via #525 

Prometheus supports the binary OR, AND, UNLESS operation when a native histogram is there in the query, this PR updates the related functions to take in account for native histograms.

### DIFF
--- a/execution/binary/vector.go
+++ b/execution/binary/vector.go
@@ -255,61 +255,28 @@ func (o *vectorOperator) execBinaryOr(lhs, rhs model.StepVector) (model.StepVect
 	ts := lhs.T
 	step := o.pool.GetStepVector(ts)
 
-	if lhs.HistogramIDs == nil && rhs.HistogramIDs == nil {
-		for i, sampleID := range lhs.SampleIDs {
-			jp := o.hcJoinBuckets[sampleID]
-			jp.ats = ts
-			step.AppendSample(o.pool, o.outputSeriesID(sampleID+1, 0), lhs.Samples[i])
-		}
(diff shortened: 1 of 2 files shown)


execution/binary/vector.go
```diff
@@ -265,6 +265,19 @@ func (o *vectorOperator) execBinaryOr(lhs, rhs model.StepVector) (model.StepVect
 			step.AppendSample(o.pool, o.outputSeriesID(0, sampleID+1), rhs.Samples[i])
 		}
 	}
+
+	for i, histogramID := range lhs.HistogramIDs {
```
harry671003, line 269: Could you add a test case for this?

MichaHoffmann:
What happens if native histogram and float series collide in same hash bucket? Can you add a test for that?

GiedriusS (approved):

MichaHoffmann:
Can you please add some testcases for mixed (histogram vs float) many-to-many matches just to be super sure that those work as intended? Code itself looks solid to me and from running through it in my head i didnt see anything wrong so i would think those test cases will jsut work - but we should have them regardless

Saumya40-codes:
> Can you please add some testcases for mixed (histogram vs float) many-to-many matches just to be super sure that those work as intended? Code itself looks solid to me and from running through it in my head i didnt see anything wrong so i would think those test cases will jsut work - but we should have them regardless

I've added few cases where lhs and rhs are can be (native histogram or a float series), does it needs any changes?

MichaHoffmann (approved):

PR #527, opened 2025-03-17
Fix failure in predict_linear

harry671003:
### Issue
Our continuous correctness tests found an issue with predict_linear. Added the failing test for illustrating the problem.

I'll try to fix it.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -246,6 +246,15 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
 			query: `predict_linear(http_requests_total{route="/"}[1h:1m] offset 1m, 60)`,
 		},
 		{
+			name: "predict_linear correctness",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1"} 1+1.1x10
+			    http_requests_total{pod="nginx-2"} 2+2.3x50`,
+			query: `predict_linear({__name__="http_requests_total", pod!~"nginx-1"}[5m] @ start(), -0.37690610678629094)`,
+			end:   time.Unix(600, 0),
+			start: time.Unix(300, 0),
+		},
+		{
 			name: "duplicate label fuzz",
 			load: `load 30s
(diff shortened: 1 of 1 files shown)


engine/engine_test.go
```diff
@@ -245,6 +245,15 @@ func TestQueriesAgainstOldEngine(t *testing.T) {
...
 		},
+		{
+			name: "predict_linear with step invariant",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1"} 1+1.1x10
+			    http_requests_total{pod="nginx-2"} 2+2.3x50`,
+			query: `predict_linear({__name__="http_requests_total",pod!~"nginx-1"}[5m] @ start(), -0.37690610678629094)`,
+			end:   time.Unix(600, 0),
+			start: time.Unix(300, 0),
+		},
```
yeya24, line 256: Is this an issue for predict linear only or it can impact any function that takes matrix selector with step invariant?

MichaHoffmann:
We could just push step inviarnace up in a preprocessor like here: https://github.com/thanos-io/promql-engine/blob/main/logicalplan/plan.go#L335

engine/engine_test.go
harry671003: It will only impact functions that are at modifier unsafe that takes matrix selector as arg.

https://github.com/prometheus/prometheus/blob/308c8c48c15c74a929c430447df3d3c3a3d4001f/promql/functions.go#L1914

So far only applicable to predict_linear.

MichaHoffmann:
I see, I wonder if we maybe should jsut fall back for now for correctness sake - this seems to be fairly niche usecase that we must weigh against added complexity

harry671003:
> I see, I wonder if we maybe should jsut fall back for now for correctness sake - this seems to be fairly niche usecase that we must weigh against added complexity

I'm okay with doing the fallback. There is one concern, we'll have to exclude the functions.test file from acceptance tests.
If that is okay, I can create a PR.

PR #487, opened 2024-09-19
Update prometheus

harry671003:
### Changes
- This PR updates the  prometheus to https://github.com/prometheus/prometheus/commit/65f610353919
- This is the initial PR that addresses the following breaking changes:
  - https://github.com/prometheus/prometheus/pull/13904 - make lookback and matrix selections left-open and right-closed
  - https://github.com/prometheus/prometheus/pull/14677 - Exponential interpolation
  - https://github.com/prometheus/prometheus/pull/14413 - Kahan summation changes
  - https://github.com/prometheus/prometheus/pull/14655 - Fall back to full label sets when sorting by label
  - https://github.com/prometheus/prometheus/pull/14821 - Correctly handle unary negation

- Tests are expected to fail because the following breaking (and possibly other)  are not addressed yet.
  - https://github.com/prometheus/prometheus/pull/14477 - Delayed name dropping
  - https://github.com/prometheus/prometheus/pull/14688 - Moved the native histogram binary operation tests into acceptance tests.
  - https://github.com/prometheus/prometheus/pull/14611 and more - Added annotations to native histogram

### Note
- The PR is made against a separate branch [prometheus-update-65f610353919](https://github.com/thanos-io/promql-engine/tree/prometheus-update-65f610353919) to co-ordinate the work.

PR #541, opened 2025-03-24
Reset avg accumulator correctly

harry671003:
### Issue
This bug was introduced in https://github.com/thanos-io/promql-engine/pull/487.
When acc.Reset() is called all the parameters should be reset to their defaults.

The issue was captured in our continuous correctness tests.


MichaHoffmann (approved):

yeya24 (approved):
Thanks

PR #543, opened 2025-03-26
handle histogram_avg using DetectHistogramStatsOptimizer

yeya24:
Handle histogram_avg using DetectHistogramStatsOptimizer as it only needs to access sum and count.


fpetkovski (approved):
Nice

MichaHoffmann (approved):

PR #544, opened 2025-03-26
Tracing: enable tracing for main query execution steps

pedro-stanaka:
Signed-off-by: Pedro Tanaka <pedro.tanaka@shopify.com>


engine/engine.go
```diff
@@ -237,14 +248,21 @@ type Engine struct {
 }
 
 func (e *Engine) MakeInstantQuery(ctx context.Context, q storage.Queryable, opts *QueryOpts, qs string, ts time.Time) (promql.Query, error) {
+	span, ctx := tracing.StartSpanFromContext(ctx, "engine.MakeInstantQuery")
+	defer span.Finish()
```
harry671003, line 252: Do we expect the engine to take a lot of time before query execution phase? Does these spans add value?

engine/engine.go
```diff
@@ -528,8 +569,21 @@ type compatibilityQuery struct {
 }
 
 func (q *compatibilityQuery) Exec(ctx context.Context) (ret *promql.Result) {
+	span, ctx := tracing.StartSpanFromContext(ctx, "compatibilityQuery.Exec")
+	defer span.Finish()
+	span.SetTag("query_type", q.t.String())
+	span.SetTag("query_string", q.String())
+	if q.t == RangeQuery {
```
harry671003, line 576: nit: Should we just set start=end and step=0 for instant query?

harry671003 (approved):
Thanks for the great work! Left some minor comments.

As a next step, should the time taken in each operator be measured?

PR #546, opened 2025-03-29
Replace Bingo with go tool

fpetkovski:
Since the release of go 1.24, we can manage tools directly in a go.mod file.

We should also replace goimports with gci for better import sorting.


pedro-stanaka (approved):
lgtm

MichaHoffmann (approved):

yeya24 (approved):
Thanks a lot!

PR #547, opened 2025-04-02
feat: Add support for limitk aggregations

Saumya40-codes:
- add support for limitk
- have tested it against a dummy [test function](https://github.com/Saumya40-codes/promql-engine/blob/c8727eeeb9e78118cfe67e476656aea69c09c318/engine/limitk_test.go)

### To be done
- add tests
- add support for native histograms (it is in prometheus)
- Should we use actually use heap for this?


harry671003:
Thanks for working on this. 

Currently, the engine throws Unsupported error for limitk which allows Thanos (or Cortex) to retry the query using Promethues engine. But with the current state of the PR, the engine will not throw the error.

I think this makes the native histogram support necessary for the new operators before the PR can be merged.

Saumya40-codes:
> I think this makes the native histogram support necessary for the new operators before the PR can be merged.

Yes previously it was failing.

I've tried to cover native histograms and mixed types in the latest commit. For the mixed ones though, had to do it differently as its first 'k' should be from float and histograms as one.

Saumya40-codes:
Ah, also for this I wonder if test based on range query should be added. As the first 'k' samples might differ by timestamps

for instance:

below are the samples at timestamps (where it changes)
```
[0 1 2 3 4] 1498000
[0 1 3 4] 1500000
```

Thus here answer would be different for limitk when k >= 3

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +209,113 @@ func (a *kAggregate) init(ctx context.Context) error {
...
 			}
 		}
 	}
 
 	s := a.vectorPool.GetStepVector(t)
 	for _, h := range a.heaps {
-		// The heap keeps the lowest value on top, so reverse it.
-		if len(h.entries) > 1 {
-			sort.Sort(sort.Reverse(h))
-		}
```
Saumya40-codes, line 230: Gets handled at 

https://github.com/thanos-io/promql-engine/blob/4055a112d1eaba4282ec4a3dd2856b36d8507f95/engine/sort.go#L69

execution/aggregate/khashaggregate.go
```diff
@@ -61,10 +61,14 @@ func NewKHashAggregate(
 		compare = func(f float64, s float64) bool {
 			return f < s
 		}
-	} else {
+	} else if aggregation == parser.BOTTOMK {
```
GiedriusS, line 64: Maybe we should error out if some other, unsupported aggregation is passed? Also, it will protect us against any potential bugs because previously we were always having this `compare` function.

execution/aggregate/khashaggregate.go
```diff
@@ -133,11 +137,11 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 			result = append(result, a.GetPool().GetStepVector(vector.T))
```
GiedriusS, line 137: Above we have:
```
// Skip steps where the argument is less than or equal to 0.
```

Could we add a test with negative values on limitk?

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +209,113 @@ func (a *kAggregate) init(ctx context.Context) error {
...
-	for i, sId := range sampleIDs {
-		h := a.inputToHeap[sId]
-		switch {
-		case h.Len() < k:
-			heap.Push(h, &entry{sId: sId, total: samples[i]})
+func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampleIDs []uint64, samples []float64, histogramIDs []uint64, histograms []*histogram.FloatHistogram) {
+	groupsRemaining := len(a.heaps)
+
+	if a.aggregation != parser.LIMITK {
+		for i, sId := range sampleIDs { // BOTTOMK, TOPK
```
GiedriusS, line 216: Maybe we could make the code more straightforward by writing this instead of putting it into a comment:

```
if a.aggregation == parser.BOTTOMK || a.aggregation == parser.TOPK {
...
} else {
...
}
```

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +209,113 @@ func (a *kAggregate) init(ctx context.Context) error {
...
+				}
+			}
+		}
+	} else {
+		if len(histogramIDs) == 0 {
+			for i, sId := range sampleIDs {
+				h := a.inputToHeap[sId]
+				switch {
+				case h.Len() < k:
+					heap.Push(h, &entry{sId: sId, total: samples[i]})
```
GiedriusS, line 237: This also only works with limitk, right? So, we probably need to explicitly avoid this branch by hardcoding the aggregations above into an `if` statement.

Saumya40-codes:
>  add tests for range queries (specifically, when a particular sample which existed on some previous time stamp 
`i` but is not longer present at current timestamp `j`, this doesn't provide right results on prometheus side)

For this, there is PR on prometheus side which will fix it: `https://github.com/prometheus/prometheus/pull/16404`  (Fix: was here:  `https://github.com/prometheus/prometheus/pull/16404/files#diff-ad2761bb162a2f12326b18f2714ef54f4cd476cfe3388f77b2b20c5ff82e4faaR3357-R3359`

Though it also adds support for variable scalar parameters in aggregations

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +209,113 @@ func (a *kAggregate) init(ctx context.Context) error {
...
+						groupsRemaining--
+					}
+
+					if groupsRemaining == 0 {
+						break
+					}
+				}
+			}
+		} else {
+			// Mixed type histograms or native histograms only
```
GiedriusS, line 249: We can see that, probably not worth adding this as a comment.

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +209,113 @@ func (a *kAggregate) init(ctx context.Context) error {
...
 	s := a.vectorPool.GetStepVector(t)
 	for _, h := range a.heaps {
-		// The heap keeps the lowest value on top, so reverse it.
-		if len(h.entries) > 1 {
-			sort.Sort(sort.Reverse(h))
-		}
-
 		for _, e := range h.entries {
-			s.AppendSample(a.vectorPool, e.sId, e.total)
+			if a.aggregation != parser.LIMITK || e.histogramSample == nil {
```
GiedriusS, line 302: I suggest doing this differently to make it more future-proof:

```
if e.histogramSample != nil {
  s.AppendHistogram(...)
} else {
  s.AppendSample(...)
}
```

What kind of aggregation it is and what kind of sample types are supported by them are handled one layer above.

engine/engine_test.go
```diff
@@ -3538,6 +3538,33 @@ min without () (
...
 		},
+		{
+			name: "limitK",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", series="1"} 1
+			    http_requests_total{pod="nginx-2", series="3"} 2
+			    http_requests_total{pod="nginx-3", series="2"} 8
+			    http_requests_total{pod="nginx-4", series="5"} 6
+			    http_requests_total{pod="nginx-5", series="4"} 8
+			    http_requests_total{pod="nginx-6", series="7"} 15
```
GiedriusS, line 3549: Can we also add tests with mixed sample types and one with only native histograms?

execution/aggregate/khashaggregate.go
Saumya40-codes: Yes. Changed from just "else" to explicitly checking the type aggregation

engine/engine_test.go
Saumya40-codes: It gets handled [here](https://github.com/thanos-io/promql-engine/pull/547/files#diff-cbf9ce27965b15f1e13be995f164fd6b33a19381364c4be3e8c047d528994495R5492-R5499)

MichaHoffmann:
Random question: does it make sense to also implement something in Store API on thanos side to utilize hints we set? The expensive part is to fetch everything - we could benefit from just not doing that if we throw most of it away in the engine at the end.

execution/aggregate/khashaggregate.go
```diff
@@ -61,10 +61,16 @@ func NewKHashAggregate(
 		compare = func(f float64, s float64) bool {
 			return f < s
 		}
-	} else {
+	} else if aggregation == parser.BOTTOMK {
 		compare = func(f float64, s float64) bool {
 			return s < f
 		}
+	} else if aggregation == parser.LIMITK {
+		compare = func(f1, f2 float64) bool {
```
saswatamcode, line 69: Do we even need to assign this?
As we are only ever using compare in topk/bottomk below?

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
...
+				}
+			}
+		} else {
+			histogramIndex := 0
+			sampleIndex := 0
+
+			for histogramIndex < len(histogramIDs) || sampleIndex < len(sampleIDs) {
+				var currentID uint64
+
+				// Pick the relatively first id in the sample
```
saswatamcode, line 257: Not a blocker, but you can probably just have a comment that describes merge loop above, instead of these small ones.

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
...
-			heap.Push(h, &entry{sId: sId, total: samples[i]})
+func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampleIDs []uint64, samples []float64, histogramIDs []uint64, histograms []*histogram.FloatHistogram) {
+	groupsRemaining := len(a.heaps)
 
-		case h.compare(h.entries[0].total, samples[i]) || (math.IsNaN(h.entries[0].total) && !math.IsNaN(samples[i])):
-			h.entries[0].sId = sId
-			h.entries[0].total = samples[i]
+	if a.aggregation == parser.TOPK || a.aggregation == parser.BOTTOMK {
+		for i, sId := range sampleIDs {
+			h := a.inputToHeap[sId]
```
saswatamcode, line 219: Maybe small nit, but with histogram code here, `h` is maybe not the greatesy var name. I wonder if we can just name it sampleHeap?

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
...
-		switch {
-		case h.Len() < k:
-			heap.Push(h, &entry{sId: sId, total: samples[i]})
+func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampleIDs []uint64, samples []float64, histogramIDs []uint64, histograms []*histogram.FloatHistogram) {
+	groupsRemaining := len(a.heaps)
 
-		case h.compare(h.entries[0].total, samples[i]) || (math.IsNaN(h.entries[0].total) && !math.IsNaN(samples[i])):
-			h.entries[0].sId = sId
-			h.entries[0].total = samples[i]
+	if a.aggregation == parser.TOPK || a.aggregation == parser.BOTTOMK {
```
saswatamcode, line 217: Maybe a switch is better here too?

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
 	return nil
 }
 
-func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampleIDs []uint64, samples []float64) {
-	for i, sId := range sampleIDs {
-		h := a.inputToHeap[sId]
-		switch {
-		case h.Len() < k:
-			heap.Push(h, &entry{sId: sId, total: samples[i]})
+func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampleIDs []uint64, samples []float64, histogramIDs []uint64, histograms []*histogram.FloatHistogram) {
```
saswatamcode, line 214: This is handling a few cases now, I'd add a comment, detailing those for posterity.

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
...
+					}
+
+					if groupsRemaining == 0 {
+						break
+					}
+				}
+			}
+		} else {
+			histogramIndex := 0
+			sampleIndex := 0
```
saswatamcode, line 252: We can probably have
haveSample := sampleIndex < len(sampleIDs)
haveHistogram := histogramIndex < len(histogramIDs)

to make the code below a bit more cleaner

execution/aggregate/khashaggregate.go
```diff
@@ -205,41 +211,112 @@ func (a *kAggregate) init(ctx context.Context) error {
...
-			s.AppendSample(a.vectorPool, e.sId, e.total)
+			if e.histogramSample == nil {
+				s.AppendSample(a.vectorPool, e.sId, e.total)
+			} else {
+				s.AppendHistogram(a.vectorPool, e.histId, e.histogramSample)
+			}
 		}
 		h.entries = h.entries[:0]
 	}
+
```
saswatamcode, line 311: Not a blocker, but this might be a nice method to have for samplesHeap slices?

saswatamcode (commented):
Thanks! This already looks really good. Some small nits/comments

execution/aggregate/khashaggregate.go
Saumya40-codes: heap.Push is internally calling
```
func (s samplesHeap) Less(i, j int) bool {
	if math.IsNaN(s.entries[i].total) {
		return true
	}
	return s.compare(s.entries[i].total, s.entries[j].total)
}
```

so not assigning something to compare might just call nil pointer de ref

Saumya40-codes: added a nil check in Less(i,j int) for limitk

engine/distributed_test.go
```diff
@@ -214,6 +214,7 @@ func TestDistributedAggregations(t *testing.T) {
 		{name: "group", query: `group by (pod) (bar)`},
 		{name: "topk", query: `topk by (pod) (1, bar)`},
 		{name: "bottomk", query: `bottomk by (pod) (1, bar)`},
+		/* 		{name: "limitk", query: `limitk by (pod) (1, bar)`}, */
```
Saumya40-codes, line 217: Looking into this.

In this in few cases the sample provided by the engine differs from the what actual first 'k' samples should be 
(which works right in normal implementation giving same o/p as prometheus engine)

Saumya40-codes:
> Random question: does it make sense to also implement something in Store API on thanos side to utilize hints we set? The expensive part is to fetch everything - we could benefit from just not doing that if we throw most of it away in the engine at the end.

Thanks for this! 

(re: meeting) Think we will work on this after implementing aggregations.

execution/aggregate/khashaggregate.go
```diff
@@ -211,34 +207,40 @@ func (a *kAggregate) init(ctx context.Context) error {
...
+	case parser.LIMITK:
 		if len(histogramIDs) == 0 {
 			for i, sId := range sampleIDs {
-				h := a.inputToHeap[sId]
+				sampleHeap := a.inputToHeap[sId]
 				switch {
-				case h.Len() < k:
-					heap.Push(h, &entry{sId: sId, total: samples[i]})
+				case sampleHeap.Len() < k:
```
GiedriusS, line 240: There's only one case here so we can convert this into an `if`. But in general we can only arrive here if `sampleHeap.Len() < k` so I think we can remove this check in general. Maybe we can do some defensive programming: `if sampleHeap.Len() >= k { panic("bug, please report") }` (or something similar)

Saumya40-codes: Ah right should be just `if`

This condition though is useful when their is grouped aggregation `limitk(2, something) by (some)` so for some label we would've collected `k` metrics but we still need to iterate for the rest of the groups 👀

engine/distributed_test.go
```diff
@@ -251,6 +250,8 @@ func TestDistributedAggregations(t *testing.T) {
 		{name: "query with numeric timestamp", query: `sum(bar @ 140.000)`},
 		{name: "query with range and @end() timestamp", query: `sum(count_over_time(bar[1h] @ end()))`},
 		{name: `subquery with @end() timestamp`, query: `bar @ 100.000 - bar @ 150.000`},
+		{name: "limitk", query: `limitk by (pod,zone) (1, bar)`}, // this is more of a condition when o/p will be consistent with prometheus, in certain cases
```
GiedriusS, line 253: Maybe we should mention explicitly that this is because sample/histogram IDs are internal in each leaf node and the distributed engine doesn't know about them? 

Also, you wrote `by (pod, zone)` so that we would avoid having to write custom test code, right?

Saumya40-codes: > Also, you wrote `by (pod, zone)` so that we would avoid having to write custom test code, right?

 yes, so final o/p will be just combination of all the dedup (remote  `limikt(1,bar) by (pod,zone)`) o/p we get from the leaf queries, so it be deterministic throughout

saswatamcode (approved):
Thanks! I think this looks quite good! Let's merge and iterate with limit_ratio!

PR #549, opened 2025-04-08
Add projection pushdown optimizer

yeya24:
Fixes #496 

This PR adds an implementation of Projection Pushdown optimizer which checks the expression and injects projection to vector selector accordingly.

Note:

- It supports projection pushdown for aggregation.
- It supports projection pushdown for one to one binary expression. Also support one-to-many and many-to-one binary expressions mainly on the low cardinality side. Thanks for the inspiration from https://github.com/thanos-io/promql-engine/pull/550
- It supports functions like label_replace and label_join. And other functions such as `scalar`, `absent` and `absent_over_time` by inspiration from https://github.com/thanos-io/promql-engine/pull/550
- Changes made to skip merging select if the vector selector has projection. For details of why we don't do a full support for merge projection, see https://github.com/thanos-io/promql-engine/pull/549#issuecomment-2823226765
- Storage layer needs to return a label of series hash for proper deduplication.

It is able to pass promqlsmith based fuzz test locally by using a mock implementation of projection querier.

PR #550, opened 2025-04-11
Add optimizer for extracting projection labels

fpetkovski:
This commit adds a logical optimizer for extracting projection labels from a query.

### DIFF
--- a/logicalplan/set_projection_labels.go
+++ b/logicalplan/set_projection_labels.go
@@ -1,0 +1,152 @@
+package logicalplan
+
+import (
+	"maps"
+	"slices"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/util/annotations"
+	"github.com/thanos-io/promql-engine/query"
+)
+
+// setProjectionLabels is an optimizer that sets the projection labels for all vector selectors.
+// If a projection is already set as a matcher, it will be materialized in the selector
+// and the matcher will be removed.
+// This is useful for sending projections as part of a remote query.
+type setProjectionLabels struct{}
+
+const seriesIDLabel = "__series__id"
+
+func (s setProjectionLabels) Optimize(expr Node, _ *query.Options) (Node, annotations.Annotations) {
+	var hasProjections bool
+	TraverseBottomUp(nil, &expr, func(_ *Node, current *Node) bool {
+		switch e := (*current).(type) {
+		case *VectorSelector:
+			// Check if a projection is already set in the node.
+			hasProjections = e.Projection.Include || len(e.Projection.Labels) > 0
+		}
+		return hasProjections
+	})
+	if hasProjections {
+		return expr, annotations.Annotations{}
+	}
+
+	var projection Projection
+	return s.optimize(expr, projection)
+}
+
+func (s setProjectionLabels) optimize(expr Node, projection Projection) (Node, annotations.Annotations) {
+	var stop bool
+	Traverse(&expr, func(current *Node) {
+		if stop {
+			return
+		}
+		switch e := (*current).(type) {
+		case *Aggregation:
+			switch e.Op {
+			case parser.TOPK, parser.BOTTOMK:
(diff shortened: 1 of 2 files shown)


logicalplan/set_projection_labels.go
```diff
@@ -0,0 +1,152 @@
...
+		switch e := (*current).(type) {
+		case *Aggregation:
+			switch e.Op {
+			case parser.TOPK, parser.BOTTOMK:
+				projection = Projection{}
+			default:
+				projection = Projection{
+					Labels:  slices.Clone(e.Grouping),
+					Include: !e.Without,
+				}
```
yeya24, line 53: May I know how your scanner is implemented when it is a without? Does it return series hash label in this case?

https://github.com/thanos-io/promql-engine/pull/549/files#diff-2d378c97867270a9cf6883930a4a229b3928d637e4c5c558abf5f4bd913f2f8cR41

This is what I implemented as a mock storage which does projection and always return series hash when projection is available. So maybe there is some difference here.

logicalplan/set_projection_labels.go
```diff
@@ -0,0 +1,152 @@
...
+				Include: e.VectorMatching.On,
+				Labels:  append([]string{seriesIDLabel}, e.VectorMatching.MatchingLabels...),
+			}, e.VectorMatching.Include)
+			s.optimize(lowCard, lcProjection)
+			stop = true
+		case *VectorSelector:
+			slices.Sort(projection.Labels)
+			projection.Labels = slices.Compact(projection.Labels)
+			e.Projection = projection
+			projection = Projection{}
```
yeya24, line 107: Do you also need to change MergeSelect or just modify the scanner to get it work with Projection? Like the selects you merged might have different projection labels and you need to fetch both and project only required labels for each select

logicalplan/set_projection_labels.go
fpetkovski: When it is `without`, it returns the hash over all labels, and the labels which are not in the `Labels` slice.

logicalplan/set_projection_labels.go
fpetkovski: Hmm this is a good question, we do not use `MergeSelect` anymore, but I think we should modify it to be a union of all selectors that it merges.

logicalplan/set_projection_labels.go
yeya24: Do you need to include series hash in the grouping label for without? So that it can be removed after the aggregation.

Or you have another place to remove series hash label?

fpetkovski: I'll need to take a look at the details, will post a bit later.

pedro-stanaka: I believe we dont need to include because the hash will be just an implementation detail of the storage. We use it for things like: partitioning, map series to iterators, etc.

yeya24: I think it is required as it affects the query correctness.

I am trying to use the optimizer with the query fuzzer but it seems it just fail on without queries if hash label is not excluded here.

If it requires certain behavior from the storage to remove the hash probably need to document it clearly to make sure there is no correctness problem.

fpetkovski: We include the hash label only when Include is true. We also have an explicit select for the series hash label for some binary joins as shown in tests.

yeya24: If we don't have hash label when without then how do we pass the duplicate labels check in the engine? Do we have to disable it.

I am trying to run some correctness tests and see some issues based on my setup. Maybe you can try to add a similar correctness test using a mock storage and see if there is any issue.

yeya24: I see. But I can still see some correctness issues after disabling duplicate check so maybe you can take a closer look

MichaHoffmann: Let's disable them here too - with a hash column we won't get issues though since we never drop it i think

PR #554, opened 2025-04-21
engine: remove counter suffix annotation

MichaHoffmann:
This commit removes the annotation that is added when a metric has a name that is not indicative of a counter. This annotation is mostly spam that the query author cannot act upon.


fpetkovski:
Is there an easy way to make annotations opt-in when creating an engine?

jbradleynh:
Just as a comment to keep this PR alive, I'd love to see this incorporated. The messaging around metric naming is not helpful and can frequently clutter up the user UI, especially unhelpful where the underlaying metric is in fact a real counter but just don't conform to the naming convention.

PR #556, opened 2025-04-28
Improve propagate matchers optimizer

yeya24:
This PR adds several improvements and bug fixes for the propagate matcher optimizer.

Improvements:
- Support vector matching with matching labels
- Support one-to-many and many-to-one matching
- Single duplicate matcher won't stop matcher propagation. Old logic stops as long as there is one duplicate matcher. 

Bugs:
- Metric name matcher should be included in the final list of matchers
- When making union of label matcher, properly merge label matcher. The old logic replaces `{instance="1"}` with `{instance!="2"}`, which results in wrong result

Others:
- More unit tests
- Promqlsmith based correctness test


fpetkovski (approved):
Thanks for the extensive fixes @yeya24 !

MichaHoffmann (approved):

PR #557, opened 2025-05-04
Add support for limit_ratio aggregation

### LINKED ISSUE #515: Implement limitk and limit_ratio aggregations
Quoting [PromQL docs](https://prometheus.io/docs/prometheus/latest/querying/operators/) here: 

```
limitk (sample n elements)
limit_ratio (sample elements with approximately 𝑟 ratio if 𝑟 > 0, and the complement of such samples if 𝑟 = -(1.0 - 𝑟))

...

limitk and limit_ratio also return a subset of the input samples, including the original labels in the result vector, these are experimental operators that must be enabled with --enable-feature=promql-experimental-functions.

...

To sample 10 timeseries, for example to inspect labels and their values, we could write:

limitk(10, http_requests_total)
To deterministically sample approximately 10% of timeseries we could write:

limit_ratio(0.1, http_requests_total)
Given that limit_ratio() implements a deterministic sampling algorithm (based on labels' hash), you can get the complement of the above samples, i.e. approximately 90%, but precisely those not returned by limit_ratio(0.1, ...) with:

limit_ratio(-0.9, http_requests_total)
```

Note: They do not need to be implemented efficiently; but if we want to make use of distributed evaluation it suffices that we onboard them into the engine naively at first.

Saumya40-codes:
Closes #515 on engine side (following up #547)

This PR adds support for limit_ratio aggregation. `limit_ratio(r, metric` returns `r` ratio of the samples (where `0 <= r <=1`, though r can be negative, in such cases, the sample not in ratio  `1+r` will be returned

- Adds support for native histograms
- Enables distributed execution for limit_ratio
- Enables promqlAcceptance testing for `limitk/limit_ratio`

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2017,6 +2017,45 @@ sum by (grpc_method, grpc_code) (
 			step:  2 * time.Second,
 		},
 		{
+			name: "limit_ratio by (series)",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", series="1"} 1+2.1x50
+			    http_requests_total{pod="nginx-5", series="1"} 2+1.3x40
+			    http_requests_total{pod="nginx-3", series="2"} 5+3.4x50
+			    http_requests_total{pod="nginx-7", series="2"} 8.4+2.3x50
+			    http_requests_total{pod="nginx-4", series="2"} 2.5+2.3x50`,
+			query: `limit_ratio(0.3, http_requests_total) by (series)`,
+			start: time.Unix(0, 0),
+			end:   time.Unix(3000, 0),
+			step:  2 * time.Second,
+		},
+		{
+			name: "limit_ratio with ratio exceeding ratio limit", // in limit_ratio exceeded ratio limit should be capped to -1 or 1 (whichever is relatively closer)
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", series="1"} 1+2.1x50
+			    http_requests_total{pod="nginx-6", series="1"} 2+1.3x40
+			    http_requests_total{pod="nginx-4", series="2"} 5+3.4x30
+			    http_requests_total{pod="nginx-9", series="2"} 8.4+2.3x50
+			    http_requests_total{pod="nginx-2", series="2"} 2.5+2.3x50`,
+			query: `limit_ratio(23456, http_requests_total) and limit_ratio(-4567, http_requests_total)`,
+			start: time.Unix(0, 0),
+			end:   time.Unix(3000, 0),
+			step:  2 * time.Second,
+		},
+		{
+			name: "limit_ratio with NaN value as ratio param",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", series="1"} 1+2.1x50
+			    http_requests_total{pod="nginx-5", series="1"} 2+1.3x40
+			    http_requests_total{pod="nginx-3", series="2"} 5+3.4x50
+			    http_requests_total{pod="nginx-7", series="2"} 8.4+2.3x50
+			    http_requests_total{pod="nginx-4", series="2"} 2.5+2.3x50`,
+			query: `limit_ratio(NaN, http_requests_total)`,
+			start: time.Unix(0, 0),
+			end:   time.Unix(3000, 0),
+			step:  2 * time.Second,
+		},
+		{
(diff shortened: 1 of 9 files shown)


Saumya40-codes:
Implementation looks to be right to me now 👀 I'll add more necessary test, we can now do acceptance test now though for limitk/limit_ratio against testdata in prometheus

execution/aggregate/khashaggregate.go
```diff
@@ -293,10 +320,30 @@ func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampl
...
+			}
+		}
 	}
 
 	s := a.vectorPool.GetStepVector(t)
 	for _, sampleHeap := range a.heaps {
+		// for topk and bottomk the heap keeps the lowest value on top, so reverse it.
+		if (a.aggregation == parser.TOPK || a.aggregation == parser.BOTTOMK) && len(sampleHeap.entries) > 1 {
+			sort.Sort(sort.Reverse(sampleHeap))
+		}
```
Saumya40-codes, line 346: I was so wrong for this in https://github.com/thanos-io/promql-engine/pull/547#discussion_r2030410865 🤦🏽

Thankfully such case (or wrong result) only arise when we use limitk and one of the [testcase](https://github.com/thanos-io/promql-engine/pull/557/files#diff-cbf9ce27965b15f1e13be995f164fd6b33a19381364c4be3e8c047d528994495R3725)  added detects it

engine/engine_test.go
```diff
@@ -2004,6 +2003,71 @@ sum by (grpc_method, grpc_code) (
...
+		},
+		{
+			name: "combined kaggregates",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", series="1"} 1+2.1x50
+			    http_requests_total{pod="nginx-5", series="1"} 2+1.3x50
+			    http_requests_total{pod="nginx-3", series="2"} 5+3.4x50
+			    http_requests_total{pod="nginx-7", series="2"} 8.4+2.3x50
+			    http_requests_total{pod="nginx-4", series="4"} 2.5+2.3x50`,
+			query: `limitk(5, topk(3, limit_ratio(0.8, http_requests_total)) or bottomk(3, limit_ratio(-0.2, http_requests_total)))`,
```
Saumya40-codes, line 2066: I've capped limit here to 5 (max possible), turns out for range query the o/p (of limitk) doesn't remain consistent on different test runs for this particular test for prometheus (and also for thanos)

execution/aggregate/khashaggregate.go
```diff
@@ -293,10 +320,30 @@ func (a *kAggregate) aggregate(t int64, result *[]model.StepVector, k int, sampl
...
+			if addRatioSample(ratio, a.series[histId]) {
+				heap.Push(sampleHeap, &entry{histId: histId, histogramSample: histograms[i]})
+			}
+		}
 	}
 
 	s := a.vectorPool.GetStepVector(t)
 	for _, sampleHeap := range a.heaps {
+		// for topk and bottomk the heap keeps the lowest value on top, so reverse it.
+		if (a.aggregation == parser.TOPK || a.aggregation == parser.BOTTOMK) && len(sampleHeap.entries) > 1 {
```
GiedriusS, line 344: Very small nit: maybe we can just call `sort.Sort(...)` without checking `len(sampleHeap.entries) > 1`? If there's only 1 or 0 elements, then it will do nothing either way.

GiedriusS (approved):
:rocket:

Saumya40-codes:
The limitk part consumes much of the space all while figuring out which sample should be picked, maybe that logic can be abstract out where we can call something like `nextSample` for limitk which will return the next sample and all we have to do is to push it to the heap 👀

Saumya40-codes:
> The limitk part consumes much of the space all while figuring out which sample should be picked, maybe that logic can be abstract out where we can call something like `nextSample` for limitk which will return the next sample and all we have to do is to push it to the heap 👀

Perhaps doing something like: https://github.com/prometheus/prometheus/blob/main/promql/engine.go#L3283

Where we can get either float or histogram sample and I think we *can* do sampleId = sampleId[1:] and same for histogram, and all we have to do is to compare the first element everytime 👀

saswatamcode (approved):

PR #561, opened 2025-05-13
Fix fuzz failure counting

harry671003:
Fixing a bug introduced in #506 which commented out the increment of failures when results mismatched.

There was also a fuzz test failure which was previously missed because of the above bug.

```
fuzz: elapsed: 5s, execs: 219 (40/sec), new interesting: 123 (total: 135)
--- FAIL: FuzzEnginePromQLSmithInstantQuery (5.14s)
    --- FAIL: FuzzEnginePromQLSmithInstantQuery (0.10s)
        enginefuzz_test.go:274: load 30s
            			http_requests_total{pod="nginx-1", route="/"} 41.00+0.20x40
            			http_requests_total{pod="nginx-2", route="/"}  1+2.41x40
        enginefuzz_test.go:276: query: absent_over_time({__name__="http_requests_total",route=""}[4m]), start: 170000, end: 170000, interval: 0s
        enginefuzz_test.go:282: case 89 error mismatch.
            new result: {route=""} => 1 @[170000]
            old result: {} => 1 @[170000]
        enginefuzz_test.go:297: failed 1 test cases
    
    Failing input written to testdata/fuzz/FuzzEnginePromQLSmithInstantQuery/3d678e498ce0d8bd
    To re-run:
    go test -run=FuzzEnginePromQLSmithInstantQuery/3d678e498ce0d8bd
```

Here the problem is that Prometheus uses labels.Builder() to create series for absent. 
See: https://github.com/prometheus/prometheus/blob/df1b4da348a7c2f8c0b294ffa1f05db5f6641278/promql/functions.go#L1858

labels.Builder will treat route="" as missing label and will not output that. For consistency, Thanos engine should also use labels.Builder().

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2207,6 +2207,15 @@ avg by (storage_info) (
 			    http_requests_total{pod="nginx-1", series="1"} 1+2x40`,
 			query: `rate(http_requests_total[20s:10s] offset 20s)`,
 		},
+		{
+			name: "fuzz absent_over_time empty matcher",
+			load: `load 30s
+			    http_requests_total{pod="nginx-1", route="/"} 41.00+0.20x40
+			    http_requests_total{pod="nginx-2", route="/"}  1+2.41x40`,
+			query: `absent_over_time({__name__="http_requests_total",route=""}[4m])`,
+			start: time.UnixMilli(170000),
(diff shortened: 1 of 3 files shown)


fpetkovski (approved):
Thanks for the fix Harry

PR #563, opened 2025-05-18
execution: Utilize hints to limit the number of samples fetch when using limitk

Saumya40-codes:
This PR follows up #547 after adding liimitk aggregation

This is based on -> https://github.com/thanos-io/promql-engine/pull/547#issuecomment-2800094704

In limitk we take the first `k` samples of a given metric, which gets useful when you want to inspect that metric but just executing `metric_name` becomes costly as it might be of high cardinality. limitk limits the amount of samples in such case.

In current implementation we fetch all the samples based on vector selector, only take first `k`.  We can utilize hints we set under certain conditions, to only fetch required samples (which gets taken care during q.Select() on thanos side)

## Limitations

If query is something like `limitk(2, topk(5, some_metric))` then setting limits will give wrong results, additionally we should also refrain from doing this when a grouping (by/without) is used.

### DIFF
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -18,6 +18,7 @@ package execution
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


MichaHoffmann:
The contract in the storage layer would be "if func=limitk ~ only fetch hints.limit samples" right?

Saumya40-codes:
> The contract in the storage layer would be "if func=limitk ~ only fetch hints.limit samples" right?

If hints.Limit is 0 it doesn't put any restriction (for no. of samples to fetch), and here we only setting hints for limitk, so should we have that check at storage layer in thanos?

execution/execution.go
```diff
@@ -54,6 +55,15 @@ func New(ctx context.Context, expr logicalplan.Node, storage storage.Scanners, o
 }
 
 func newOperator(ctx context.Context, expr logicalplan.Node, storage storage.Scanners, opts *query.Options, hints promstorage.SelectHints) (model.VectorOperator, error) {
+	if hints.Limit != 0 {
+		switch expr.(type) {
+		case *logicalplan.Aggregation, *logicalplan.Binary:
+			hints.Limit = 0
```
yeya24, line 61: I wonder if it is the right thing to use Limit in hints to limit samples as that was introduced to limit number of series.
Shouldn't we introduce another hint for it?

PR #569, opened 2025-05-21
plan,engine: add partition labels

MichaHoffmann:
Add a "PartitionLabelSets" method to remote engine api. This makes sense since a remote engine can have a big labelset but we only need a subset of those to actually perform a partition of all engines. Operating on a smaller labelset there helps us pushing down more expressions, since we know if we lose one of the unimportant labels from the labelset we can still keep pushing down.

### DIFF
--- a/api/remote.go
+++ b/api/remote.go
@@ -23,7 +23,16 @@ type RemoteEndpoints interface {
 type RemoteEngine interface {
 	MaxT() int64
 	MinT() int64
+
+	// The external labels of the remote engine. These are used to limit fanout. The engine uses these to
+	// not distribute into remote engines that would return empty responses because their labelset is not matching.
 	LabelSets() []labels.Labels
+
+	// The external labels of the remote engine that form a logical partition. This is expected to be
+	// a subset of the result of "LabelSets()". The engine uses these to compute how to distribute a query.
+	// It is important that, for a given set of remote engines, these labels do not overlap meaningfully.
+	PartitionLabelSets() []labels.Labels
+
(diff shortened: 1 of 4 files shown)


engine/distributed.go
```diff
@@ -45,6 +45,10 @@ func (l remoteEngine) LabelSets() []labels.Labels {
 	return l.labelSets
 }
 
+func (l remoteEngine) PartitionLabelSets() []labels.Labels {
+	return l.labelSets
```
MichaHoffmann, line 49: This is a safe default. Engines who want to specify different partition labels can implement RemoteEngine differently.

api/remote.go
```diff
@@ -23,7 +23,14 @@ type RemoteEndpoints interface {
 type RemoteEngine interface {
 	MaxT() int64
 	MinT() int64
+
+	// The external labels of the remote engine.
```
fpetkovski, line 27: Maybe we can elaborate a bit more what these two label sets can be used for?

MichaHoffmann: added a bit more commentary

logicalplan/distribute_test.go
```diff
@@ -57,7 +57,7 @@ func TestDistributedExecution(t *testing.T) {
 			expr: `sum by (pod) (rate(http_requests_total[5m]))`,
 			expected: `
 sum by (pod) (dedup(
-  remote(sum by (pod, region) (rate(http_requests_total[5m]))), 
+  remote(sum by (pod, region) (rate(http_requests_total[5m]))),
```
harry671003, line 60: Should we add a test case where the PartitionLabelSets() and LabelSets() are different?

MichaHoffmann: done!

fpetkovski (approved):

harry671003 (approved):

PR #577, opened 2025-05-27
Fix NH unsupported functions

harry671003:
stddev_over_time, stdvar_over_time, quantile_over_time, max_over_time, min_over_time etc ignores histogram samples. 
For NH samples they must return empty result.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2251,11 +2251,39 @@ avg by (storage_info) (
 			step:  0,
 		},
 		{
-			name: "native histogram unsupported functions",
+			name: "native histogram stddev_over_time unsupported",
+			load: `load 2m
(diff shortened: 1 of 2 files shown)


engine/engine_test.go
```diff
@@ -2240,6 +2240,13 @@ avg by (storage_info) (
 			    http_requests_total{"label:name"="labelvalue"} 1+2x40`,
 			query: `rate(http_requests_total{"label:name"="labelvalue"}[20s:10s] offset 20s)`,
 		},
+		{
+			name: "native histogram unsupported functions",
```
yeya24, line 2244: Can we be more specific about the function?
Also let's add test cases for other functions we changed in this PR?

yeya24 (approved):
Thanks!

harry671003:
I will close this PR as @GiedriusS  pointed out it won't be needed after we update Prometheus.

PR #580, opened 2025-06-04
Add series count to operator telemetry

yeya24:


execution/function/scalar.go
```diff
@@ -42,7 +42,7 @@ func (o *scalarOperator) Explain() (next []model.VectorOperator) {
 func (o *scalarOperator) Series(ctx context.Context) ([]labels.Labels, error) {
 	start := time.Now()
 	defer func() { o.AddExecutionTimeTaken(time.Since(start)) }()
-
+	o.SetSeriesCount(int64(0))
```
GiedriusS: 0 is the zero value for int64 so we shouldn't need this?

engine/explain_test.go
```diff
@@ -255,7 +293,7 @@ func renderAnalysisTree(node *engine.AnalyzeOutputNode, level int) string {
 		result.WriteString(strings.Repeat("|   ", level-1) + "|---")
 	}
 
-	result.WriteString(fmt.Sprintf("%s: %d peak: %d\n", node.OperatorTelemetry.String(), totalSamples, peakSamples))
+	result.WriteString(fmt.Sprintf("%s: series: %d total_samples: %d peak_samples: %d\n", node.OperatorTelemetry.String(), seriesCount, totalSamples, peakSamples))
```
MichaHoffmann: max_series maybe? since sometimes its just an upper bound ( though.. only really for "or" maybe we should use "series" normally and if its binary many-to-many we can use "max_series" ?

engine/explain_test.go
```diff
@@ -122,31 +122,66 @@ func assertExecutionTimeNonZero(t *testing.T, got *engine.AnalyzeOutputNode) boo
...
+}
+
 func TestQueryAnalyze(t *testing.T) {
 	opts := promql.EngineOpts{Timeout: 1 * time.Hour}
-	series := storage.MockSeries(
-		[]int64{240, 270, 300, 600, 630, 660},
-		[]float64{1, 2, 3, 4, 5, 6},
-		[]string{labels.MetricName, "foo"},
-	)
+	serieses := []storage.Series{
```
pedro-stanaka: nit: seriesList maybe?

engine/explain_test.go
yeya24: This is just the code we use in the unit test to compare.
I can change the interface to be `MaxSeriesCount` and `SetMaxSeriesCount`. In the actual UX like Thanos UI or API, Idk if we want to call it differently for each operator. Maybe just call it `max_series` makes more sense there

yeya24: I feel `MaxSeriesCount` and `SetMaxSeriesCount` can be also confusing though.

engine/explain_test.go
yeya24: Done

execution/function/scalar.go
yeya24: Done

engine/explain_test.go
MichaHoffmann: Ah yes this is only for test nvm then let's keep it as is!

MichaHoffmann (approved):

PR #583, opened 2025-06-05
engine: fix limitratio to limit the map access and update sort logic 

Saumya40-codes:
limit the number of map access calls in limit_ratio, as it won't be needed to check whether to add samples or not.

For instant queries:
limitk, limit_ratio doesn't require to sort values at the end.

### DIFF
--- a/engine/sort.go
+++ b/engine/sort.go
@@ -101,8 +101,10 @@ func valueCompare(order sortOrder, l, r float64) bool {
 	}
 	if order == sortOrderAsc {
 		return l < r
+	} else if order == sortOrderDesc {
+		return l > r
 	}
-	return l > r
+	return false
 }
 
 func (s sortFuncResultSort) comparer(samples *promql.Vector) func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


engine/sort.go
```diff
@@ -155,6 +156,10 @@ func (s aggregateResultSort) comparer(samples *promql.Vector) func(i, j int) boo
 		if lblsCmp != 0 {
 			return lblsCmp < 0
 		}
+
+		if s.sortOrder == noValueSort {
```
GiedriusS: Where is this used? It looks very weird. Maybe we can just avoid calling this function (& sorting) overall?

Saumya40-codes: https://github.com/thanos-io/promql-engine/blob/main/engine/sort.go#L100C1-L104C5 for limit k/_ratio, we set sortOrder to noValueSort.

Of what I saw, we would need sort if some grouping is used, though sorting sample values would be just extra step as its not needed in two of these aggregations.

> Maybe we can just avoid calling this function (& sorting) overall?

Right, but is it ok if order in which samples are, differs? this is just for when some groupings are used for [instant queries](https://github.com/thanos-io/promql-engine/blob/main/engine/engine.go#L253)

PR #584, opened 2025-06-05
Keep track of series and next execution time

yeya24:
Instead of tracking a single execution time for the operator, separate series and next stage and keep track their execution time.
This can be useful to understand operator performance.

### DIFF
--- a/execution/telemetry/telemetry.go
+++ b/execution/telemetry/telemetry.go
@@ -20,8 +20,11 @@ type OperatorTelemetry interface {
 
 	MaxSeriesCount() int64
 	SetMaxSeriesCount(count int64)
-	AddExecutionTimeTaken(time.Duration)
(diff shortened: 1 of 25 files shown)


execution/telemetry/telemetry.go
```diff
@@ -106,6 +123,24 @@ func (ti *TrackedTelemetry) ExecutionTimeTaken() time.Duration {
 	return ti.ExecutionTime
```
harry671003, line 123: return ti.NextTime + ti.SeriesTime?

MichaHoffmann (approved):

execution/telemetry/telemetry.go
yeya24: We still have `ExecutionTime` so I return that.
We can also remove it and return `ti.NextTime + ti.SeriesTime`

PR #587, opened 2025-06-08
all: make coalesce a logical node

MichaHoffmann:
Making "coalesce" a logical node, allows us to move it around in the execution tree and eventually coalesce later and evaluate more of the query concurrently.
This will allow us to subsequently treat local execution almost like remote execution, so that we can compute aggregations in shards.

### DIFF
--- a/execution/function/absent.go
+++ b/execution/function/absent.go
@@ -57,13 +57,19 @@ func (o *absentOperator) loadSeries() {
 
 		// https://github.com/prometheus/prometheus/blob/df1b4da348a7c2f8c0b294ffa1f05db5f6641278/promql/functions.go#L1857
 		var lm []*labels.Matcher
-		switch n := o.funcExpr.Args[0].(type) {
-		case *logicalplan.VectorSelector:
-			lm = append(n.LabelMatchers, n.Filters...)
-		case *logicalplan.MatrixSelector:
-			v := n.VectorSelector
-			lm = append(v.LabelMatchers, v.Filters...)
-		default:
+		logicalplan.TraverseBottomUp(nil, &o.funcExpr.Args[0], func(parent, node *logicalplan.Node) bool {
+			switch n := (*node).(type) {
+			case *logicalplan.VectorSelector:
+				lm = append(n.LabelMatchers, n.Filters...)
+			case *logicalplan.CheckDuplicateLabels, *logicalplan.Coalesce, *logicalplan.MatrixSelector:
+			// ignore logical nodes between the absent function and the first vectorselector
+			default:
+				lm = lm[:0]
+			}
+			return false
+		})
+
+		if len(lm) == 0 {
 			o.series = []labels.Labels{labels.EmptyLabels()}
 			return
 		}
--- a/logicalplan/coalesce.go
+++ b/logicalplan/coalesce.go
@@ -1,0 +1,69 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package logicalplan
+
+import (
+	"github.com/prometheus/prometheus/util/annotations"
+
+	"github.com/thanos-io/promql-engine/query"
+)
+
+type CoalesceOptimizer struct{}
+
+func (c CoalesceOptimizer) Optimize(expr Node, opts *query.Options) (Node, annotations.Annotations) {
+	numShards := opts.NumShards()
+
+	TraverseBottomUp(nil, &expr, func(parent, e *Node) bool {
+		switch t := (*e).(type) {
+		case *VectorSelector:
+			if parent != nil {
+				// we coalesce matrix selectors in a different branch
+				if _, ok := (*parent).(*MatrixSelector); ok {
+					return false
+				}
+			}
+			exprs := make([]Node, numShards)
+			for i := range numShards {
+				vs := t.Clone().(*VectorSelector)
+				vs.Shard = i
+				vs.NumShards = numShards
+				exprs[i] = vs
+			}
+			*e = &Coalesce{Exprs: exprs}
+		case *MatrixSelector:
+			// handled in *parser.Call branch
+			return false
+		case *FunctionCall:
+			// non-recursively handled in execution.go
+			if t.Func.Name == "absent_over_time" {
+				return true
+			}
+			var (
+				ms   *MatrixSelector
+				marg int
+			)
+			for i := range t.Args {
+				if arg, ok := t.Args[i].(*MatrixSelector); ok {
+					ms = arg
+					marg = i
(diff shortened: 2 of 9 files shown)


logicalplan/plan.go
```diff
@@ -25,6 +25,7 @@ var DefaultOptimizers = []Optimizer{
 	SortMatchers{},
 	MergeSelectsOptimizer{},
 	DetectHistogramStatsOptimizer{},
+	CoalesceOptimizer{},
```
yeya24, line 28: Should we make this opt-in first?

MichaHoffmann: If we do then we don't decide concurrently anymore - maybe it should just be a generic engine pass - we probably never don't want it

MichaHoffmann: Decode*

logicalplan/coalesce.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+			exprs := make([]Node, numShards)
+			for i := range numShards {
+				vs := t.Clone().(*VectorSelector)
+				vs.Shard = i
+				vs.NumShards = numShards
+				exprs[i] = vs
+			}
+			*e = &Coalesce{Exprs: exprs}
+		case *MatrixSelector:
```
yeya24, line 34: Do we plan to handle matrix selector without function?

logicalplan/coalesce.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (c CoalesceOptimizer) Optimize(expr Node, opts *query.Options) (Node, annotations.Annotations) {
+	numShards := opts.NumShards()
+
+	TraverseBottomUp(nil, &expr, func(parent, e *Node) bool {
+		switch t := (*e).(type) {
+		case *VectorSelector:
+			if parent != nil {
+				// we coalesce matrix selectors in a different branch
+				if _, ok := (*parent).(*MatrixSelector); ok {
```
yeya24, line 22: How about subqueries?

logicalplan/logical_nodes.go
```diff
@@ -570,3 +580,33 @@ func isAvgAggregation(expr *Node) bool {
 	}
 	return false
 }
+
+type Coalesce struct {
+	// We assume to always have at least one expression
+	Exprs []Node `json:"-"`
+}
+
+func (c *Coalesce) ReturnType() parser.ValueType { return parser.ValueTypeVector }
```
yeya24, line 589: Maybe we should do Exprs[0].ReturnType()?

yeya24 (approved):
Thanks. This is epic!
Just some small questions

query/options.go
```diff
@@ -20,6 +20,10 @@ type Options struct {
 	DecodingConcurrency      int
 }
 
+func (o *Options) NumShards() int {
+	return max(o.DecodingConcurrency, 1)
+}
```
yeya24, line 25: I am thinking this shard size could be different for different operators? I wonder if the default value works for all operators

logicalplan/coalesce.go
harry671003: Plain matrix selector is unsupported by Thanos engine.

logicalplan/coalesce.go
```diff
@@ -0,0 +1,69 @@
...
+
+package logicalplan
+
+import (
+	"github.com/prometheus/prometheus/util/annotations"
+
+	"github.com/thanos-io/promql-engine/query"
+)
+
+type CoalesceOptimizer struct{}
```
harry671003, line 12: Should the optimizer be called something like ConcurrentDecodeOptimizer?

query/options.go
harry671003: Default is numCores / 2?

yeya24: Yeah I meant if different operators should use the same concurrency

fpetkovski:
This change potentially modifies the contract with the scanner. I would like to test it with our downstream project this week before we merge it.

execution/function/absent.go
```diff
@@ -57,13 +57,19 @@ func (o *absentOperator) loadSeries() {
...
 		// https://github.com/prometheus/prometheus/blob/df1b4da348a7c2f8c0b294ffa1f05db5f6641278/promql/functions.go#L1857
 		var lm []*labels.Matcher
-		switch n := o.funcExpr.Args[0].(type) {
-		case *logicalplan.VectorSelector:
-			lm = append(n.LabelMatchers, n.Filters...)
-		case *logicalplan.MatrixSelector:
-			v := n.VectorSelector
-			lm = append(v.LabelMatchers, v.Filters...)
-		default:
+		logicalplan.TraverseBottomUp(nil, &o.funcExpr.Args[0], func(parent, node *logicalplan.Node) bool {
```
fpetkovski, line 60: Shouldn't this be in the logical plan?

logicalplan/coalesce.go
fpetkovski: I think the idea is to allow for things like 
```
sum(
  coalesce(
    sum(shard_a), 
    sum(shard_b)
  )
)
```
down the line.

logicalplan/codec.go
```diff
@@ -219,6 +219,19 @@ func unmarshalNode(data []byte) (Node, error) {
 			return nil, err
 		}
 		return u, nil
+	case CoalesceNode:
```
fpetkovski, line 222: Would this cause remote execution to be planned in the central node?

storage/prometheus/scanners.go
```diff
@@ -126,28 +120,23 @@ func (p Scanners) NewMatrixSelector(
...
-		operators = append(operators, exchange.NewConcurrent(operator, 2, opts))
+	mat, err := NewMatrixSelector(
+		model.NewVectorPool(opts.StepsBatch),
+		selector,
+		call.Func.Name,
+		arg,
+		arg2,
+		opts,
+		logicalNode.Range,
+		vs.Offset,
```
fpetkovski, line 131: we could simplify this by passing in the entire `vs` object.

execution/function/absent.go
MichaHoffmann: The sector labels of the vector sector of the absent function? I think we could put it into the functioncall node and populate it in the plan yeah

logicalplan/coalesce.go
MichaHoffmann: Can parent of vectorselector be a subquery? I guess you could have `last_over_time(http_request_total[10m:5s])`.

logicalplan/coalesce.go
harry671003: Should a different optimizer should be responsible for sharding aggregations?

PR #591, opened 2025-06-11
execution: binary and memory tweaks

MichaHoffmann:
"and" can only have output series if lhs and rhs have a collision, this means we can get away with only initializing the output map for lhs series that have coresponding rhs collisions.


yeya24 (approved):
Thanks!

PR #592, opened 2025-06-12
execution: add tmin/tmax_over_time functions

MichaHoffmann:
We add the tmin/tmax_over_time range functions. They give the timestamp of the last sample that had the maximum/minimum value in the range window.
Additionally we enable ext functions by default. ExtLookback was already a engine parameter and guarding by another conditional seemed like weird API.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,7 +29,6 @@ linters:
     - godot
     - gofmt
     - gci
-    - gofmt
     - gosimple
     - govet
     - ineffassign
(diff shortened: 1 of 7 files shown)


.golangci.yml
```diff
@@ -29,7 +29,6 @@ linters:
     - godot
     - gofmt
     - gci
-    - gofmt
```
MichaHoffmann, line 32: duplicate

MichaHoffmann:
> Does it make sense to propose timestamp functions in Prometheus?

https://github.com/prometheus/prometheus/pull/16722 !

MichaHoffmann:
> Do we need these functions as X if they get added to Prometheus?

mh, maybe not; but they are behind experimental flag at least

MichaHoffmann:
Those functions will come when we bump prometheus, closing

PR #594, opened 2025-06-13
all: bump to latest prometheus

MichaHoffmann:
Bump to latest prometheus and fix fallout. Also made XFunctions always enabled.


engine/engine.go
```diff
@@ -71,10 +70,6 @@ type Opts struct {
 	// SelectorBatchSize specifies the maximum number of samples to be returned by selectors in a single batch.
 	SelectorBatchSize int64
 
-	// EnableXFunctions enables custom xRate, xIncrease and xDelta functions.
-	// This will default to false.
-	EnableXFunctions bool
```
yeya24: Can we bring back this feature flag? This can impact downstream users of the engine as not everyone wants to enable X functions

MichaHoffmann: I think engine is wrong place to restrict functions - if we implement it here and make it configurable through extlookback it's weird to have another flag; on the other hand down stream can do much fancier blocking of expressions by looking at ast.

GiedriusS: Should downstream code know about all custom functions?

MichaHoffmann: We still expose the map, we can just register it if client wants experimental functions from Prometheus too

MichaHoffmann: We can take this change out of this PR though

MichaHoffmann: flag is back!

MichaHoffmann:
Fuzzing fails because of: https://cloud-native.slack.com/archives/C01AUBA4PFE/p1749978859618789

execution/function/histogram.go
```diff
@@ -208,6 +281,7 @@ func (o *histogramOperator) loadSeries(ctx context.Context) error {
 	)
 
 	o.series = make([]labels.Labels, 0)
+	o.inputSeriesNames = make([]string, len(series))
```
MichaHoffmann, line 284: so this is kinda gross, but needed to pass the promql acceptance tests since the series name is part of the expected warning.

harry671003:
Would you be taking another look at the upgrade now? Looks like the upstream blockers are merged now.

I can help if you aren't actively looking into this now.

PR #595, opened 2025-06-17
Fix sort for native histogram

harry671003:
Sort function in Prometheus engine removes histogram samples for both range and instant queries.
See: https://github.com/prometheus/prometheus/blob/cfa922e67734594a6438e6b3cb063a66196ba71e/promql/functions.go#L505

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2260,6 +2260,26 @@ avg by (storage_info) (
 			end:   time.UnixMilli(0),
 			step:  0,
 		},
+		{
+			name: "native histogram sort",
+			load: `load 2m
+			    http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20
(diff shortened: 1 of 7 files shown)


MichaHoffmann:
I wonder why the sorting function decides if we include histograms or not

MichaHoffmann (approved):

fpetkovski:
PromQL gets weirder by the day.

harry671003:
Don't merge this yet. Found another case where the new engine returns wrong results when there is an inner sort() which filters our the NH samples.

```
           (
                histogram_fraction(
                  scalar(-{__name__="http_request_duration_seconds"} @ 57486872.643),
                  scalar(max without (pod) ({__name__="http_request_duration_seconds"})),
                  (
                      sort({__name__="http_request_duration_seconds"} @ end())
                    unless
                      -{__name__="http_request_duration_seconds"} @ 1490240032.762
                  )
                )
              or
                (
                    -count without () ({__name__="http_request_duration_seconds"})
                  +
                    {__name__="http_request_duration_seconds",pod!~"ngi.*"} offset -4m56s
                )
            )
```

MichaHoffmann:
Let's exclude sort from the fuzz tests and document the difference to PromQL

harry671003:
https://github.com/thanos-io/promql-engine/pull/600 will exclude sort() and it's family functions from native histogram fuzz tests. I'll raise a separate PR to update the README.

PR #596, opened 2025-06-18
Fix output series ID for native histogram binary arithmetic

harry671003:
### Notes
The seriesIDs for binary arithmetic involving native histograms are not computed correctly.

I'll admit I'm not entirely confident about the implementation - I mostly mirrored the logic from float arithmetic.

Example query:

```
avg({__name__="http_request_duration_seconds"})
    or
(
    {__name__="http_request_duration_seconds"}
    +
    {__name__="http_request_duration_seconds", pod!="nginx-1"}
)
```

Without the fix, the test case in the PR fails with an `index out of range` error during the `or` operation. 
```
Diff:
        --- Expected
        +++ Actual
        @@ -1,7 +1,2 @@
        -(*promql.Result)({} =>
        -{count:2.333333333333333, sum:14, (0.5,1]:1.6666666666666665, (1,2]:0.6666666666666667} @[0]
        -{pod="nginx-2"} =>
        -{count:4, sum:28, (0.5,1]:4} @[0]
        -{pod="nginx-3"} =>
        -{count:4, sum:28, (0.5,1]:4} @[0])
        +(*promql.Result)(unexpected error: runtime error: index out of range [2] with length 2)
```

The root cause is that the `+` operation assigns an incorrect ID to the histogram output, which breaks subsequent operations like `or`.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2270,6 +2270,38 @@ avg by (storage_info) (
 			end:   time.UnixMilli(300000),
 			step:  15 * time.Second,
 		},
+		{
+			name: "native histogram nested binary#1",
+			load: `load 2m
+			    http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20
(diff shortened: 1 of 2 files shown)


MichaHoffmann:
Oh yeah using the id in join bucket only makes sense for and and unless - we could use it too for one to one but maybe as optimization later

yeya24 (approved):

PR #597, opened 2025-06-19
Compact histograms after scalar operations

harry671003:
Prometheus engine compacts histograms after scalar operations involving them to remove the empty buckets.

See: https://github.com/prometheus/prometheus/blob/69906bb4f5f9e62255bced373c56fc13a3f61093/promql/engine.go#L2951

We should do the same in Thanos engine.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2260,6 +2260,16 @@ avg by (storage_info) (
 			end:   time.UnixMilli(0),
 			step:  0,
 		},
+		{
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
Thanks! Do we have those test cases covered in the existing fuzzy test?

harry671003:
> Do we have those test cases covered in the existing fuzzy test?

Yes, they were discovered from the fuzzy tests. The tests are not stable and so is disabled in main.

PR #598, opened 2025-06-20
Implement timestamp() for native histograms

harry671003:
### Notes
- Implementing timestamp() for native histograms
- Timestamp function should return a float sample with t=T, v=T/1000 for both native histograms and floats

From the docs: https://prometheus.io/docs/prometheus/latest/querying/functions/#timestamp
> **timestamp()**
> timestamp(v instant-vector) returns the timestamp of each of the samples of the given vector as the number of seconds since January 1, 1970 UTC. It acts on float and histogram samples in the same way.

### DIFF
--- a/execution/function/timestamp.go
+++ b/execution/function/timestamp.go
@@ -16,14 +16,16 @@ import (
 )
 
 type timestampOperator struct {
+	pool *model.VectorPool
 	next model.VectorOperator
 
 	series []labels.Labels
 	once   sync.Once
 }
 
-func newTimestampOperator(next model.VectorOperator, opts *query.Options) model.VectorOperator {
+func newTimestampOperator(pool *model.VectorPool, next model.VectorOperator, opts *query.Options) model.VectorOperator {
 	oper := &timestampOperator{
+		pool: pool,
 		next: next,
 	}
 	return telemetry.NewOperator(telemetry.NewTelemetry(oper, opts), oper)
@@ -65,7 +67,7 @@ func (o *timestampOperator) loadSeries(ctx context.Context) error {
 }
 
 func (o *timestampOperator) GetPool() *model.VectorPool {
-	return o.next.GetPool()
+	return o.pool
 }
 
 func (o *timestampOperator) Next(ctx context.Context) ([]model.StepVector, error) {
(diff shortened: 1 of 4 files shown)


execution/function/timestamp.go
```diff
@@ -76,13 +78,34 @@ func (o *timestampOperator) Next(ctx context.Context) ([]model.StepVector, error
...
 		return nil, err
 	}
+	if len(in) == 0 {
+		return nil, nil
+	}
+	result := o.pool.GetVectorBatch()
 	for _, vector := range in {
-		for i := range vector.Samples {
-			vector.Samples[i] = float64(vector.T / 1000)
+		out := o.pool.GetStepVector(vector.T)
```
fpetkovski, line 90: Why do we need these changes? Can't apply the same pattern to histograms as we do for floats?

execution/function/timestamp.go
```diff
@@ -76,13 +78,34 @@ func (o *timestampOperator) Next(ctx context.Context) ([]model.StepVector, error
...
+		value := float64(vector.T / 1000)
+
+		sampleTimestamps := make([]float64, len(vector.Samples))
```
fpetkovski: I'm not sure it makes sense to do direct allocation if we're injecting a pool

execution/function/timestamp.go
harry671003: When we have both samples and histograms in the input StepVector, we need to return a StepVector with only Samples populated.

Eg:
```
load 30s
  http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:3 sum:14.00 buckets:[1 2]}}+{{schema:0 count:4 buckets:[1 2 1]}}x20
  http_request_duration_seconds{pod="nginx-2"} 1x10 {{schema:0 count:2 sum:14.00 buckets:[2]}}+{{schema:0 count:6 buckets:[2 2 2]}}x10`
```

I'm not sure if we should do this in place.

```
for _, vector := range in {
		for i := range vector.Samples {
			vector.Samples[i] = float64(vector.T / 1000)
		}
		for i := range vector.Histograms {
			vector.Samples = append(vector.Samples, float64(vector.T)/1000)
			vector.SampleIDs = append(vector.SampleIDs, vector.HistogramIDs[i])
		}
		vector.Histograms = nil
		vector.HistogramIDs = nil
	}
}
```

harry671003:
Could you take another look @fpetkovski?

fpetkovski:
@harry671003 would you mind rebasing this so we can merge it?

PR #600, opened 2025-06-23
Test: Enable fuzz tests for native histograms

harry671003:
### Changes
This PR mainly enables native histogram fuzz tests. It also addresses some of the issue identified while testing.

In addition it makes the following changes:
- Excluding duplicate label set errors from test comparison
  - Duplicate label check in Thanos engine doesn't always work similar to Prometheus engine.
- Removing known optimizations from sample stats checks
  -  Thanos engine will output less samples
- Fixing the buckets for NH load to avoid zero buckets.
  -  Histogram.Equals() doesn't equate histograms with zero values.
- Compare histograms approximately.
- Allow fractional errors while comparing floats

#### Excluding known edge cases
Excluding some known edge cases from native histgram fuzz tests such as:
- sort() and other family functions. Thanos engine doesn't filter out NH samples. Prometheus does.
- predict_linear with step invariant - Thanos engine is unable to handle this case today.
- timestamp() - Fix in #598

#### Race condition in Prometheus engine
- https://github.com/thanos-io/promql-engine/actions/runs/16230914496/job/45833407585
- There appears to be some race condition in old engine when the storage is shared across multiple runs.
- This causes the Prometheus engine to not return the buckets.
- This is not reproducible outside the fuzz tests and unable to find the root cause.
```
        enginefuzz_test.go:341: case 31 error mismatch.
            new result: {pod="nginx-1"} =>
            {count:66, sum:0, (0.5,1]:1, (1,2]:2, (2,4]:63} @[0]
            {count:66, sum:0, (0.5,1]:1, (1,2]:2, (2,4]:63} @[4000]
            
        old result: {pod="nginx-1"} =>
            {count:66, sum:0} @[0]
            {count:66, sum:0} @[4000]
            {count:66, sum:0} @[8000]    
```

Fix: https://github.com/prometheus/prometheus/pull/16879

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/thanos-io/promql-engine/engine"
 	"github.com/thanos-io/promql-engine/execution/model"
 	"github.com/thanos-io/promql-engine/execution/warnings"
+	"github.com/thanos-io/promql-engine/extlabels"
 	"github.com/thanos-io/promql-engine/logicalplan"
 	"github.com/thanos-io/promql-engine/query"
 	"github.com/thanos-io/promql-engine/storage/prometheus"
@@ -2313,6 +2314,21 @@ or
 			start: time.UnixMilli(0),
 			end:   time.UnixMilli(124000),
 			step:  15 * time.Second,
+		},
+		{
+			name: "fuzz native histogram approx float comparison",
+			load: `load 2m
+			    http_request_duration_seconds{pod="nginx-1"} {{schema:0 count:30 sum:14.00 buckets:[27 2 1]}}+{{schema:0 count:30 buckets:[27 2 1]}}x20
+			    http_request_duration_seconds{pod="nginx-2"} {{schema:-2 count:58 sum:4368.00 buckets:[54 2 2]}}+{{schema:-2 count:58 buckets:[54 2 2]}}x30`,
+			query: `
+-(
+    -{__name__="http_request_duration_seconds"}
+  /
+    histogram_stdvar({__name__="http_request_duration_seconds"})
+)`,
+			start: time.UnixMilli(83000),
+			end:   time.UnixMilli(160000),
+			step:  time.Minute + 16*time.Second,
 		},
 	}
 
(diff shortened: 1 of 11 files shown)


engine/engine_test.go
```diff
@@ -6011,7 +6106,39 @@ var (
...
 			}
 
-			return l.Equals(r)
+			// Copied from https://github.com/prometheus/prometheus/blob/3d245e31d31774f62ff18c36039315fa55fe252c/model/histogram/float_histogram.go#L471
+			// and extended to use approx comparison instead of exact match.
+			if l.Schema != r.Schema || !compareFloats(l.Count, r.Count) || !compareFloats(l.Sum, r.Sum) {
+				return false
+			}
+
+			if l.UsesCustomBuckets() {
```
yeya24, line 6115: This is more a question from me. Do we support custom buckets in thanos engine today?

engine/enginefuzz_test.go
```diff
@@ -39,15 +40,28 @@ type testCase struct {
...
 		switch n := node.(type) {
 		case *parser.Call:
-			if n.Func.Name == "scalar" {
+			switch n.Func.Name {
+			case "scalar":
+				// Optimized to step_invariant in Thanos engine.
+				valid = false
+				return errors.New("error")
+			case "histogram_count", "histogram_sum", "histogram_avg":
+				// Optimized using DetectHistogramStatsOptimizer and will return smaller samples than Prometheus engine.
```
yeya24, line 64: IIUC Prometheus engine also has similar optimizations. Does it do anything differently that the samples are different

yeya24 (approved):
Thanks. This is great work.
I have some small questions and nothing blocks this PR

engine/engine_test.go
yeya24: And do we preserve those custom buckets in the NH samples as part of the query results. IIUC we have iterators for NH samples so custom buckets or not are abstracted away by that iterator

engine/enginefuzz_test.go
harry671003: Yes, you're right. I need to another look at why the samples are still different.

engine/engine_test.go
harry671003: I'm not sure if we support in this engine.
This code is directly copied from Prometheus with changes to support approx equality.
https://github.com/prometheus/prometheus/blob/9a0bbb60bc3eb68d045aae7535d34f4d02b959f1/model/histogram/float_histogram.go#L482

yeya24: Just curious if custom buckets will be part of the query results. All good.

engine/enginefuzz_test.go
harry671003: I took another look at this. The optimizer in Prometheus engine isn't as advanced as that used in Thanos engine. So it doesn't always use the optimized iterator.

I think we should fix it upstream. I don't have b/w right now. I'll create an issue upstream.
https://github.com/prometheus/prometheus/issues/16916

yeya24: Thanks. I am good to wait for upstream fix

Issue #573, opened 2025-05-27
Incorrect peak sample stats


yeya24:
Query using promql engine with 1h range and 20min step, got peak samples 82678.

```
curl 'http://localhost:10902/api/v1/query_range?query=sum(up)&start=1748296266.667&end=1748299866.667&step=1200&engine=prometheus&stats=all' | jq .

{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {},
        "values": [
          [
            1748296266.667,
            "5987"
          ],
          [
            1748297466.667,
            "5982"
          ],
          [
            1748298666.667,
            "6181"
          ],
          [
            1748299866.667,
            "6466"
          ]
        ]
      }
    ],
    "stats": {
      "timings": {
        "evalTotalTime": 4.260582792,
        "resultSortTime": 7.5e-07,
        "queryPreparationTime": 4.0209e-05,
        "innerEvalTime": 4.260507667,
        "execQueueTime": 0,
        "execTotalTime": 4.260590167
      },
      "samples": {
        "totalQueryableSamples": 82678,
        "peakSamples": 82678
      }
    },
    "analysis": {}
  }
}
```

Same using thanos engine, got peak samples 1. It doesn't need to match what promql engine has but 1 is also incorrect. Step batch size is 10 and peak samples must > 1 in the vector selector.

```
curl 'http://localhost:10902/api/v1/query_range?query=sum(up)&start=1748296266.667&end=1748299866.667&step=1200&engine=thanos&stats=all' | jq .

{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {},
        "values": [
          [
            1748296266.667,
            "5987"
          ],
          [
            1748297466.667,
            "5982"
          ],
          [
            1748298666.667,
            "6181"
          ],
          [
            1748299866.667,
            "6466"
          ]
        ]
      }
    ],
    "stats": {
      "timings": {
        "evalTotalTime": 0,
        "resultSortTime": 0,
        "queryPreparationTime": 0,
        "innerEvalTime": 0,
        "execQueueTime": 0,
        "execTotalTime": 0
      },
      "samples": {
        "totalQueryableSamples": 82678,
        "peakSamples": 1
      }
    },
    "analysis": {}
  }
}
```

[...]

PR #604, opened 2025-07-07
Fixing Peak Sample Count Per Series

Naman-B-Parlecha:
Resolve: #573

Currently, the peak sample count is incorrectly reported as 1 because `currStepSamples` is reset to 0 at every step iteration, preventing proper accumulation of peak samples across series.

To fix this we can maintain a vector to track sample count per step across all series and add up the sample at each step. Once all the all sample per step of all series are processed we can update the telemetry for all series

### DIFF
--- a/engine/explain_test.go
+++ b/engine/explain_test.go
@@ -371,7 +371,7 @@ func TestAnalyzPeak(t *testing.T) {
 	analyzeOutput = explainableQuery.Analyze()
 
 	t.Logf("value of peak = %v", analyzeOutput.PeakSamples())
-	require.Equal(t, int(200), analyzeOutput.PeakSamples())
+	require.Equal(t, int64(200), analyzeOutput.PeakSamples())
 
 	result := renderAnalysisTree(analyzeOutput, 0)
 	expected := `[duplicateLabelCheck]: max_series: 2 total_samples: 0 peak_samples: 0
--- a/engine/explain_test.go
+++ b/engine/explain_test.go
@@ -251,7 +251,6 @@ func TestQueryAnalyze(t *testing.T) {
 		}
 	}
 }
-
 func TestAnalyzeOutputNode_Samples(t *testing.T) {
 	t.Parallel()
 	ng := engine.New(engine.Opts{EngineOpts: promql.EngineOpts{Timeout: 1 * time.Hour}, EnableAnalysis: true, DecodingConcurrency: 2})
@@ -299,9 +298,9 @@ func TestAnalyzeOutputNode_Samples(t *testing.T) {
 |   |   |---[duplicateLabelCheck]: max_series: 2 total_samples: 0 peak_samples: 0
 |   |   |   |---[coalesce]: max_series: 2 total_samples: 0 peak_samples: 0
 |   |   |   |   |---[concurrent(buff=2)]: max_series: 1 total_samples: 0 peak_samples: 0
-|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 0 mod 2): max_series: 1 total_samples: 1010 peak_samples: 20
+|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 0 mod 2): max_series: 1 total_samples: 1010 peak_samples: 200
 |   |   |   |   |---[concurrent(buff=2)]: max_series: 1 total_samples: 0 peak_samples: 0
-|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 1 mod 2): max_series: 1 total_samples: 1010 peak_samples: 20
+|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 1 mod 2): max_series: 1 total_samples: 1010 peak_samples: 200
 `
 	require.EqualValues(t, expected, result)
 }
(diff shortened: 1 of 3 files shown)


yeya24:
I think it is better to do https://github.com/thanos-io/promql-engine/issues/573#issuecomment-2915226490.
We can decouple updating peak samples from `IncrementSamplesAtTimestamp`.

Naman-B-Parlecha:
Sure refactoring!

Naman-B-Parlecha:
Hey @harry671003 thanks for the clarification,
i m not entirely sure at what point we should be updating the peak but i believe we want to do something similar to what happens in prometheus? 

@yeya24 is more qualified to answer this i believe as he has worked closely with the engine

harry671003:
I thought a little bit more about this:

Let's define the Peak samples first. Peak samples for an operator is the total samples the operator processed in a single `Next()` call. Maximum Peak samples would be the peak samples across all operators.

One way to implement this is to use the telemetry operator
https://github.com/thanos-io/promql-engine/blob/main/execution/telemetry/telemetry.go#L204

```

func (t *Operator) Next(ctx context.Context) ([]model.StepVector, error) {
	start := time.Now()
	totalSamplesBefore := t.OperatorTelemetry.Samples().TotalSamples() // Get the samples before calling Next()
	defer func() { t.OperatorTelemetry.AddNextExecutionTime(time.Since(start)) }()
	out, err := t.inner.Next(ctx)
	if err != nil {
		return nil, err
	}
 	totalSamplesAfter :=  t.OperatorTelemetry.Samples().TotalSamples() // Get the samples after calling Next()
 	t.OperatorTelemetry.UpdatePeak(totalSamplesAfter - totalSamplesBefore) // The diff is the peak samples.
 	
 	return out, err
}
```

Wdyt @yeya24 @Naman-B-Parlecha ?

Naman-B-Parlecha:
> @Naman-B-Parlecha
> 
> Yes, what you're doing is fundamentally the same. This is just a cleaner way of doing it.
> 
> * We decouple the IncrementSamples() and UpdatePeaks()
> * We only do it in telemetry.Operator

Thanks for the clarification 🙌
Should I refactor the code in this pr or would u like to make a fresh pr and add changes to it?

harry671003:
Use the same PR.
Also include unit tests.

execution/telemetry/telemetry.go
```diff
@@ -203,8 +203,18 @@ func (t *Operator) Series(ctx context.Context) ([]labels.Labels, error) {
...
+
 	defer func() { t.OperatorTelemetry.AddNextExecutionTime(time.Since(start)) }()
-	return t.inner.Next(ctx)
+	out, err := t.inner.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	totalSamplesAfter := t.OperatorTelemetry.Samples().TotalSamples
+	t.OperatorTelemetry.Samples().UpdatePeak(int(totalSamplesAfter) - int(totalSamplesBefore))
```
harry671003: nit: Should this be just t.OperatorTelemetry.UpdatePeak()

harry671003: Also, we no longer need the other calls to `UpdatePeak`. Could you remove it from `IncrementSamplesAtTimestamp()`.

engine/explain_test.go
```diff
@@ -332,3 +331,58 @@ func renderAnalysisTree(node *engine.AnalyzeOutputNode, level int) string {
...
+	result := renderAnalysisTree(analyzeOutput, 0)
+	expected := `[duplicateLabelCheck]: max_series: 2 total_samples: 0 peak_samples: 0
+|---[concurrent(buff=2)]: max_series: 2 total_samples: 0 peak_samples: 0
+|   |---[aggregate] sum by ([pod]): max_series: 2 total_samples: 0 peak_samples: 0
+|   |   |---[duplicateLabelCheck]: max_series: 2 total_samples: 0 peak_samples: 0
+|   |   |   |---[coalesce]: max_series: 2 total_samples: 0 peak_samples: 0
+|   |   |   |   |---[concurrent(buff=2)]: max_series: 1 total_samples: 0 peak_samples: 0
+|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 0 mod 2): max_series: 1 total_samples: 1010 peak_samples: 200
+|   |   |   |   |---[concurrent(buff=2)]: max_series: 1 total_samples: 0 peak_samples: 0
+|   |   |   |   |   |---[matrixSelector] rate({[__name__="http_requests_total"]}[10m0s] 1 mod 2): max_series: 1 total_samples: 1010 peak_samples: 200
```
harry671003, line 385: Is the peak samples correct here?
For the selector `{[__name__="http_requests_total"]}[10m0s]`

Conditions:
- Steps batch is 10.
- Selectors shards = 2

Calculation:
- Each Next() call has 10 steps. 
- Each step has 1 series (due to sharding)
- For each step:
  - Each series has a sample every 30 seconds.
  - In [10m] range, there will be 10*60s/30s = 20 samples
  
- Total samples in each step = 10 steps * 1 series * 20 samples = 200

Looks correct.

harry671003 (commented):
Great work! Just some comments.

execution/telemetry/telemetry.go
Naman-B-Parlecha: > nit: Should this be just t.OperatorTelemetry.UpdatePeak()

Nope this should be `t.OperatorTelemetry.Samples().UpdatePeak()` itself

harry671003: I think we should add a new method in the interface to UpdatePeak()

```
type OperatorTelemetry interface {
	fmt.Stringer

	MaxSeriesCount() int
	SetMaxSeriesCount(count int)
	ExecutionTimeTaken() time.Duration
	AddSeriesExecutionTime(time.Duration)
	SeriesExecutionTime() time.Duration
	AddNextExecutionTime(time.Duration)
	NextExecutionTime() time.Duration
	IncrementSamplesAtTimestamp(samples int, t int64)
	UpdatePeak(samples int)
	Samples() *stats.QuerySamples
	LogicalNode() logicalplan.Node
}
```

Naman-B-Parlecha: Yems was thinking the same refactoring!

Naman-B-Parlecha: Added @harry671003

execution/telemetry/telemetry.go
```diff
@@ -203,8 +205,18 @@ func (t *Operator) Series(ctx context.Context) ([]labels.Labels, error) {
 
 func (t *Operator) Next(ctx context.Context) ([]model.StepVector, error) {
 	start := time.Now()
+	totalSamplesBefore := t.OperatorTelemetry.Samples().TotalSamples
```
yeya24: Seems all tests failed here due to panic.
This is because `t.OperatorTelemetry.Samples()` returns nil when `OperatorTelemetry` is NoopTelemetry if `EnableAnalysis` is set to false. We need to gracefully handle this case.

Naman-B-Parlecha: sure i will look into this

Naman-B-Parlecha:
@yeya24 PTAL

execution/telemetry/telemetry.go
```diff
@@ -203,8 +205,22 @@ func (t *Operator) Series(ctx context.Context) ([]labels.Labels, error) {
 
 func (t *Operator) Next(ctx context.Context) ([]model.StepVector, error) {
 	start := time.Now()
+	totalSamplesBefore := t.OperatorTelemetry.Samples()
+	if totalSamplesBefore == nil {
+		return t.inner.Next(ctx)
```
yeya24: I am not sure we should just call next in this case and skip other calls such as `AddNextExecutionTime`.
Since `OperatorTelemetry` is an interface, it is possible that `t.OperatorTelemetry.Samples()` is nil but `AddNextExecutionTime` is implemented properly.

Naman-B-Parlecha: We were doing it the same before this implementation so thought this would be the best go?

harry671003: How about we change the interface to add a method to get the samples?

NoOp can return 0.

Naman-B-Parlecha: Yep that works just fine

Naman-B-Parlecha: Maybe if nil 
We can just default the before and after to 0? 

so we dont have to add a new interface? @harry671003 @yeya24 what do u recommend?

harry671003: That works too

Naman-B-Parlecha: refactored, keeping it minimal as we dont want to keep adding new interfaces for every action

PR #605, opened 2025-07-09
WIP: Updating prometheus

harry671003:
WIP


.github/workflows/test.yaml
```diff
@@ -76,9 +76,9 @@ jobs:
       - name: Run unit tests
         run: make test
 
-  test-tag-stringlabels:
+  test-tag-slicelabels:
```
harry671003, line 79: Prometheus made stringlabels the default in 
https://github.com/prometheus/prometheus/issues/16064

MichaHoffmann (approved):

execution/aggregate/khashaggregate.go
```diff
@@ -108,12 +108,19 @@ func (a *kAggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
+			if math.IsNaN(val) {
+				return nil, errors.New("Parameter value is NaN")
+			}
+			if val > math.MaxInt64 {
 				return nil, errors.Newf("Scalar value %v overflows int64", val)
 			}
+			if val < math.MinInt64 {
+				return nil, errors.Newf("Scalar value %v underflows int64", val)
+			}
+			// || val < math.MinInt64 || math.IsNaN(val)
```
GiedriusS: Some random leftover?

execution/function/histogram.go
```diff
@@ -35,46 +35,68 @@ type histogramOperator struct {
...
+	case "histogram_quantile":
+		o.scalar1Op = nextOps[0]
+		o.vectorOp = nextOps[1]
+		o.scalar1Points = make([]float64, opts.StepsBatch)
+	case "histogram_fraction":
+		o.scalar1Op = nextOps[0]
+		o.scalar2Op = nextOps[1]
+		o.vectorOp = nextOps[2]
+		o.scalar1Points = make([]float64, opts.StepsBatch)
+		o.scalar2Points = make([]float64, opts.StepsBatch)
```
GiedriusS, line 83: It should be a bug if something else is passed here, right? What do you think about adding `default: panic("BUG: unsupported function passed")` to make sure we don't hit this?

yeya24 (approved):
Thanks!

PR #610, opened 2025-07-25
plan: add test for bug in distributed engine

MichaHoffmann:
The actually generated plan is:
```
group by (region) (
  dedup(
    remote(group by (region) (foo)),
    remote(group by (region) (foo)))
) * on (region) bar`,
```

which means that "bar" is fetched with Series API not as remote query.

### DIFF
--- a/logicalplan/distribute_test.go
+++ b/logicalplan/distribute_test.go
@@ -33,6 +33,16 @@ func TestDistributedExecution(t *testing.T) {
 		expected          string
 	}{
 		{
+			name: "this is a bug",
+			expr: `group by (region) (foo) * on (region) bar`,
+			expected: `
+group by (region) (
+  dedup(
+    remote(group by (region) (foo)),
+    remote(group by (region) (foo)))
+) * on (region) dedup(remote(bar), remote(bar))`,
+		},
+		{
 			name:     "selector",
 			expr:     `http_requests_total`,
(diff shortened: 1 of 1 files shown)


logicalplan/distribute_test.go
```diff
@@ -568,7 +581,7 @@ dedup(
   remote(max_over_time(sum_over_time(sum_over_time(metric[5m])[45m:10m])[15m:15m])) [1970-01-01 07:05:00 +0000 UTC, 1970-01-01 12:00:00 +0000 UTC])`,
 		},
 		{
-			name: "subquery with a total 4h range is cannot be distributed",
+			name: "subquery with a total 6h range is cannot be distributed",
```
fpetkovski, line 584: Hm maybe we want to add a new test case?

logicalplan/distribute.go
```diff
@@ -257,6 +258,22 @@ func (m DistributedExecutionOptimizer) Optimize(plan Node, opts *query.Options)
 		*current = m.distributeQuery(current, engines, m.subqueryOpts(parents, current, opts), minEngineOverlap)
 		return true
 	})
+
+	// HACK: Postprocess rewrite all "Selectors" that we missed to remote queries as base case
+	stop := false
+	Traverse(&plan, func(current *Node) {
```
fpetkovski, line 264: I don't understand why the optimizer won't catch this case.

PR #611, opened 2025-07-31
engine: dont merge in defer

MichaHoffmann:


fpetkovski:
Does this solve the race condition?

MichaHoffmann:
> Does this solve the race condition?

no idea yet, wasnt able to reproduce it; its a guess

PR #624, opened 2025-09-12
Fix MergeSelectsOptimizer dropping all matchers with the same name

afhassan:
`MergeSelectsOptimizer` optimizes selectors that share the same matchers where one selector is a subset of the other(s). The least restrictive selector is used to replace all common selectors, with filters being added to each replaced selector to return the correct series.

However there is an unexpected behavior for this optimizer when more than one matcher share the same name. For example:
```
expr:          X{a!~"b",a=~"b",c="d"}/X{a=~"b"}
expected:      filter([a!~"b" c="d"], X{a=~"b"}) / X{a=~"b"}
actual:        filter([c="d"], X{a=~"b"}) / X{a=~"b"}
```
In this case, the least restrictive selector `X{a=~"b"}` is used to replace the other selector `X{a!~"b",a=~"b",c="d"}`. A filter for the two matchers `a!~"b"` and `c="d"` should be added to return the correct series. However, when removing the matcher already present in the chosen least restrictive selector `X{a=~"b"}`, we actually drop all matchers with name `a` which includes `a!~"b"`, leaving only `c="d"`.

This behavior results in correctness issues when comparing the query ran with and without the optimizer. The fix is to strictly drop matchers with the same name, operation, and value. Instead of dropping matchers by name only.

### DIFF
--- a/logicalplan/merge_selects.go
+++ b/logicalplan/merge_selects.go
@@ -75,18 +75,8 @@ func replaceMatchers(selectors matcherHeap, expr *Node) {
 			filters := make([]*labels.Matcher, len(matchers))
 			copy(filters, matchers)
 
-			// All replacements are done on metrics name only,
-			// so we can drop the explicit metric name selector.
-			filters = dropMatcher(labels.MetricName, filters)
-
-			// Drop filters which are already present as matchers in the replacement selector.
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
Thanks! The fix looks good.

PR #637, opened 2025-10-14
ringbuffer: implement initial count_over_time ringbuffer

MichaHoffmann:


ringbuffer/overtime.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (r *CountOverTimeBuffer) MaxT() int64 { return r.lastTimestamp }
+
+func (r *CountOverTimeBuffer) Push(t int64, v Value) {
+	// Set the lastSample sample for the current evaluation step.
+	r.lastTimestamp = t
+
+	// Set the first sample for each evaluation step where the currently read sample is used.
+	for i := 0; i < len(r.stepRanges) && t > r.stepRanges[i].mint && t <= r.stepRanges[i].maxt; i++ {
```
fpetkovski, line 88: For range queries we can have overlapping windows between subsequent eval steps. Are we accounting for that here?

MichaHoffmann: yeah we just add to the corresponding stepRange if the sample overlaps

MichaHoffmann: since we loop over all step ranges, we will add the same sample to the overlapping ones too

ringbuffer/overtime.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (r *CountOverTimeBuffer) Reset(mint int64, evalt int64) {
+	if r.stepRanges[0].mint == mint {
+		return
+	}
+
+	lastSample := len(r.stepRanges) - 1
+	var (
+		nextMint = r.stepRanges[lastSample].mint + r.step
+		nextMaxt = r.stepRanges[lastSample].maxt + r.step
```
fpetkovski, line 114: should we add a `nextStep` helper function that returns the bucket for the next step?

MichaHoffmann: I think we should yeah, this is copied 1:1 from rate; do you mind if i clean up in a follwoup PR?

PR #641, opened 2025-10-15
Simplify warnings

fpetkovski:
This commit tries to simplify warnings by returning them from scalar accumulators instead of requiring a context to be passed in. It tries to return string constants wherever possible to avoid extra allocations.

Vector accumulators still need some work since they can raise multiple warnings.

### DIFF
--- a/execution/aggregate/accumulator.go
+++ b/execution/aggregate/accumulator.go
@@ -4,10 +4,7 @@
 package aggregate
 
 import (
-	"context"
 	"math"
-
-	"github.com/thanos-io/promql-engine/execution/warnings"
 
 	"github.com/efficientgo/core/errors"
 	"github.com/prometheus/prometheus/model/histogram"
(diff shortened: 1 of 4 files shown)


MichaHoffmann:
I think we need the warnings type and the (warning, error) return signature because "histogram.Add" or any other could return an unkown error.
I think further that we could just take that unkonwn error and map it to "BadHistogramOperationWarning" and simplify the signature to only return an error that we always add to annotations - this would simplify everything alot.. because really we know all errors that histogram.Add can return and they really all map to known warnings already.

execution/aggregate/accumulator.go
```diff
@@ -62,47 +64,43 @@ func (s *sumAcc) AddVector(ctx context.Context, float64s []float64, histograms [
...
-			return err
+			return nil, err
 		}
 	} else {
 		t := h.Copy()
 		if s.histSum, err = t.Add(s.histSum); err != nil {
 			if errors.Is(err, histogram.ErrHistogramsIncompatibleSchema) {
-				warnings.AddToContext(annotations.MixedExponentialCustomHistogramsWarning, ctx)
-				return nil
+				return annotations.MixedExponentialCustomHistogramsWarning, nil
```
MichaHoffmann: why dont we just retrn the error and have a "ErrorToWarning" method in some package that maps some known errors to warnings - and then we add the error that was returned to annotations in the caller. I think this would make the flow much simpler while at the same time pretty much all errors here just ocnvert to warnings and a dropped series for the timestamp anyway.
No error that t.Add could return should justifiably stop query execution.

MichaHoffmann (approved):

execution/aggregate/accumulator.go
fpetkovski: I think that going from warning to error and back to warning can be error prone and cumbersome to manage. We can just take the warning as it is and add it to the context. Computations should not raise errors anyway, those can only come from the network.

PR #643, opened 2025-10-16
Use constants for not-seen

fpetkovski:
There are some magic numbers which indicate that a series is not seen. These are hard to interpret without context, so adding some constants wont hurt.


fpetkovski:
I think we might need proper range query tests for these functions, the latest implementation does not look correct.

PR #644, opened 2025-10-16
ringbuffer: revert interface changes

MichaHoffmann:
My change was bad and I should feel bad - we actually need a ReadIntoLast method to properly handle the "extSample" between range steps. It could be that the lastSample from previous step is actually this sample and should replace the last sample in the buffer.


fpetkovski (approved):

fpetkovski:
Closes https://github.com/thanos-io/promql-engine/pull/643

PR #646, opened 2025-10-17
execution: add extwarning and extmath packages

MichaHoffmann:
The plan is that we can import aggregators in ringbuffers too; but also that we group all math-y stuff together in the future; it is currently scattered in execution/functions/functions.go and ringbuffer/functions.go


MichaHoffmann:
-

PR #664, opened 2025-12-04
all: point claude at removing vector pool

MichaHoffmann:
Instructed claude to refactor the vectoroperator engine into a reader pattern. Benchmarks show some improvements.


execution/model/vector.go
```diff
@@ -26,21 +26,48 @@ type StepVector struct {
...
+}
+
+func (s *StepVector) AppendSample(id uint64, val float64) {
+	s.SampleIDs = append(s.SampleIDs, id)
+	s.Samples = append(s.Samples, val)
+}
+
+// AppendSampleWithHint appends a sample and lazily pre-allocates capacity if needed.
+// Use this when you know the expected number of samples to avoid repeated slice growth.
+func (s *StepVector) AppendSampleWithHint(id uint64, val float64, hint int) {
```
fpetkovski: Should we call this `AppendSampleWithSizeHint`

execution/noop/operator.go
```diff
@@ -19,10 +19,13 @@ type operator struct {
 
 func NewOperator(opts *query.Options) model.VectorOperator {
 	scanner := prometheus.NewVectorSelector(
-		model.NewVectorPool(0),
 		noopSelector{},
 		opts,
-		0, 0, false, 0, 1,
+		0,     // offset
```
fpetkovski, line 24: Does it make sense to create alias types?

engine/projection_test.go
```diff
@@ -214,7 +214,7 @@ func TestProjectionWithFuzz(t *testing.T) {
 	queryTime := time.Unix(600, 0)
 
 	t.Logf("Running %d fuzzy tests with seed %d", testRuns, seed)
-	for i := 0; i < testRuns; i++ {
+	for i := range testRuns {
```
fpetkovski, line 217: FFR: We should add modernize to the linter.

execution/aggregate/hashaggregate.go
```diff
@@ -140,14 +132,16 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
 		}
 		a.lastBatch = nil
 	}
+
 	for {
-		next, err := a.next.Next(ctx)
+		n, err := a.next.Next(ctx, a.tempBuf)
```
fpetkovski, line 141: Do we want to initialize `tempBuf` to the batch size?

execution/aggregate/hashaggregate.go
```diff
@@ -156,31 +150,36 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
 	if a.tables[0].timestamp() == math.MinInt64 {
-		return nil, nil
+		return 0, nil
 	}
 
-	result := a.vectorPool.GetVectorBatch()
+	n := 0
 	for i := range a.tables {
-		if a.tables[i].timestamp() == math.MinInt64 {
+		if n >= len(buf) {
```
fpetkovski: can we just do `for i in range n` in the loop? I guess this is also resilient in cases where a step has no samples?

execution/aggregate/hashaggregate.go
```diff
@@ -156,31 +150,36 @@ func (a *aggregate) Next(ctx context.Context) ([]model.StepVector, error) {
...
 			break
 		}
-		result = append(result, a.tables[i].toVector(ctx, a.vectorPool))
+		ts := a.tables[i].timestamp()
+		if ts == math.MinInt64 {
```
fpetkovski: This was removed above, do we still need it here?

execution/aggregate/hashaggregate.go
fpetkovski: Ah ok, I just saw it's initialized here: https://github.com/thanos-io/promql-engine/pull/664/changes#diff-8cffb8d32523e51056687466b98e3a24a7b27643c72bbaf0a0757ec210947e10R206

execution/exchange/coalesce.go
```diff
@@ -84,93 +83,134 @@ func (c *coalesce) Series(ctx context.Context) ([]labels.Labels, error) {
...
 	var minTs int64 = math.MaxInt64
 	var errChan = make(errorChan, len(c.operators))
+	vectorCounts := make([]int, len(c.operators))
+
 	for idx, o := range c.operators {
 		// We already have a batch from the previous iteration.
 		if c.inVectors[idx] != nil {
+			mu.Lock()
+			if len(c.inVectors[idx]) > 0 && minTs > c.inVectors[idx][0].T {
+				minTs = c.inVectors[idx][0].T
```
fpetkovski: Maybe we can use `max` here, it's easier to understand what's going on.

execution/exchange/coalesce.go
```diff
@@ -84,93 +83,134 @@ func (c *coalesce) Series(ctx context.Context) ([]labels.Labels, error) {
...
-				return
-			}
 
-			mu.Lock()
-			if minTs > in[0].T {
-				minTs = in[0].T
+			if n > 0 {
+				c.inVectors[opIdx] = c.tempBufs[opIdx][:n]
+				mu.Lock()
+				if minTs > c.tempBufs[opIdx][0].T {
```
fpetkovski: Same as above, we can use `max`

execution/exchange/coalesce.go
MichaHoffmann: we can use min! if minTs happens to be bigger we replace it with the smaller thing.

PR #667, opened 2025-12-06
engine: fix preallocated result slice

MichaHoffmann:
We do not properly calculate the result slices. For instant queries they of course only need to hold space for at most one sample and for longer range queries they need to hold more space. We just need to calculate that from the range query parameters properly.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -273,7 +273,7 @@ func (e *Engine) MakeInstantQuery(ctx context.Context, q storage.Queryable, opts
 	}
 	e.metrics.totalQueries.Inc()
 	return &compatibilityQuery{
-		Query:      &Query{exec: exec, opts: opts},
+		Query:      &Query{exec: exec, opts: qOpts},
 		engine:     e,
 		plan:       optimizedPlan,
 		warns:      warns,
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -531,6 +531,13 @@ type compatibilityQuery struct {
 	scanners engstorage.Scanners
 }
 
+func (q *compatibilityQuery) totalSteps() int {
+	if q.step == 0 {
+		return 1
+	}
+	return int((q.end.UnixMilli()-q.start.UnixMilli())/q.step.Milliseconds() + 1)
```
fpetkovski: Don't we have this calculation in Options?

MichaHoffmann: Yeah it should probably live there but we don't have it yet.

fpetkovski: We have this: https://github.com/thanos-io/promql-engine/blob/main/query/options.go#L23-L38

MichaHoffmann: Yeah but that caps it to the batch size so it's not the number of results steps I think

fpetkovski: But I mean we can at least consolidate :)

MichaHoffmann: I'll add a method in the query opts for this tomorrow!

PR #680, opened 2026-01-06
Change `RemoteEndpoints` interface to support remote engines pruning

SuperPaintman:
This PR changes the `RemoteEndpoints` interface. Now it accepts 'hints' from the caller when selecting remote engines. These hints will be used to prune TSDBInfos (as described in this https://github.com/thanos-io/thanos/pull/8599).

## Notes

~~**Note 1**:~~

~~I added two interfaces: `RemoteEndpointsV2` (as suggested here https://github.com/thanos-io/thanos/pull/8599#issuecomment-3653228727), and `RemoteEndpointsV3` which can be extended without making any new breaking changes.~~

~~Let me know which one you prefer, and I'll remove the other one.~~

## Verification

- [x] Unit tests.
    - Let's discuss the interface / implementation first. Once that's settled, I'll add unit tests. 
- [x] Benchmark
    - ~~I'm going to add a benchmark for dist-engine optimizers with synthetic data (as described in the main issue: https://github.com/thanos-io/thanos/pull/8597)~~.
    - I think they should be added in the `thanos` repo.

--

Issue: https://github.com/thanos-io/thanos/pull/8597

CC: @MichaHoffmann


api/remote.go
```diff
@@ -16,10 +18,46 @@ type RemoteQuery interface {
...
+// also may safely ignore them and return all available remote engines.
+//
+// NOTE(Aleksandr Krivoshchekov):
+// We add a new interface as a temporary backward compatibility.
+// RemoteEndpoints will be replaced with it in a future breaking change.
+//
+// Unlike RemoteEndpointsV2, this interface can be extended with more hints
+// in the future, without making any breaking changes.
+type RemoteEndpointsV3 interface {
+	EnginesV3(query RemoteEndpointsQuery) []RemoteEngine
```
fpetkovski: This makes it easy to extend endpoints, but third implementations could break since might not see the new parameter when it's added. With a positional argument they'll see it through a compiler error. Wdyt @MichaHoffmann?

api/remote.go
```diff
@@ -44,6 +82,71 @@ func (m staticEndpoints) Engines() []RemoteEngine {
 	return m.engines
 }
 
+func (m staticEndpoints) EnginesV2(mint, maxt int64) []RemoteEngine {
+	return m.engines
```
fpetkovski, line 57: Should we apply filtering here?

SuperPaintman: I tried, but since pruning is optional, I decided to not change `DistributedExecutionOptimizer` implementation, because it assumes that some engines [might not have the requested range and fall back](https://github.com/thanos-io/promql-engine/blob/3988aa4704b558553871f66e6c20e996d8b4020a/logicalplan/distribute.go#L316-L329).

Some [some unit tests](https://github.com/thanos-io/promql-engine/blob/3988aa4704b558553871f66e6c20e996d8b4020a/logicalplan/distribute_test.go#L848) check for that, and if we apply filter in static endpoints, they fail.

But now when I'm thinking about that again, I'm not sure that's an issue anymore.

api/remote.go
MichaHoffmann: im fine with enginesv2(mint, maxt int64) and explict parameters i think!

SuperPaintman: I think that all `RemoteEndpointsQuery` fields should be optional in the sense that the implementation can safely ignore some/all of them.

Worst case: optimizers will have to go through more engines/labels than without pruning.

api/remote.go
MichaHoffmann: I think yes, the API contract suggest that we only should return intersecting engines.

SuperPaintman: Are you suggesting something like this?
```go
func (m staticEndpoints) Engines(mint, maxt int64) []RemoteEngine {
	var engines []RemoteEngine
	for _, e := range m.engines {
		if e.MaxT() < mint || e.MinT() > maxt {
			continue
		}
		engines = append(engines, e)
	}
	return engines
}
```

SuperPaintman: If so, I'm a bit hesitant about filtering engines themselves here (Maybe I don't fully understand the original implementation).

The optimizers seem to expect all engines to be available, especially for handling `absent()` (in `DistributedExecutionOptimizer.distributeAbsent`). If the engine is removed from query (e.g. if it has a data gap for the requested range), dist engine will return incorrect results. 

Also, filtering engines breaks `TestDistributedExecutionWithLongSelectorRanges/skip_distributing_queries_with_timestamps_outside_of_the_range_of_an_engine`:

```diff
--- Expected
+++ Actual
@@ -1 +1 @@
-sum(sum by (region) (metric @ 18000.000))
+sum(dedup(remote(sum by (region) (metric @ 18000.000))))
```

Let me know if I'm mistaken, but it seems like an invalid result.

The [current "real" implementation](https://github.com/thanos-io/thanos/blob/094aa1d5b7f9291166484601fa90cc1c863983c5/pkg/query/remote_engine.go#L149-L188) also assumes all engines are return.

The intent of the `mint`/`maxt` is to prune internal metadata (like `TSDBInfos`) within each engine, to reduce unnecessary  computations later.

Does that make sense, or am I missing something?

fpetkovski: Gotcha, this indeed makes sense, especially since you brought up `absent`. I think we can leave this as it is and optimize later.

fpetkovski (approved):
This seems good to me, still has to be tested through Thanos but I am okay with merging the change.

SuperPaintman:
@fpetkovski could you merge this PR? Or is there anything you would like me to change?

fpetkovski:
@SuperPaintman I was planning to do a last quick local review today and I will merge it. I resolved conflicts against main.

fpetkovski:
@SuperPaintman will merge this now, there are some minor cleanup I'd like to do but that can be done on main. Thanks for the contribution.

PR #682, opened 2026-01-12
execution: batch by series optimization for matrix selectors with high window overlap

afhassan:
This PR explores using a series-batching optimization to reduce peak `RingBuffer` memory usage for queries with long lookback windows and high cardinality.

Problem: Queries like `sum(increase(metric[24h]))` with a small step size create relatively large ring buffer for matrix selectors. This is done once per series and can quickly bloat memory causing OOM for high cardinality queries. 

Solution: When matrix selector window overlap exceeds a certain number of steps (example 100), the optimizer switches from step batching to series batching:

- Processes **all steps** for a small batch of series (default: 1000)
- Reuse a fixed pool of ring buffers for all series batches
- This reduces peak memory significantly and avoids OOM

Changes:
- Added EnableHighOverlapBatching option to engine
- Extended SelectorBatchSize optimizer to detect high overlap queries 
  - If detected, adjust StepsBatch to cover all steps and set a series batch size instead
- Implemented BufferPool for ring buffer reuse across series batches
- Added Clear() method to all buffer types for proper state reset before reuse

_This PR is work in progress... The main blocker is validating that series batching can work for certain functions/operators that need all series present for execution_

### DIFF
--- a/logicalplan/set_batch_size.go
+++ b/logicalplan/set_batch_size.go
@@ -13,14 +13,39 @@ import (
 // SelectorBatchSize configures the batch size of selector based on
 // aggregates present in the plan.
 type SelectorBatchSize struct {
-	Size int64
+	// DefaultBatchSize is the series batch size for standard batching.
+	// Applied to vector selectors under aggregations.
+	DefaultBatchSize int64
+
+	// EnableHighOverlapBatching reduces memory for queries with long lookback windows.
+	EnableHighOverlapBatching bool
+
+	// HighOverlapBatchSize is the series batch size for high-overlap queries. Defaults to 1000.
+	HighOverlapBatchSize int64
+
+	// HighOverlapThreshold is the overlap threshold that triggers the optimization. Defaults to 100.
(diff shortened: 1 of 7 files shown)


logicalplan/set_batch_size.go
```diff
@@ -13,14 +13,39 @@ import (
...
+
+	// HighOverlapThreshold is the overlap threshold that triggers the optimization. Defaults to 100.
+	HighOverlapThreshold int64
 }
 
 // Optimize configures the batch size of selector based on the query plan.
 // If any aggregate is present in the plan, the batch size is set to the configured value.
 // The two exceptions where this cannot be done is if the aggregate is quantile, or
 // when a binary expression precedes the aggregate.
-func (m SelectorBatchSize) Optimize(plan Node, _ *query.Options) (Node, annotations.Annotations) {
```
MichaHoffmann, line 23: im not sure this will even reach into ringbuffers, right now this only applies to vectors in direct aggregations i think?

PR #697, opened 2026-03-20
fix(fuzz): address fuzzer failures for step-invariant and count_values

sylr:
## Summary

- Skip sample validation for expressions using the `@` modifier, since the Thanos step invariant operator caches results via `sync.Once` and replays without re-counting samples per step (telemetry difference, not correctness)
- Exclude `count_values` from fuzzer validation, since it stringifies float values into labels, amplifying last-digit float64 precision differences into label mismatches
- Add fuzzer corpus entry for `FuzzNativeHistogramQuery`

## Test plan

- [x] `FuzzNativeHistogramQuery/df60c8694f1c9759` passes
- [x] `FuzzEnginePromQLSmithRangeQuery/59c2955b78c86bab` passes
- [x] All other fuzzer corpus entries pass
- [x] Full `go test ./...` passes (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/engine/enginefuzz_test.go
+++ b/engine/enginefuzz_test.go
@@ -90,6 +90,14 @@ func validateExpr(expr parser.Expr, testType testType) bool {
 
 	parser.Inspect(expr, func(node parser.Node, path []parser.Node) error {
 		switch n := node.(type) {
+		case *parser.AggregateExpr:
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
Thanks!

MichaHoffmann (approved):

PR #705, opened 2026-04-07
execution: push limits into telemetry

MichaHoffmann:
OperatorTelemetry already has hooks to track samples per timestamp, we can use the same hooks that are already called to enforce the configured limits probably too?


engine/engine.go
```diff
@@ -314,7 +314,6 @@ func (e *Engine) MakeInstantQueryFromPlan(ctx context.Context, q storage.Queryab
 	if err != nil {
 		return nil, err
 	}
-	e.metrics.totalQueries.Inc()
```
MichaHoffmann, line 317: what

fpetkovski (approved):

yeya24:
@PaurushGarg Please help take a look at the PR as you added max samples limit. Thanks

PR #708, opened 2026-04-23
engine: add deterministic operator IDs, pass through context

GiedriusS:
Create deterministic operator IDs that only depend on the position of the operator in the tree. This will allow merging multiple trees if, for example, query-frontend shards the same query into multiple ones.

This also allows attaching arbitrary metadata on the Thanos side without the PromQL engine knowing anything about it.

I haven't checked 100% all cases whether all IDs are REALLY deterministic but that will come up during implementation on the other side. I played around a bit and this fully enables enriching Explain functionality with fan-out data.

V2 of https://github.com/thanos-io/promql-engine/pull/490.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -500,7 +500,7 @@ func (q *Query) Explain() *ExplainOutputNode {
 }
 
 func (q *Query) Analyze() *AnalyzeOutputNode {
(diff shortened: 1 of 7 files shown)


fpetkovski (approved):

fpetkovski:
Looks like there could be some loop that causes tests to time out.

PR #711, opened 2026-05-28
execution/binary: make duplicate series error message deterministic

### LINKED ISSUE #710: execution/binary: "found duplicate series for the match group" error is non-deterministic
## Summary

\`errManyToManyMatch.Error()\` in \`execution/binary/utils.go\` formats the two duplicate series labels in whichever order they arrive on the \`original\`/\`duplicate\` slots. That order is non-deterministic upstream — driven by \`StepVector\` iteration which is itself fed by Go map iteration (see e.g. \`execution/scan/subquery.go\`) — so the rendered error string can differ across runs even when the underlying error is identical.

## Root cause

\`execution/binary/utils.go:43-47\`:

\`\`\`go
func (e *errManyToManyMatch) Error() string {
    group := e.original.MatchLabels(e.matching.On, e.matching.MatchingLabels...)
    msg := "found duplicate series for the match group %s on the %s hand-side of the operation: [%s, %s]" +
        ";many-to-many matching not allowed: matching labels must be unique on one side"
    return fmt.Sprintf(msg, group, e.side, e.original.String(), e.duplicate.String())
}
\`\`\`

\`e.original\` and \`e.duplicate\` are filled in by the caller in whichever order the duplicate pair is encountered. Across two engine instances evaluating the same query (e.g. Cortex's fuzz test comparing two binaries), the order can differ.

## Downstream impact

Same shape of bug also exists in upstream Prometheus's \`promql/engine.go\` (prometheus/prometheus#18809). The combined non-determinism causes a flaky test in Cortex: cortexproject/cortex#7546. A tactical downstream workaround exists in cortexproject/cortex#7550, but the cleaner fix is to make the upstream error deterministic on both engines.

## Proposed fix

Sort the two label strings before formatting in \`Error()\`. Minimum-surface fix, no new imports needed (string \`<\` comparison is enough for two values). PR follows.

sandy2008:
The `errManyToManyMatch.Error()` method previously formatted the two conflicting series in whichever order they happened to arrive on the `original` / `duplicate` slots. That order is driven by upstream `StepVector` iteration which is itself fed by Go map iteration (see `execution/scan/subquery.go`), so the rendered error string is non-deterministic across runs and across parallel engine workers.

Same root cause as the upstream Prometheus issue/PR (prometheus/prometheus#18809 / prometheus/prometheus#18810). Tracked downstream as a flaky test in Cortex: cortexproject/cortex#7546 (tactical workaround in cortexproject/cortex#7550 — removable once both upstream fixes land and are vendored).

Fix: lexicographically sort the two label-set strings before formatting the message. One-line if-swap, no new imports, preserves the existing message format so existing matchers keep working.

A regression test in the new `execution/binary/utils_test.go` exercises both right-hand and left-hand side cases as well as adversarial label values containing the same framing punctuation (`[`, `]`, `,`, `{`, `}`, `"`), and asserts byte-identical output regardless of which series came first. Verified by temporarily reverting the swap: all subtests FAIL with the unsorted ordering; restoring the patch makes them PASS.

Fixes #710.

### DIFF
--- a/execution/binary/utils.go
+++ b/execution/binary/utils.go
@@ -42,9 +42,18 @@ func newManyToManyMatchError(matching *parser.VectorMatching, original, duplicat
 
 func (e *errManyToManyMatch) Error() string {
 	group := e.original.MatchLabels(e.matching.On, e.matching.MatchingLabels...)
+	// The choice of which series is reported as "original" vs "duplicate" is
+	// driven by upstream StepVector / map iteration order and is therefore
+	// non-deterministic across runs. Sort the rendered label-set strings so
+	// the error message is stable; downstream consumers (e.g. Cortex,
+	// see cortexproject/cortex#7546) compare these messages.
+	original, duplicate := e.original.String(), e.duplicate.String()
+	if duplicate < original {
+		original, duplicate = duplicate, original
+	}
 	msg := "found duplicate series for the match group %s on the %s hand-side of the operation: [%s, %s]" +
 		";many-to-many matching not allowed: matching labels must be unique on one side"
-	return fmt.Sprintf(msg, group, e.side, e.original.String(), e.duplicate.String())
+	return fmt.Sprintf(msg, group, e.side, original, duplicate)
 }
 
 func shouldDropMetricName(op parser.ItemType, returnBool bool) bool {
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):
Thank you

PR #713, opened 2026-06-27
execution/binary: dedup arithmetic join outputMap by include combo

yeya24:
## Problem

For vector arithmetic joins (`group_left`/`group_right` and one-to-one), `initJoinTables` stores one `outputMap` entry per `(highCard, lowCard)` matching pair. When many low-card series share a join key over the query window — for example high-cardinality `kube_*` metrics where pods are recreated under new `uid`s on the same node — the map grows to the full pair fan-out (hundreds of millions of entries) and dominates querier heap, which can OOM the querier, even though most of those entries resolve to the **same** output series.

## Change

The output series identity of an arithmetic join depends on the low-card side only through the values of the `matching.Include` labels (see `resultMetric`, which reads the low-card side solely via `lowCard.Get(ln)` for `ln` in `matching.Include`).

This PR de-duplicates low-card series by a dense "combo" id of their `Include`-label values and keys `outputMap` by `(highCard, combo)`. Runtime lookups in `execBinaryArithmetic` translate the matched low-card id through this combo map. The fan-out collapses to one entry per `(highCard, distinct Include combo)`.

This is **lossless**: identical results and series ordering. The `and`/`or`/`unless` set-operation paths are unchanged.

## Testing

`engine/outputmap_dedup_test.go` compares this engine against the Prometheus reference engine for:

- temporal-churn collapse — many low-card series sharing one `Include` value, non-overlapping in time (terminated with `stale` markers so each step is a legal many-to-one);
- distinct-`Include` preservation — different `Include` values must not be over-collapsed;
- empty `group_left()`;
- one-to-one `*`.

`BenchmarkBinaryOutputMapDedup` exercises the churn shape (`b.ReportAllocs`).

Locally: scoped `go test ./engine/... ./execution/... ./logicalplan/...` (includes the existing Prometheus reference-engine comparison and fuzz correctness tests), `-tags slicelabels` build, `gofmt`, and `go vet` all pass.

### DIFF
--- a/engine/binary_join_test.go
+++ b/engine/binary_join_test.go
@@ -1,0 +1,339 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package engine_test
+
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/thanos-io/promql-engine/engine"
+	"github.com/thanos-io/promql-engine/execution/binary"
+	"github.com/thanos-io/promql-engine/execution/model"
+	"github.com/thanos-io/promql-engine/query"
+
+	"github.com/efficientgo/core/testutil"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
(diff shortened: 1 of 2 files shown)


yeya24:
Closed due to #714

Issue #483, opened 2024-08-15
Distributed Engine: push down even more expressions


MichaHoffmann:
I have been thinking what makes an expression computable in a remote engine and to be reassembled in the root engine. I think the following might be a sound algorithm that enables us to push down even more expressions. 

# Problem

Assume we have remote engines that are partitioned by a label `P`. Imagine an expression like `topk(10, sum by (P, instance) (X))` and assume that `instance` is a high-cardinality label. In the current implementation we would have to evaluate `sum by (P, instance) (X)` remotely and fetch all values of instance into the root engine to compute the surrounding `topk` even though we would not need most of them, clearly we can evaluate `topk` per engine and only coalesce the top 10 values for the inner expression since we dont have any overlap. But the algorithm right now doesnt know that.

# Observation

Imagine we are traversing the AST of a PromQL expression from the bottom up. Assume we have expression nodes P and C, where P is the parent of node C. If the expression at node C still has its original `P` label I want to conjecture that we can evaluate C in the remote engines and how we coalesce in the root engine depends on the parent node P. If P is nil and we are at the root we just join the series together, if its an aggregation we do the aggregation in the root engine, etc.

What is meant with the heuristic that expression node C is having the original `P` label still? I think we can define this recursive. In the base case we are at a VectorSelector, by definition it will have `P` (disregarding constant expressions and absent functions right now). If we are a node that has children and all children have the `P` label still then we also have it if we are not a `relabel_replace` function that overwrites it, if we are not aggregating it away (i.e. `sum without(P) (X)` would lose P) or if we are not a binary expression that doesnt have `P` in its matching labels (i.e. `A * ignoring (P) B` would lose P). In all other cases unless im forgetting something we should still have the partition label P and should be computable in a remote engine.

# Suggestion

We rewrite the distribution algorithm by checking recursively if we keep the partition labels `P`, if so we traverse up until we cannot anymore. Once a parent P would lose the label we stop recursion and evaluate the expression henceforth in the root engine, we coalesce depending on P

Is there anything obiously wrong with this approach?

PR #716, opened 2026-07-07
docs: add engine architecture paper draft

fpetkovski:
Add docs/architecture.md, a paper-structured overview of the engine: the columnar batched volcano model, up-front Series() contract and integer series identity, caller-owned buffers as the defence against GC pressure, streaming ring buffers for range functions (and their per-series cardinality limitation), the two parallelism axes, the optimizer framework including projection pushdown, and distributed execution as an algebra-aware plan rewrite.

Also refresh the README design section to match the current implementation: replace the removed vector-pool description with the caller-owns-buffer contract, document the MaxSamples limit, and describe the actual concurrency model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,21 +58,23 @@ One challenge with the streamed execution model is knowing how much memory to al
 To work around this issue, operators expose a `Series()` method which returns the labels for all time series that they will ever produce (for all `Next()` calls). Operators at the very bottom of the tree, like vector and matrix selectors, have this information since they are responsible for loading data from storage. Other operators can then call `Series()` on the downstream operator and pre-compute all possible outputs.
 
 Even though this might look like an expensive operation, its cost is identical to just one evaluation step. Knowing sizes of input and output vectors also allows us to:
-* allocate memory very precisely by properly sizing vector pools (see section below),
+* allocate output vectors very precisely by sizing them to the exact number of series (see section below),
 * use arrays instead of maps for indexing data, leading to faster execution times due to having less allocations and using index-based lookups, and
 * use tight loops in operators by eliminating conditional statements associated with maps.
 
-#### Vector pools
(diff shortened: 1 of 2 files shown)


MichaHoffmann:
I feel in the "Distributed Execution" section we could dive a bit more into how we compute what part of the expression can be distributed. The article just calls it the "highest" node but doesnt describe how we find it.
It used to be the lowest aggregations in the tree, but we generalized it since to track the partition labels through the expression tree and distribute the first node that loses them like here: https://github.com/thanos-io/promql-engine/issues/483

README.md
pedro-stanaka, line 116: Maybe lets "link" the new arch doc here as well.

docs/architecture.md
```diff
@@ -0,0 +1,401 @@
...
+producer operator from its consumer through a bounded channel. A background
+goroutine calls the child's `Next` ahead of demand and enqueues filled buffers;
+the consumer swaps buffer ownership without copying and returns emptied buffers
+for reuse. Inserting a concurrent exchange above expensive stages (aggregations,
+deduplication, remote calls, each scan shard) lets each stage of the pipeline
+run on its own core, so an operator can begin work on step *n* while its child
+computes step *n+1*.
+
+**Intra-operator (data) parallelism.** Selectors are fanned out into several
+shards, each reading a disjoint hash partition of the matching series, each
```
pedro-stanaka:
```suggestion
shards, each reading/decoding a disjoint index range of the selector’s matched series list,
```

pedro-stanaka (approved):
lg - the document could reuse some images we already have to make it more parseable. We have many outdated references in readme vs real implementation as well, but maybe it is something for another PR.

MichaHoffmann (approved):

pedro-stanaka (approved):

PR #718, opened 2026-07-10
Support extended range functions on native histograms

erin-noe-payne:
## Summary

`xrate` / `xincrease` / `xdelta` returned incorrect results on Prometheus **native histograms**. This PR makes the extended range functions correct for native histograms end-to-end.

## Root cause

The bug had two layers:

1. **`ringbuffer/functions.go` (`extendedRate`)** — for histogram samples the function called `histogramRate` and returned immediately (`// TODO - support extended rate for histograms`), skipping the per-second division and all boundary/windowing logic. As a result `xrate` and `xincrease` returned **identical** histograms: `xrate` never divided by the range, so it was effectively an *increase*, off by a factor of the range in seconds (e.g. 300× for a 5m window).

2. **`storage/prometheus/matrix_selector.go`** — even so, that branch was unreachable: the matrix selector rejected native histograms for extended functions with `ErrNativeHistogramsNotSupported` before `extendedRate` was ever called, only tracked `metricAppearedTs` for float samples, and only applied the extended-lookback buffering (keep the last sample before the range start) to floats.

### `ringbuffer/functions.go`
- Implement the histogram branch of `extendedRate` with full parity to the float path:
  - `xincrease` "zero" injection for a newly-appeared, single-valued series (`sameHistogramValues` helper);
  - drop the pre-range point when it is too far from `rangeStart` (`xrate`/`xdelta`);
  - adjust-to-range scaling when the samples cover the whole range (`xrate`/`xdelta`);
  - the per-second division for `xrate`.
- Change `extendedRate` to return `(float64, *histogram.FloatHistogram, bool, warnings.Warnings, error)`, mirroring `extrapolatedRate`. This lets histogram cases **emit no sample** (`ok=false`) instead of a bogus float `0` (single sample, or a range mixing floats and histograms) and propagate the same warning bits (`WarnNotCounter`, `WarnNotGauge`, `WarnMixedFloatsHistograms`). The `xrate`/`xincrease`/`xdelta` wrappers now return `extendedRate` directly.
- Add an explicit top-level mixed float/histogram guard to `extendedRate` (parity with `extrapolatedRate`): any range containing both sample types returns no sample with `WarnMixedFloatsHistograms`, **regardless of ordering**. Previously only histogram-first ranges were caught (inside `histogramRate`); a **float-first** range fell into the float path, read `.V.F` (`== 0`) from the histogram samples, and emitted a silent bogus zero — inconsistent with `rate`/`increase`/`delta` on identical input. This is reachable now that the `ErrNativeHistogramsNotSupported` gate is gone (a real float→native-histogram migration).
- Factor the repeated `isCounter && !isRate` guard into a single `isXIncrease` local, used on both the histogram and float paths for readability.

### `storage/prometheus/matrix_selector.go`
- Remove the `ErrNativeHistogramsNotSupported` gate so native histograms flow into the extended functions.
- Set `metricAppearedTs` for histogram samples (previously only floats), preventing the "metric still hasn't appeared" panic on pure-histogram series.
- Feed histograms through the extended-lookback buffer, retaining the last sample **strictly before** the range start as the baseline (parity with the float path via `ReadIntoLast`). Without this, instant/first-step queries under-counted the leading edge.

## Semantics notes
- The extended (non-extrapolating) semantics are preserved: the delta already spans the window (baseline = the pre-window sample), and only the per-second division is applied for `xrate`.
- The histogram scaling deliberately uses `float64(selectRange) / 1000` rather than the float path's integer `float64(selectRange / 1000)`, avoiding a sub-second truncation inaccuracy. This is a minor, intentional divergence from the float path.

## For maintainer review
- `ErrNativeHistogramsNotSupported` is now unused but left **declared** to avoid a breaking API change for downstream consumers. Happy to remove it (and its `errors` import) if you'd prefer.

## Testing

Prometheus core has no `x*` functions, so the cross-engine fuzz suite cannot validate them and the tests are bespoke.

**`ringbuffer/functions_test.go` (new, white-box unit tests):**
- per-second relationship `xrate == xincrease / rangeSeconds` **and** `xrate != xincrease`, with **bucket-level** assertions (`PositiveBuckets`/`Count`/`ZeroCount`), not just `Sum`/`Count`;
- single-sample `xincrease` zero-injection (and `xrate` emitting nothing);
- mixed float/histogram range (**both orderings**: float-first and histogram-first, across `xincrease`/`xrate`/`xdelta`) → no sample + `WarnMixedFloatsHistograms`;
- table cases: `xdelta` raw delta, counter-reset correction (`xincrease`) vs. no correction (`xdelta`) with bucket-level checks, drop-too-far pre-range baseline, all-samples-before-range, flat-series zero-injection inside and past the injection window, duplicate timestamps, and a non-zero offset;
- warning bits: not-counter (`xincrease` on gauge), not-gauge (`xdelta` on counter), mixed floats/histograms;
- a schema change across the range exercising `histogramRate`'s min-schema / `CopyToSchema` path.

**`engine/engine_test.go`:**
- `TestXFunctionsWithNativeHistograms` now asserts `xincrease`/`xrate`/`xdelta` return histograms (no longer an error) and checks the per-second relationship via `histogram_sum(...)`;
- `TestXFunctionsNativeHistogramsRangeQuery` — a multi-step range query exercising buffer ext-lookback retention, `lastSample` carry-over, and the histogram `ReadIntoLast` branch;
- `TestXIncreaseNativeHistogramPreRangeBaseline` — an instant query whose range start falls between two samples, pinning the strictly-before-range baseline (fails if the pre-range point is dropped);
- `TestXFunctionsFloatToHistogramTransition` — a range spanning a float→native-histogram migration, asserting the extended functions agree with `rate`/`increase`/`delta` (no bogus float sample) for a mixed range.

`go test ./...`, `go vet ./...`, and `gofmt` all pass.

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2655,7 +2655,7 @@ func TestXFunctionsRangeQuery(t *testing.T) {
 					Metric: labels.New(),
 					Floats: []promql.FPoint{
 						{T: 00_000, F: 1},
-						{T: 20_000, F: 9}, // TODO: this seems odd, feels like it should be 5
+						{T: 20_000, F: 5},
 						{T: 40_000, F: 0},
 						{T: 60_000, F: 0},
 					},
@@ -2803,9 +2803,13 @@ func TestXFunctionsWithNativeHistograms(t *testing.T) {
 	require.Greater(t, xincSum/xrateSum, 100.0, "xrate must be divided by the range, not equal to xincrease")
 }
 
-// TestXFunctionsNativeHistogramsRangeQuery exercises the multi-step plumbing
-// (buffer ext-lookback retention, lastSample carry-over, and the histogram
-// ReadIntoLast branch in the matrix selector) that the instant-only test skips.
+// TestXFunctionsNativeHistogramsRangeQuery exercises multi-step range queries
+// with concrete numeric assertions. It uses non-overlapping windows (the
+// intended way to call these functions, since they look back to the sample
+// before the range start): back-to-back (step == range) which tiles the range
+// exactly, and gapped (step > range) with a sample landing on the next range
+// boundary, which is the regression guard for the prefetched-baseline bug in
+// the matrix selector.
 func TestXFunctionsNativeHistogramsRangeQuery(t *testing.T) {
 	opts := promql.EngineOpts{
 		Timeout:              1 * time.Hour,
@@ -2814,11 +2818,27 @@ func TestXFunctionsNativeHistogramsRangeQuery(t *testing.T) {
 		EnableAtModifier:     true,
 	}
 
+	mkHist := func(mult float64) *histogram.FloatHistogram {
+		return &histogram.FloatHistogram{
+			Schema:          0,
+			ZeroThreshold:   0.001,
+			ZeroCount:       mult,
+			Count:           10 * mult,
+			Sum:             20 * mult,
+			PositiveSpans:   []histogram.Span{{Offset: 0, Length: 2}},
+			PositiveBuckets: []float64{4 * mult, 5 * mult},
+		}
+	}
+
 	lStorage := teststorage.New(t)
 	defer lStorage.Close()
 
 	app := lStorage.Appender(context.TODO())
-	testutil.Ok(t, generateFloatHistogramSeries(app, 100, false))
+	// Sums: t=0->20, 30->40, 60->60, 90->80, 120->100, 150->120.
+	for i, sec := range []int64{0, 30, 60, 90, 120, 150} {
+		_, err := app.AppendHistogram(0, labels.FromStrings(labels.MetricName, "h"), time.Unix(sec, 0).UnixMilli(), nil, mkHist(float64(i+1)))
+		testutil.Ok(t, err)
+	}
 	testutil.Ok(t, app.Commit())
 
 	newEngine := engine.New(engine.Opts{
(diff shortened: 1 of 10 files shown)


erin-noe-payne:
Thanks for the detailed review. Pushed `0580e97` addressing all three findings, plus two related boundary bugs that surfaced while fixing #1. Summary below.

## Finding #1 (High) — boundary sample lost in range queries

Root cause: the extended functions have three sources that decide whether a sample is the pre-range baseline or in-window — `ringbuffer.Reset` retention, the prefetched `lastSample`, and the iterator loop — and they disagreed at the range boundary. The `lastSample` path used a strict `t > mint` check where `mint` had already been bumped to the iterator resume cursor (`buffer.MaxT()+1`), so a sample landing exactly on the range start that had been prefetched on the previous step was neither pushed nor re-read by the iterator, and the stale point retained by `Reset` was used instead.

Fix: classify all three sources against an immutable `rangeStart` (the semantic range boundary) with the same ext-lookback bound, keeping the mutated `mint` purely as the iterator resume cursor. This is the "the paths must agree" point, generalized to all three.

This also corrected a latent float bug: the `gaps between steps` case now returns `5` instead of `9`, matching the `// TODO: this seems odd, feels like it should be 5` that was already in that test.

## Finding #2 (Medium) — xincrease zero-injection could suppress `WarnNotCounter`

The zero-injection path inspected only `samples[0].CounterResetHint`. Since `sameHistogramValues` compares via `FloatHistogram.Equals`, which ignores `CounterResetHint`, an equal-valued later sample carrying `GaugeType` was treated as identical and the warning was dropped. Fixed by scanning every sample for a gauge hint (the same way `histogramRate` does) before the early return.

## Finding #3 — single-sample histograms differ from floats

Confirmed intentional and documented in code: a single-sample histogram `xrate`/`xdelta` (and `xincrease` past the injection window) emits no sample, because there is no meaningful zero-shaped histogram to return — this matches Prometheus' `histogramRate`, which returns nil for fewer than two points. The float path's `0` is left unchanged.

## Two additional fixes found while working through #1

- **In-window clobber (overlapping windows):** with overlapping steps, `Reset` retains the whole in-window suffix, so the buffer's last element can be an in-window point. Classifying the prefetched sample against the mutated cursor could `ReadIntoLast`-overwrite that in-window point. Fixed by the `rangeStart` classification above.
- **Stale baseline beyond ext-lookback:** when a step gap exceeds `ExtLookbackDelta`, a pre-range sample older than the lookback could be resurrected as a baseline. `Reset` already drops such points; the `lastSample` and iterator paths now apply the same ext-lookback bound, so all three agree. The effective ext-lookback is now stored once on the selector and shared by the buffer and `selectPoints`.

## Tests

- **Rewrote `TestXFunctionsNativeHistogramsRangeQuery`** from an overlapping-only, type/cardinality-only check into numeric assertions: back-to-back `xincrease` tiling the range, a gapped case that pins the boundary baseline (the #1 regression guard), and `xrate == xincrease / rangeSeconds`.
- **Corrected `TestXFunctionsRangeQuery/gaps between steps`**: `9 → 5`.
- **New `TestXIncreaseOverlappingBoundaryNoClobber`**: counter `100,200,0,10` at `0,10s,10.001s,20s`, `xincrease(m[20s])` at 10s step → `110` at 20s (was `10`).
- **New `TestXIncreaseStaleBaselineBeyondExtLookback`** (+ native-histogram variant): two pre-range samples older than a 1h lookback, `xincrease(m[10s])` at 2h step → no stale baseline used.
- **New `TestExtendedRateHistogramZeroInjectionWarns`**: equal-valued samples with `NotCounterReset` then `GaugeType` must still warn.
- **New `TestExtendedRateHistogramSingleSampleUndefined`**: single-sample `xdelta` and post-window `xincrease` emit nothing.

All of the regression tests were confirmed to fail against the pre-fix code.

## Verification

`go test ./...` passes, except a pre-existing flaky wall-clock timing assertion (`TestQueryAnalyze/time()`), which fails ~1/5 on a clean tree independent of this change. `gofmt`, `go vet`, and `git diff --check` are clean.

## Out of scope (possible follow-up)

The classic (non-extended) `rate`/`increase`/`delta` iterator path has a pre-existing, unrelated exact-cursor drop: with overlapping windows it can drop a sample landing exactly on `buffer.MaxT()+1` (i.e. 1ms after the last retained sample). It predates this PR and does not affect the extended functions, which are now robust to overlapping windows. Happy to open a separate issue for it if useful.

MichaHoffmann (approved):

PR #725, opened 2026-08-13
Replace extlabels.DropReserved with  Prometheus Labels.DropReserved

PaurushGarg:
**Problem**
`matrixSelector.loadSeries()` calls `extlabels.DropReserved(lbls, b)` for every series to remove reserved labels. This method internally allocates a new `[]Label` slice per series. At high cardinality, this adds to the heap profile.
  
  **Fix**
 Replace `extlabels.DropReserved(lbls, b)` with the Prometheus `lbls.DropReserved(schema.IsMetadataLabel)`. This method handles the common case — removing `__name__` — without copying the entire label set, saving allocations at high cardinality.

  
  **Benchmark**
  No latency impact. The behavior is identical — both drop the same labels using the same filter function (`schema.IsMetadataLabel`). The only difference is the native method avoids copying the label set into a new slice.

### DIFF
--- a/storage/prometheus/matrix_selector.go
+++ b/storage/prometheus/matrix_selector.go
@@ -14,7 +14,6 @@ import (
 	"github.com/thanos-io/promql-engine/execution/model"
 	"github.com/thanos-io/promql-engine/execution/parse"
 	"github.com/thanos-io/promql-engine/execution/telemetry"
-	"github.com/thanos-io/promql-engine/extlabels"
 	"github.com/thanos-io/promql-engine/query"
 	"github.com/thanos-io/promql-engine/ringbuffer"
 	"github.com/thanos-io/promql-engine/warnings"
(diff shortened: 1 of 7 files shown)


fpetkovski (approved):
Nice, good change.

storage/prometheus/matrix_selector.go
```diff
@@ -269,13 +269,12 @@ func (o *matrixSelector) loadSeries(ctx context.Context) error {
...
 		o.scanners = make([]matrixScanner, len(series))
 		o.series = make([]labels.Labels, len(series))
-		var b labels.ScratchBuilder
 
 		for i, s := range series {
 			origLbls := s.Labels()
 			lbls := origLbls
 			if o.functionName != "last_over_time" && o.functionName != "first_over_time" {
-				lbls = extlabels.DropReserved(lbls, b)
+				lbls = lbls.DropReserved(schema.IsMetadataLabel)
```
yeya24, line 277: We have `extlabels.DropReserved(lbls, b)` used in multiple places in the codebase. Can you check if we should change them as well?

PaurushGarg: Thanks for this call out.
I have updated. 
Now the codebase is consistent - no `extlabels.DropReserved` usage, all using the prometheus method.

yeya24 (approved):

PR #726, opened 2026-08-13
storage/prometheus: add projectedSelector for engine-level label projection

### LINKED ISSUE #689: High cardinality Joins caused OOM kill due to large result labels

yeya24:
## Summary

Implements engine-level label projection at the series selector layer, as suggested by @fpetkovski and @MichaHoffmann in #691.

When the `ProjectionOptimizer` determines that a downstream aggregation (e.g. `sum`, `count`) does not need all labels, the `projectedSelector` wraps the underlying `SeriesSelector` and strips unnecessary labels before they are materialized in memory. This prevents OOM kills on high-cardinality queries where label materialization dominates heap usage (e.g. 2.7M series × 20 labels = 11.8GB just for labels).

## Design

- `projectedSelector` wraps any `SeriesSelector` and applies include/exclude label projection on `GetSeries()`
- Appends a `__series_hash__` label to preserve series identity after projection (needed for dedup/matching)
- Activated via `WithSeriesHashLabel` functional option on `NewPrometheusScanners` — callers that do not set it get no projection (backward compatible)
- No-op when `seriesHashLabel` is empty or projection is nil/empty-exclude

## Motivation

From production OOM investigation (AMP cell-16): a customer query `sum(avg_over_time(k8s_pod_memory_usage_bytes{...}[1h])) or count(...)` materialized 2.7M series labels (11.8GB) that were immediately discarded by the `sum()`/`count()` aggregation. With projection pushdown at the selector level, these labels are never materialized.

## Relationship to #691

PR #691 applies projection at the binary operator level (result metric materialization). This PR applies projection earlier — at the selector level — which prevents label materialization in **any** consumer (aggregations, functions, binary ops). Both are complementary; this one addresses the case where the consuming operator does not need labels at all.

## Tests

- 10 unit tests for `projectedSelector` (include/exclude modes, sharding, identity preservation)
- 19 deterministic engine tests (instant + range, all aggregation types)
- 10,000 random fuzz queries comparing projected vs non-projected engine output
- Benchmarks showing memory savings scale with label count and series count

## Open questions for discussion

1. Should the cost evaluation (when to apply projection) be a separate optimizer pass?
2. Should projection be applied before or after the SelectorPool cache? (Currently after — pool hash includes projection info)
3. Integration with storage-level projection hints (Thanos #8893)

Fixes #689 (partially — complements #691)

### DIFF
--- a/storage/prometheus/projected_selector.go
+++ b/storage/prometheus/projected_selector.go
@@ -1,0 +1,143 @@
+// Copyright (c) The Thanos Community Authors.
+// Licensed under the Apache License 2.0.
+
+package prometheus
+
+import (
+	"context"
+	"slices"
+	"strconv"
+	"sync"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+
+	"github.com/thanos-io/promql-engine/logicalplan"
+)
+
+// projectedSelector wraps a SeriesSelector and applies label projection
+// on GetSeries(). This reduces memory by only materializing the labels
+// that downstream operators actually need (e.g. grouping labels for aggregations).
+type projectedSelector struct {
+	selector        SeriesSelector
+	projection      *logicalplan.Projection
+	seriesHashLabel string
+
+	once   sync.Once
+	series []SignedSeries
+}
+
+// NewProjectedSelector creates a new projected selector that wraps the given selector
+// and applies label projection based on the provided projection hints.
+// seriesHashLabel is the label name used to store the original series hash
+// for identity preservation (typically "__series_hash__").
+// If seriesHashLabel is empty, projection is not applied at the selector level
+// (the remote storage or queryable wrapper is expected to handle it instead).
+func NewProjectedSelector(selector SeriesSelector, projection *logicalplan.Projection, seriesHashLabel string) SeriesSelector {
+	if projection == nil || isNoOpProjection(projection) || seriesHashLabel == "" {
+		return selector
+	}
+	return &projectedSelector{
+		selector:        selector,
+		projection:      projection,
+		seriesHashLabel: seriesHashLabel,
+	}
+}
+
+// isNoOpProjection returns true if the projection would not actually remove any labels.
+// An exclude-mode projection with no labels to exclude is a no-op.
+func isNoOpProjection(p *logicalplan.Projection) bool {
+	return !p.Include && len(p.Labels) == 0
+}
+
+func (p *projectedSelector) Matchers() []*labels.Matcher {
+	return p.selector.Matchers()
+}
+
+func (p *projectedSelector) GetSeries(ctx context.Context, shard, numShards int) ([]SignedSeries, error) {
+	var err error
+	p.once.Do(func() { err = p.loadSeries(ctx) })
+	if err != nil {
+		return nil, err
+	}
+
+	return seriesShard(p.series, shard, numShards), nil
+}
+
+func (p *projectedSelector) loadSeries(ctx context.Context) error {
+	series, err := p.selector.GetSeries(ctx, 0, 1)
+	if err != nil {
+		return err
+	}
+
+	p.series = make([]SignedSeries, 0, len(series))
+	for i, s := range series {
+		projected := p.projectSeries(s.Series)
+		p.series = append(p.series, SignedSeries{
+			Series:    projected,
+			Signature: uint64(i),
+		})
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


storage/prometheus/projected_selector.go
```diff
@@ -0,0 +1,143 @@
...
+type projectedSelector struct {
+	selector        SeriesSelector
+	projection      *logicalplan.Projection
+	seriesHashLabel string
+
+	once   sync.Once
+	series []SignedSeries
+}
+
+// NewProjectedSelector creates a new projected selector that wraps the given selector
```
GiedriusS, line 41: What's the use case of this? Is this e.g. for applying projections on data coming from the TSDB that doesn't apply them right now?

storage/prometheus/projected_selector.go
```diff
@@ -0,0 +1,143 @@
...
+		lset:   projectedLabels,
+	}
+}
+
+// projectLabels applies the projection to a label set, keeping or removing
+// labels based on the include/exclude mode. Always preserves __name__ is NOT
+// done here — the projection optimizer decides which labels to include.
+// The series hash label is appended to preserve original series identity.
+func (p *projectedSelector) projectLabels(lset labels.Labels) labels.Labels {
+	builder := labels.NewBuilder(labels.EmptyLabels())
```
GiedriusS, line 186: This forces new allocations (and a lot of them) so if we end up merging this I would like this to be under a flag somewhere.

storage/prometheus/projected_selector.go
```diff
@@ -0,0 +1,143 @@
...
+	series []SignedSeries
+}
+
+// NewProjectedSelector creates a new projected selector that wraps the given selector
+// and applies label projection based on the provided projection hints.
+// seriesHashLabel is the label name used to store the original series hash
+// for identity preservation (typically "__series_hash__").
+// If seriesHashLabel is empty, projection is not applied at the selector level
+// (the remote storage or queryable wrapper is expected to handle it instead).
+func NewProjectedSelector(selector SeriesSelector, projection *logicalplan.Projection, seriesHashLabel string) SeriesSelector {
```
GiedriusS, line 47: I personally believe this needs to be solved at the StoreAPI/leaf/etc. level. I have a fork, for example, of the TSDB that respects the projection hints but if it's under a flag then I would be ok with merging this.

yeya24: I agree with that as well. But for those projects without a storage layer that respects projection hints, having a layer to allow projection pushdown can be helpful to prevent OOM kills in downstream operators.

This change is already optional under `WithSeriesHashLabel`.

storage/prometheus/projected_selector.go
yeya24: Right we don't support projection pushdown yet in Cortex but it is hard to address #689 without a proper projection pushdown. This PR tries to do it at the engine level.

GiedriusS (commented):
Maybe this could be implemented simply as a separate optimizer?

yeya24:
@GiedriusS Can you elaborate more why this can be implemented as an optimizer? It doesn't change any of the existing logical plan.

PR #733, opened 2026-08-31
Distribute plain selectors in binop

fpetkovski:
In expressions like `clamp_max(metric_a, scalar(metric_b))` only one selector is distributed and the second is left as is.

### DIFF
--- a/logicalplan/distribute.go
+++ b/logicalplan/distribute.go
@@ -278,6 +278,18 @@ func (m DistributedExecutionOptimizer) computeDistributionPoints(plan *Node, par
 				}
 			}
 			marks[current] = struct{}{}
+			return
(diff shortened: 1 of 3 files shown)


MichaHoffmann (approved):

MichaHoffmann (approved):

fpetkovski:
I think the optimizer needs some refactoring, over time the implementation drifted from the intended algorithm. The way it should work is:
* Traversing bottom down, mark each node as distributive if it preserves partition labels
* In a second pass, traverse bottom up and distribute each node which itself is distributive but its parent is not.

I will merge this PR for now and do the refactoring as a follow up. I will also document the algorithm for the optimizer.

PR #737, opened 2026-09-11
feat(storage): share selector pool entries across different range windows

PaurushGarg:
When a query uses the same metric with different range windows, like `rate(A[5m]) / rate(A[2m])`, the SelectorPool created separate cache entries because the hash included mint (which differs due to the range window). This caused duplicate Select/gRPC calls for the same data.

Now mint is removed from the pool hash key, so both selectors match the same cache entry. And, the cached selector's mint is widened to the earliest value so the TSDB query covers both ranges. The chunk iterator already seeks to the correct position, so no extra data processing is needed. The pool is per-query, so the widening is scoped and safe.
  
hints.Func could also be removed from the hash since most storage backends return the same data regardless of Func. However, Thanos uses Func for downsampling decisions, so Func remains in the default hash to keep it safe for all downstream projects. A SelectorHashFunc type is exposed via engine.Opts.SelectorHashFunc so callers can provide a custom hash that also excludes Func.
  
Added unit tests for shared range windows and custom hash function override.

### DIFF
--- a/storage/prometheus/pool.go
+++ b/storage/prometheus/pool.go
@@ -28,31 +28,41 @@ func NewSelectorPool(querier storage.Querier) *SelectorPool {
 }
 
 func (p *SelectorPool) GetSelector(mint, maxt, step int64, matchers []*labels.Matcher, hints storage.SelectHints) SeriesSelector {
-	key := hashMatchers(matchers, mint, maxt, hints)
-	if _, ok := p.selectors[key]; !ok {
-		p.selectors[key] = newSeriesSelector(p.querier, matchers, hints)
+	key := hashMatchers(matchers, maxt, hints)
+	if existing, ok := p.selectors[key]; ok {
+		if mint < existing.hints.Start {
+			existing.hints.Start = mint
+		}
+		return existing
 	}
+	p.selectors[key] = newSeriesSelector(p.querier, matchers, hints)
 	return p.selectors[key]
 }
 
 func (p *SelectorPool) GetFilteredSelector(mint, maxt, step int64, matchers, filters []*labels.Matcher, hints storage.SelectHints) SeriesSelector {
-	key := hashMatchers(matchers, mint, maxt, hints)
-	if _, ok := p.selectors[key]; !ok {
-		p.selectors[key] = newSeriesSelector(p.querier, matchers, hints)
+	key := hashMatchers(matchers, maxt, hints)
+	if existing, ok := p.selectors[key]; ok {
+		if mint < existing.hints.Start {
+			existing.hints.Start = mint
+		}
+		return NewFilteredSelector(existing, NewFilter(filters))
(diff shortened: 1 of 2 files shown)


storage/prometheus/pool.go
```diff
@@ -28,31 +28,41 @@ func NewSelectorPool(querier storage.Querier) *SelectorPool {
 }
 
 func (p *SelectorPool) GetSelector(mint, maxt, step int64, matchers []*labels.Matcher, hints storage.SelectHints) SeriesSelector {
-	key := hashMatchers(matchers, mint, maxt, hints)
-	if _, ok := p.selectors[key]; !ok {
-		p.selectors[key] = newSeriesSelector(p.querier, matchers, hints)
+	key := hashMatchers(matchers, maxt, hints)
+	if existing, ok := p.selectors[key]; ok {
+		if mint < existing.hints.Start {
```
fpetkovski, line 33: We can use the built-in `min` function here and bellow.

fpetkovski (commented):
The pool is per-query which is why this is safe?

storage/prometheus/pool.go
```diff
@@ -28,31 +28,41 @@ func NewSelectorPool(querier storage.Querier) *SelectorPool {
...
 	for _, m := range matchers {
 		writeMatcher(sb, m)
 	}
-	writeInt64(sb, mint)
+	// mint excluded so that different range windows (e.g. [5m] vs [2m]) share one Select().
 	writeInt64(sb, maxt)
 	writeInt64(sb, hints.Step)
-	writeString(sb, hints.Func)
+	// Only "series" affects TSDB behavior (skips chunk loading).
+	if hints.Func == "series" {
```
yeya24, line 63: This is incorrect. I do think Thanos uses Func in its downsampling logic. It sounds not safe and a behavior change to downstream projects. For projects like Cortex Func is a noop. Maybe a better way is to expose a way so that downstream projects can override this hashing logic based on their own requirements.

storage/prometheus/pool.go
PaurushGarg: @fpetkovski thanks. updated accordingly.

storage/prometheus/pool.go
PaurushGarg: @yeya24 thanks. 
makes sense. I've kept Func in the default hash so Thanos behavior is unchanged. 
For projects like Cortex where Func is a noop, I've accordingly added a SelectorHashFunc on engine.Opts so they can provide a custom hash that excludes Func and share selectors across functions.

PaurushGarg:
> The pool is per-query which is why this is safe?

@fpetkovski thanks. 
yes the pool is per-query, created per-query inside storageScanners(). So mint widening would effect that query only. I have updated this in the comment, and also in the pr description.

